Issue #2, opened 2018-10-18
Beginner Issues 


satu0king:
These are just few ideas or bugs I know right now. I will keep updating the list. Before starting anything, create an issue and await instructions. Feel free to suggest any features or bug fixes

Simulator
- [ ] Counter 
- [ ] Decimal to BCD and vice versa 
- [ ] Combinational Analysis - Make it easier to give input. 
- [x] 7 Segment - change color 
- [ ] Add undo - redo buttons 
- [ ] Undo - fixes - combinational and new project loaded: call backup()
- [ ] Test and Fix shortcuts in firefox 
- [x] Move from this to prototype 
- [x] Complete Plotting implementation
- [ ] Disable copying / combinational circuits according to assignment preferences 
- [ ] Watermark with link in embed - outside only 
- [ ] Add scroll bar to sidebar 
- [ ] Add a link to assignment desription on the simulator page
- [ ] Enable minimap to show rotated elements
- [ ] Fix Trivial cases - constant and Z state in combinational analysis  
- [ ] Configure shortcuts for Zoom/save online 
- [ ] Setup a good offline backing of circuits

bug fixes 
- [ ] Save empty circuit causes problems 
- [ ] Subcircuit of empty circuit causes problems 
- [ ] Image of empty circuit causes problems 
- [ ] Undo after deleting subcircuit (ie main circuit was deleted) causes problems. 

Ambitious Features 
- [ ] Expresso - replacement of combinational analysis 
- [x] Custom layout for sub circuits - logisim does a brilliant job at this 
- [ ] Write Test suite for simulator

Website - rails + normal front end 
- [ ] Configure back button correctly 
- [x] Flash messages ( notice ) not configured correctly 
- [ ] Assignment Grading system
- [ ] Customize Dashboard 
- [ ] Add shortcuts to getting started page
- [ ] Add a link to view project in assignment page
- [ ] Trending and Featured sections 
- [ ] Preview should go full screen
- [ ] Email for assignment submissions for students. Assignment report for mentor 
- [ ] Use CSS and beautify email
- [ ] Follow users ( probably later )
- [ ] Write Test suite for backend
- [ ] Countries and University - to users
- [ ] Add shortcut on homepage for pending assignments
- [ ] Add shortcut on simulator page for assignment description

Others 
- [ ] Powerpoint Plug-In

Documentation 
[...]

PR #29, opened 2018-10-21
Problems related to inserting subcircuit into another circuit

mohammedsafwat:
I have suggested some checks and changes that could help fix some problems related to inserting sub-circuits into other circuits. I would like to know your feedback (if I totally understand the concept) and if these fixes help to solve some of the problems related to inserting sub-circuits. Thank you!

This is related to issue #2.

### DIFF
--- a/public/js/UX.js
+++ b/public/js/UX.js
@@ -125,16 +125,15 @@ function showProperties(obj) {
         if (!obj.propagationDelayFixed)
             $('#moduleProperty-inner').append("<p>Delay: <input class='objectPropertyAttribute' type='number'  name='changePropagationDelay' min='0' max='100000' value=" + obj.propagationDelay + "></p>");
 
-
-        $('#moduleProperty-inner').append("<p>Label: <input class='objectPropertyAttribute' type='text'  name='setLabel'  value='" + escapeHtml(obj.label) + "'></p>");
-
+        if (obj.hasOwnProperty('label')) {
+            $('#moduleProperty-inner').append("<p>Label: <input class='objectPropertyAttribute' type='text'  name='setLabel'  value='" + escapeHtml(obj.label) + "'></p>");
+        }
 
         if (!obj.labelDirectionFixed) {
             var s = $("<select class='objectPropertyAttribute' name='newLabelDirection'>" + "<option value='RIGHT' "+ ["","selected"][+(obj.labelDirection=="RIGHT")] +  " >RIGHT</option><option value='DOWN' "+ ["","selected"][+(obj.labelDirection=="DOWN")] +  " >DOWN</option><option value='LEFT' " + "<option value='RIGHT'"+ ["","selected"][+(obj.labelDirection=="LEFT")] +  " >LEFT</option><option value='UP' " + "<option value='RIGHT'"+ ["","selected"][+(obj.labelDirection=="UP")] +  " >UP</option>"+"</select>");
             s.val(obj.labelDirection);
             $('#moduleProperty-inner').append("<p>Label Direction: " + $(s).prop('outerHTML') + "</p>");
         }
-
 
         if (!obj.directionFixed) {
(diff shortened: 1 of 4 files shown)


public/js/UX.js
```diff
@@ -116,7 +116,8 @@ function showProperties(obj) {
 
         $('#moduleProperty-inner').append("<div id='moduleProperty-header'>" + obj.objectType + "</div>");
         // $('#moduleProperty').append("<input type='range' name='points' min='1' max='32' value="+obj.bitWidth+">");
-        if (!obj.fixedBitWidth)
+
+        if (!obj.fixedBitWidth && obj.bitWidth)
```
satu0king:
```suggestion
        if (!obj.fixedBitWidth && obj.bitWidth)
```
All objects have bitWidth so this is not exactly required.

public/js/UX.js
```diff
@@ -125,17 +126,16 @@ function showProperties(obj) {
         if (!obj.propagationDelayFixed)
             $('#moduleProperty-inner').append("<p>Delay: <input class='objectPropertyAttribute' type='number'  name='changePropagationDelay' min='0' max='100000' value=" + obj.propagationDelay + "></p>");
 
-
-        $('#moduleProperty-inner').append("<p>Label: <input class='objectPropertyAttribute' type='text'  name='setLabel'  value='" + escapeHtml(obj.label) + "'></p>");
-
+        if (obj.label) {
```
satu0king: Same as previous comment, all objects have label

public/js/data.js
```diff
@@ -217,7 +217,7 @@ function extract(obj) {
 // Check if there is anything to backup - to be deprecated
 function checkIfBackup(scope) {
     for (var i = 0; i < updateOrder.length; i++)
-        if (scope[updateOrder[i]].length) return true;
+        if (scope.hasOwnProperty(updateOrder[i]) && scope[updateOrder[i]].length) return true;
```
satu0king: Could be useful but I don't think it is necessary.

public/js/data.js
```diff
@@ -767,27 +767,32 @@ createOpenLocalPrompt = function() {
...
         width: "auto",
         buttons: [{
             text: "Insert SubCircuit",
             click: function() {
                 if (!$("input[name=subCircuitId]:checked").val()) return;
-                simulationArea.lastSelected = new SubCircuit(undefined, undefined, globalScope, $("input[name=subCircuitId]:checked").val());
+                let subcircuitId = parseInt($("input[name=subCircuitId]:checked").val(), 10);
+                const savedData = JSON.parse(generateSaveData("Untitled"));
+                const subcircuitData = savedData.scopes.filter((circuitData) => circuitData.id === subcircuitId)
+                simulationArea.lastSelected = new SubCircuit(undefined, undefined, globalScope, subcircuitId, subcircuitData);
```
satu0king: I am not sure what is happening here. However in the SubCircuit module, if the subcircuitData is empty, it instantiates the object. It is meant to be empty when instantiated.

satu0king:
Also, it is a good practice to sync your master branch with the original repo's master and never directly change anything on your master branch. This is because your changes may not get accepted or even if they do, the original repo owner might chose to rebase or squash the commits instead. More importantly, other people's commits will not be reflected correctly in your repo. I suggest that you hard reset your master branch to reflect the CircuitVerse's master branch

public/js/data.js
mohammedsafwat: I realized that when you insert a subcircuit even if it's not empty, you always get an error that the subcircuit that you are trying to insert is empty. When I tracked the steps a little bit I found that the subcircuit data is 'undefined'. This is why I tried to pass the subcircuit data here.

mohammedsafwat:
Sorry for the late reply, thank you so much for your feedback and your advice, really appreciate it. I totally understand now.

public/js/data.js
satu0king:
```
 if (subcircuitScope.Input.length == 0 && subcircuitScope.Output.length == 0) {
        showError("SubCircuit : " + subcircuitScope.name + " is an empty circuit");
        this.delete();
        this.scope.backups = [];
        return;
    }
```
I think this condition can be removed but it is there for now. 
This is where it says it is an empty circuit. At least one input or one output should be there. Otherwise that circuit isn't doing anything wrt to the parent circuit. 
Please confirm that this is the error you were getting. Otherwise it is a serious issue that needs to be resolved asap.

mohammedsafwat: Thank you for the fast follow up! Yes, this is exactly the condition that is causing the "empty circuit" error that will be displayed when you insert a subcircuit even if has some connections

PR #46, opened 2018-11-09
fix compatibility

HypheX:


HypheX:
It turns out... javascripter.net lied to me, the keycode for + is 171... not 61   so zooming out works, but zooming in does not.

PR #64, opened 2018-12-02
Increase "recent circuits" length

jbox144:
Will provide slightly more space for projects to be seen

### DIFF
--- a/app/views/logix/index.html.erb
+++ b/app/views/logix/index.html.erb
@@ -85,7 +85,7 @@
 
     <h5 class="">Recent Circuits</h5>
     <div class="row">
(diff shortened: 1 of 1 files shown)


jbox144:
I litterally changed one character :P

satu0king:
I think 12 circuits is high enough for a page. However, I think we should look into having multiple pages for the same. Could you look into pagination? 

This might be a starting point
https://hackhands.com/pagination-rails-will_paginate-gem/

jbox144:
I'm not very experienced with ruby, or even websites in general, as of yet. I'm only capable of doing basic patches and fixes to front end code.

PR #66, opened 2018-12-02
Additional WASD inputs for direction controls

jbox144:
Bound the WASD keys to control the direction of components, similar to the arrow keys.


satu0king:
Merged

PR #80, opened 2018-12-14
Fix zooming in/out in firefox. Fix Ctrl+T shortcut (#52)

satu0king:
* Fix zooming in/out in firefox. 

* Compatibility fix for firefox.

Fixes the Ctrl +/- incompatibility on firefox for embeded circuits.

* Update listeners.js


satu0king:
Left the ctr+t shortcut for now. We should not override the default browser behaviour. It is probably better to change the shortcut key as u suggested.

Issue #47, opened 2018-11-09
Display Matrix


satu0king:
Users are making their own display using LEDs. 
![image](https://user-images.githubusercontent.com/2092958/48241888-c0c2f280-e3fe-11e8-8d1c-84762f7eeca2.png)

This is great. However, It probably is easier to have a custom display matrix component. 
![image](https://user-images.githubusercontent.com/2092958/48241960-fd8ee980-e3fe-11e8-8179-801460de9000.png)

@blane1257 what do you think of it?

satu0king:
> I will point out though the pixels on that are pretty small with too much space in-between them, and the dark color doesn't fit the theme too well.

This was just a random image from the internet. I will not be having the dark color or too much space.

blane1257:
Oh alright, good.

HypheX:
Actually, yes, a display matrix component would be awesome.

fpintos:
If it helps anyone out there:

I've built a 16x8 pixels, 16-color graphical display that can be cut-and-pasted into other projects:
https://circuitverse.org/users/1935/projects/5010

In principle, larger displays could be built by stacking this basic one, but at some point the simulation might end up with lag, I suppose.

With this as a proof of concept, it would be great to have a built-in component to reduce complexity.

Personally, I would prefer square pixels and maybe an option to not have any gaps between them.
It would also be nice to be able to configure the size and color depth of the display (but I suppose if the component provides full RGB then smaller color depths can be done by external circuitry, like I did in the project above).

fpintos:
That is a pretty awesome display.

fpintos:
Hi @blane1257 , checkout the PR I sent that will allow us to build a display matrix with zero circuitry around the LEDs: #104 

Here is a sample - note this is built with just square RGB LEDs. The matrix is full color, depending only on the controlling circuit we attach to it. :-)

![image](https://user-images.githubusercontent.com/4239171/52696927-1d193f00-2f25-11e9-8de5-bc2eb979da98.png)

PR #104, opened 2019-02-11
Add SquareRGBLed component

fpintos:
This change adds a new SquareRGBLed component with the following features:

- The visual of the LED is square, which makes it possible to build a LED matrix with less visual interference.
- The square LED has the same inputs as the existing RGBLed.
- The square LED can be turned, allowing the inputs to be on any side.
- The LED is 3x3 in size, centered on a 4x4 grid:
image
- This design allows the following constructions:
  - LEDs can be stacked without gap in a 2xN matrix and the circuitry will be outside the matrix:
image
  - A single LED can be placed on top of a 4x4 sub-circuit, which outputs the RGB values on the bottom of its layout:
image
  - A larger matrix can be created by stacking longer sub-circuits of size 4x(4*N). For example, three 4x12 drivers stacked create a uniform 3x3 LED matrix, where the pixels are spaced by one unit, both horizontally and vertically:
image
  *That being said, the thick line of the sub-circuit ends up creating an illusion that the vertical space is larger than the horizontal space; I suppose we'll live with that for now and solve it with a full LED matrix component.*

NOTE: this is related to #47, but not the full solution.

As another example, one *could* create a multiplexing display matrix by stacking a bunch of drivers; each driver takes the horizontal enable input on the size and a combined RGB input for each pixel on top; it then output the combined RGB on the bottom, next to the individual RBG inputs for the LED; the pins should be placed so the combined input and output are vertically aligned. One can make an arbitrary long row like this, effectively using 4 units of space for each LED. Then it is simply a matter of stacking several rows, multiplexing them, like a real LED matrix.
Ex:
image

Attached is a file that exercises the new component.
[SquareRGBLedTest.txt](https://github.com/CircuitVerse/CircuitVerse/files/2851006/SquareRGBLedTest.txt)

### DIFF
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -2543,7 +2543,7 @@ SquareRGBLed.prototype.customDraw = function () {
     var b = this.inp2.value;
     var c = this.inp3.value;
     ctx.strokeStyle = "#d3d4d5";
-    ctx.fillStyle = ["rgba(" + a + ", " + b + ", " + c + ", 0.8)", "rgba(227, 228, 229, 0.8)"][((a === undefined || b === undefined || c === undefined)) + 0]
+    ctx.fillStyle = ["rgb(" + a + ", " + b + ", " + c + ")", "rgba(227, 228, 229)"][((a === undefined || b === undefined || c === undefined)) + 0]
     ctx.lineWidth = correctWidth(1);
     ctx.beginPath();
     rect2(ctx, -15, -15, 30, 30, xx, yy, this.direction);
@@ -2552,7 +2552,7 @@ SquareRGBLed.prototype.customDraw = function () {
     if ((this.hover && !simulationArea.shiftDown) ||
         simulationArea.lastSelected == this ||
         simulationArea.multipleObjectSelections.contains(this)) {
-            ctx.fillStyle = "rgba(255, 255, 32,0.8)";
+            ctx.fillStyle = "rgba(255, 255, 32)";
     }
 
     ctx.fill();
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -2468,14 +2468,15 @@ RGBLed.prototype.customDraw = function() {
     ctx.fill();
 }
 
-function SquareRGBLed(x, y, scope = globalScope, dir = "UP", pinLength = 5) {
+function SquareRGBLed(x, y, scope = globalScope, dir = "UP", pinLength = 1) {
     CircuitElement.call(this, x, y, scope, dir, 8);
     this.rectangleObject = false;
     this.setDimensions(15, 15);
-    this.pinLength = pinLength === undefined ? 5 : pinLength;
-    this.inp1 = new Node(-15 - this.pinLength, -10, 0, this, 8);
-    this.inp2 = new Node(-15 - this.pinLength, 0, 0, this, 8);
-    this.inp3 = new Node(-15 - this.pinLength, 10, 0, this, 8);
+    this.pinLength = pinLength === undefined ? 1 : pinLength;
+    var nodeX = -10 - 10 * pinLength;
+    this.inp1 = new Node(nodeX, -10, 0, this, 8, "R");
+    this.inp2 = new Node(nodeX, 0, 0, this, 8, "G");
+    this.inp3 = new Node(nodeX, 10, 0, this, 8, "B");
     this.inp = [this.inp1, this.inp2, this.inp3];
     this.labelDirection = "UP";
     this.fixedBitWidth = true;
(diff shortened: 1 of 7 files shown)


fpintos:
I've made a few changes to the initial component:
- the square LED is now opaque - think of it as sitting flat on a breadboard; it would not have transparency like a straight up LED. This allows the LED to hide circuitry passing behind it, like the pins of other LEDs (but not wires, which draw on top of them).

- I've added a property to control the length of the square LED pin - this allows the connection nodes to be far away from the LED itself. These large pins draw in gray, so they don't call attention to themselves.
image

With these features we can now build a proper display matrix straight out of square LEDs:
image
This matrix is built by adding a LED with a pin length of 12, then you copy it and start pasting down. Each LED lays on top of the previous, hiding the circuitry. Once the row is complete, copy the row and start pasting to the right. In the end, you have the LEDs in a nice matrix and the connection pins are below the last row, in the same order as the LEDs. Those pins can be connected to controlling sub-circuits as usual. Since the LED can be turned, the circuitry can be on any side of the LEDs.

To make some designs easier, I've also made the LED turn on when at least one input is connected.

Here's a full example:
image

blane1257:
This is really cool! Cant wait to find use from it!

satu0king:
Very neat! loved it. This potentially has many uses.

public/js/modules.js
```diff
@@ -2468,6 +2468,109 @@ RGBLed.prototype.customDraw = function() {
...
+        var diff = 10 * (pinLength - this.pinLength);
+        var diffX = this.direction == "LEFT" ? -diff : this.direction == "RIGHT" ? diff : 0;
+        var diffY = this.direction == "UP" ? -diff : this.direction == "DOWN" ? diff : 0;
+
+        // Build a new LED with the new values; preserve label properties too.
+        var obj = new window[this.objectType](this.x + diffX, this.y + diffY, this.scope, this.direction, pinLength);
+        obj.label = this.label;
+        obj.labelDirection = this.labelDirection;
+
+        this.cleanDelete();
```
satu0king: Any reason why delete the object and create a new one instead of just moving the position of the current object?

fpintos: The reason I did it this way was to avoid having to manipulate the existing nodes, to keep them in the same position. Now I realize that deleting and recreating causes connecting wires to get deleted as well, so maybe it would be interesting to try the other way around and see if we can keep the wires connected when changing pin length.

satu0king: Yes, I agree. 
It is fine for now, it won't affect the working of the LEDs in any way. Please feel free to create a new issue/PR regarding the same.

satu0king:
I tried using the module. It is great with one issue. 
The LED is small to be kept compact. Maybe I am missing something. 

image

https://circuitverse.org/users/3/projects/8618

However, @blane1257 has created a circuit which is quite compact. Unfortunately, the nodes show up too. 

image

https://circuitverse.org/users/3/projects/1354

Is there, any improvement we can make on this?

blane1257:
an easy solution to get the nodes out of the way of course is to simply increase the pin length, i sacrificed the aesthetic however to make my circuit more compact. If it's possible to draw the LEDs above nodes and wires, this would of course be optimal, but i'm not sure if the code allows for such a change

fpintos:
@blane1257 , you are correct that if you were to extend the pins then the LEDs would be isolated from the circuitry.

The other option, as you mentioned, would be to tweak the rendering so we skip drawing the nodes from the LEDs. @satu0king , do you think that is reasonable?

And the other option which I'm willing to tackle is to simply build a proper display matrix, as a single component. Now that I got my hands and feet in the water it looks fairly straightforward - I just need to take the time to do it. :-)

fpintos:
> I tried using the module. It is great with one issue.
> The LED is small to be kept compact. Maybe I am missing something.
> 
> image
> 
> https://circuitverse.org/users/3/projects/8618
> 
> However, @blane1257 has created a circuit which is quite compact. Unfortunately, the nodes show up too.
> 
> image
> 
> https://circuitverse.org/users/3/projects/1354
> 
> Is there, any improvement we can make on this?

@satu0king , that is a cool circuit. You can eliminate the gap in your LEDs by having all led buffers of a single row in a sub-circuit. Then, in that circuit you export the RGB pins of each LED right next to each other. With that setup the LEDs will be touching.

satu0king:
> @blane1257 , you are correct that if you were to extend the pins then the LEDs would be isolated from the circuitry.
> 
> The other option, as you mentioned, would be to tweak the rendering so we skip drawing the nodes from the LEDs. @satu0king , do you think that is reasonable?
> 
> And the other option which I'm willing to tackle is to simply build a proper display matrix, as a single component. Now that I got my hands and feet in the water it looks fairly straightforward - I just need to take the time to do it. :-)

To disable node rendering for connected nodes for all circuit elements is straightforward. I had tried before but it looked confusing for some circuit element, maybe we can configure it on a per circuit element basis.

Issue #81, opened 2018-12-26
RAM component


superstewie:
I believe a RAM component should be made. When working with large RAM, weird errors start to occur around 16k bytes. I tried two methods of building RAM, one method which uses progressively larger matrices of 8 bit registers which fails around 16k bytes, and one method which uses multiples of the circuit below:
![capture](https://user-images.githubusercontent.com/36019929/50452754-b0f56400-0901-11e9-932e-040360472e43.PNG)
Which uses a 32 bit register and some multiplexers to store 4 bytes in a single register. It fails around 16 bytes.

Some of the failures are very, interesting:
![capture1](https://user-images.githubusercontent.com/36019929/50452947-d040c100-0902-11e9-81e5-48905e8b1e1e.PNG)

Where the minimap shows that there are 4 subcircuits, but the editor only shows two. Other failures have included red screens, or just plain old extreme lagging.

A RAM component with a configurable address width would be very nice.

satu0king:
Yes, a RAM component is needed.

satu0king:
The minimap problem might simply be an update delay. As for the lagging, try and use subcircuits as much as possible. The lagging is due to the rendering of the circuit and not the simulation of the circuit. 

Other failures have included red screens. This is disturbing, could you give a situation when this happened.

HypheX:
Me and blane have tried doing huge RAM's before, but even if we used

It would be lovely to see a bit width property and a separate "address bit width" property, allowing for the customization of the RAM's size to its full extent.

However, one has to keep in mind that the maximum bit width possible is 32, possibly due to limitations with the browsers, only allowing integers up to 2^32-1.
A fix for this is storing the states of multi-bit wires as arrays and not as integers, as it is currently done.

superstewie:
>  As for the lagging, try and use subcircuits as much as possible. The lagging is due to the rendering of the circuit and not the simulation of the circuit.

> Other failures have included red screens. This is disturbing, could you give a situation when this happened.

The largest circuit was built by using 4 eight bit registers, with a 2 control bit demux on the input and mux output. Then nesting those four at a time. In one situation, when using Firefox (I have also tried multiple browsers), I placed one of the subcircuits containing 4K bytes, then copy/pasted it once, and then when I pasted a selection of those two to create a 16K group of RAM, the redscreen occurred. Unfortunately I cannot reproduce it, and I did not get a screen shot of it.

[...]

PR #113, opened 2019-02-18
RAM component

fpintos:
This change implements a RAM component in CircuitVerse to address #81.

Two settings are available:
- addressWidth: 1 to 20, default=10. Controls the width of the address input.
- bitWidth: 1 to 32, default=8. Controls the width of data pins.

Amount of memory in the element is 2^addressWidth x bitWidth bits.
Minimum RAM size is: 2^1  x  1 = 2 bits.
Maximum RAM size is: 2^20 x 32 = 1M x 32 bits => 32 Mbits => 4MB.
Maximum 8-bits size: 2^20 x  8 = 1M x 8 bits => 1MB.
Default RAM size is: 2^10 x  8 = 1024 bytes => 1KB.
 
RAMs are volatile therefore this component does not persist the memory contents.

Changes to addressWidth and bitWidth cause data to be lost.
Think of these operations as being equivalent to taking a piece of RAM out of a 
circuit board and replacing it with another RAM of different size.

The contents of the RAM can be reset to zero by setting the RESET pin 1 or 
or by selecting the component and pressing the "Reset" button in the properties window.

The contents of the RAM can be dumped to the console by transitioning CORE DUMP pin to 1
or by selecting the component and pressing the "Core Dump" button in the properties window.
Address spaces that have not been written will show up as `undefined` in the core dump. 

NOTE: The maximum address width of 20 is arbitrary.
Larger values are possible, but in practice circuits won't need this much
memory and keeping the value small helps avoid allocating too much memory on the browser.
Internally we use a sparse array, so only the addresses that are written are actually
allocated. Nevertheless, it is better to prevent large allocations from happening
by keeping the max addressWidth small. If needed, we can increase the max.

An example of a circuit using the new component is shown below:
image

### DIFF
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -161,6 +161,7 @@
           <div class="icon logixModules" id="Keyboard" ></div>
           <div class="icon logixModules" id="Clock" ></div>
           <div class="icon logixModules" id="Rom" ></div>
+          <div class="icon logixModules" id="RAM" ></div>
         </div>
 
         <div class="panelHeader">Test Bench</div>
@@ -386,6 +387,7 @@
   <script src="/js/subcircuit.js"></script>
   <script src="/js/modules.js"></script>
   <script src="/js/sequential.js"></script>
+  <script src="/js/ram.js"></script>
   <script src="/js/canvasApi.js"></script>
   <script src="/js/data.js"></script>
   <script src="/js/listeners.js"></script>
--- a/app/views/simulator/embed.html.erb
+++ b/app/views/simulator/embed.html.erb
@@ -72,6 +72,7 @@
  <script src="/js/subcircuit.js"></script>
  <script src="/js/modules.js"></script>
  <script src="/js/sequential.js"></script>
+ <script src="/js/ram.js"></script>
  <script src="/js/canvasApi.js"></script>
  <script src="/js/data.js"></script>
  <script src="/js/embedListeners.js"></script>
--- a/public/img/RAM.svg
+++ b/public/img/RAM.svg
@@ -1,0 +1,27 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="70" height="70" version="1.1">
+    <defs/>
+    <rect width="60" height="40" x="5" y="15" fill="white" stroke="black"/>
+    <g fill="black" font-family="Georgia" font-size="6px" text-anchor="middle">
+        <text x="35" y="37">
+            RAM
+        </text>
+        <text x="11" y="27">
+            A
+        </text>
+        <text x="11" y="37">
+            DI
+        </text>
+        <text x="11" y="47">
+            W
+        </text>
+        <text x="58" y="37">
+            DO
+        </text>
+    </g>
+    <g fill="green">
+        <circle cx="5" cy="25" r="2"/>
+        <circle cx="5" cy="35" r="2"/>
+        <circle cx="5" cy="45" r="2"/>
+        <circle cx="65" cy="35" r="2"/>
+    </g>
+</svg>
--- a/public/js/UX.js
+++ b/public/js/UX.js
@@ -48,11 +48,13 @@ $(document).ready(function() {
 
     }
     $('.logixModules').hover(function() {
-        if (!help[this.id]) return;
+        // Tooltip can be statically defined in the prototype.
+        var tooltipText = window[this.id].prototype.tooltipText || help[this.id];
+        if (!tooltipText) return;
         $("#Help").addClass("show");
         $("#Help").empty();
         ////console.log("SHOWING")
-        $("#Help").append(help[this.id]);
+        $("#Help").append(tooltipText);
     }); // code goes in document ready fn only
     $('.logixModules').mouseleave(function() {
         $("#Help").removeClass("show");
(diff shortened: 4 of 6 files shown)


satu0king:
Looks great! will have to test it thoroughly. 

Minor feature request, would it possible to show the address and data value on the RAM component itself. 

I think this would be great for debugging, what do you say? Something like this. 
image

This feature request can be a separate issue if required.

fpintos:
I suppose we can show these values in the RAM itself; on the other hand, one could use a Tunnel for that.

satu0king:
> I suppose we can show these values in the RAM itself; on the other hand, one could use a Tunnel for that:
> image

Yes, we can get around it with flags. The RAM looks great as it is now. I will check and merge it soon.

jbox144:
> NOTE: The maximum address width of 20 is arbitrary.
Larger values are possible, but in practice circuits won't need this much

Is this a challenge?

fpintos:
> > NOTE: The maximum address width of 20 is arbitrary.
> > Larger values are possible, but in practice circuits won't need this much
> 
> Is this a challenge?

Hehe...it is an assumption, based in absolutely no data at all. :-)

It might be interesting to try and see how big we can go.
I noticed that I can't create an Array with 2^32 entries, so max addressWidth can't be 32:
```js
var a = new Array(Math.pow(2,32))
VM74:1 Uncaught RangeError: Invalid array length
    at <anonymous>:1:9
```

One less address would be ok:
```js
new Array(Math.pow(2,32)-1).length == 4294967295
true
```

So let's assume we could do 2^31 instead.
How long would it take to fill up such an array?
CircuitVerse clock can be set at most to 50ms, so assuming we can write a new entry every clock cycle it would take 107374182400ms, ie, about 1243 days, or 3-and-a-half years to fill up that RAM. :-)

But the other problem is that we can't even fill up a 2^31 array in JavaScript.
I've tried this:
```js
console.time();
var a=new Array(Math.pow(2,<max bit width you are willing to try>));
for(var i=0; i<a.length; i++) { a[i] = 0xFFFFFFFF }
console.timeEnd();
```

In my machine things were happy all the way to 26 bits.
Nodejs ran out of heap when I tried with 27 and Chrome crashed.
With 28 bits, Edge crashed as well.

So, assuming we were to go with 26, we would still need 64 arrays to support the full 32-bit address space. If we try to fill up multiple 2^26 arrays to see how far we can go....
```js
console.time(); 
var x=new Array(64); 
for(var j=0; j<x.length; j++) { 
    console.log(j);
    var a=new Array(Math.pow(2,26)); 
    for(var i=0; i<a.length; i++) { a[i] = 0xFFFFFFFF }
    x[j] = a; 
}
console.timeEnd();
```
Node blows up while filling up the 2nd array - maybe it can go further if I tweak the settings...
Both Chrome and Edge crashed on the 6th array.

So, it seems pretty reasonable to say people shouldn't be trying to fill up such large arrays in the simulator. It would be nasty to run a simulation for 3 years only for it to blow up because we run out of memory. 

That being said, if someone needs to create an address space larger than 2^20 (as currently supported) they can multiplex several RAMs.

If nothing else, I have a newfound appreciation for our physical machines and their plenty of RAM and storage.
:-)

jbox144:
I was kinda hoping the RAM component would look similar to the ROM component, meaning you could read the values directly. If I get around to it, I'll make a new RAM component complete with visual feedback, and a ROM component with custom bitwidth

fpintos:
This is something we can solve with a DMA pin - suppose this RAM exports a single "direct memory access" pin; then other DMA-aware components could connect to this RAM and directly read/write the memory in bulk.
For example, I can think of DMA components like this:
- a FileWriter component, which saves the contents of the RAM to the host OS.
- a FileReader component, which takes the contents of a file and populates the RAM with it
- a graphics card, this would connect the RAM to a bunch of LEDs
- a memory viewer - this is your example, this would show the contents of the RAM;
  - but then there is the question of how to format; 
  - there can be one that formats the data as ASCII, another that does HEX, etc.

Having this "DMA" pin allows us to decouple the RAM from what you want to do with that RAM.

satu0king:
> Having this "DMA" pin allows us to decouple the RAM from what you want to do with that RAM.

@fpintos I like this idea. 

> This is something we can solve with a DMA pin - suppose this RAM exports a single "direct memory access" pin; then other DMA-aware components could connect to this RAM and directly read/write the memory in bulk.
> For example, I can think of DMA components like this:
> 
> * a FileWriter component, which saves the contents of the RAM to the host OS.
> * a FileReader component, which takes the contents of a file and populates the RAM with it
> * a graphics card, this would connect the RAM to a bunch of LEDs
> * a memory viewer - this is your example, this would show the contents of the RAM;
>   
>   * but then there is the question of how to format;
>   * there can be one that formats the data as ASCII, another that does HEX, etc.
> 
> Having this "DMA" pin allows us to decouple the RAM from what you want to do with that RAM.

I really like this idea. It makes sense to have modular components which can work together.

satu0king:
> * addressWidth: 1 to 20, default=10. Controls the width of the address input.

Is there a specific reason the default is 10? Is it to make the default memory 1KB? I think a default of 8 bits make more sense.

satu0king:
That was some high quality code! Thank you for the PR. I like the improvement to tooltip.

fpintos:
No particular reason; I suppose I was looking at 1K as a nice round number.
😊


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

Issue #107, opened 2019-02-15
Component selection size issues


jbox144:
There are a number of components in Circuit Verse which have a selection area which extends outside their graphical size, or which has their selection area overlap their pins. These components are often troublesome, to work with, and need to be fixed.

For sake of time, I will refer to the selection area as the "Hit Box"

- [x] Decoder, Multiplexer, Demultiplexer
When the Decoder, Multiplexer or Demultiplexer is given a selection bitwidth of 1, the graphic of the component is smaller than it would be when at a bitwidth of 2 or more. At a larger bitwidth, these components are 4 units wide, and vary in height depending on their settings. However, when there is a bitwidth of 1, these components appear 2 units wide, but still retain the 4 unit wide hit box they had before.

- [x] Power and ground
The power and ground pin's hit box is straight up too big. This should otherwise be an easy fix.

fpintos:
It looks like the issue is not only with the component dimension but also with the rounding on the simulation.mouseX and mouseY; it makes the component think it is being hovered even when the mouse is half-a-square away.

PR #119, opened 2019-02-21
Fix component sizes

fpintos:
Fix for #107 

The rounded values of mouseX and mouseY caused inHover to return true too soon, effectively making the hitbox half-a-square larger in each side of the component.

I've included non-rounded values in the simulationArea and use those to determine the hit.

Then I reduced the hitbox of Power & Ground to be just the 20x20 square.
I've also fixed the size of the 1-bit Decoder, Mux & Demux.

### DIFF
--- a/public/js/logix.js
+++ b/public/js/logix.js
@@ -1102,10 +1102,8 @@ CircuitElement.prototype.fixDirection = function() {
 // NOT OVERIDABLE
 CircuitElement.prototype.isHover = function() {
 
-    var mouseX = simulationArea.mouseX;
-    var mouseY = simulationArea.mouseY;
-
-    if (Math.abs(mouseX - this.x > Math.max(this.leftDimensionX, this.rightDimensionX)) || Math.abs(mouseY - this.y > Math.max(this.upDimensionY, this.downDimensionY))) return false;
+    var mX = simulationArea.mouseXf - this.x;
+    var mY = this.y - simulationArea.mouseYf;
 
     var rX = this.rightDimensionX;
     var lX = this.leftDimensionX;
@@ -1129,9 +1127,7 @@ CircuitElement.prototype.isHover = function() {
         }
     }
 
-    if (mouseX - this.x <= rX && this.x - mouseX <= lX && mouseY - this.y <= dY && this.y - mouseY <= uY) return true;
-
-    return false;
+    return -lX <= mX && mX <= rX && -dY <= mY && mY <= uY;
 };
 
 CircuitElement.prototype.setLabel = function(label) {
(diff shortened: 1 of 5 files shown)


jbox144:
Damnit, I was too slow :P

Too be fair, I don't even know how I am supposed to debug anything.

Never worked on servers before, this is all new to me. I'm far more used to working on local applications...

satu0king:
> Damnit, I was too slow :P
> 
> Too be fair, I don't even know how I am supposed to debug anything.
> 
> Never worked on servers before, this is all new to me. I'm far more used to working on local applications...

You do not have to worry about all parts of the code. Setup docker on your system. Then run the docker instructions mentioned, this way your dev env is setup. 

Then see public/js folder for all the code. It is a bit low on documentation right now, however you should be able to get around.

fpintos:
> > Damnit, I was too slow :P
> > Too be fair, I don't even know how I am supposed to debug anything.
> > Never worked on servers before, this is all new to me. I'm far more used to working on local applications...
> 
> You do not have to worry about all parts of the code. Setup docker on your system. Then run the docker instructions mentioned, this way your dev env is setup.
> 
> Then see public/js folder for all the code. It is a bit low on documentation right now, however you should be able to get around.

In case it helps...

If you go thru the steps of installing postgres, ruby & rails then you can run the application and work on it. Just make sure you have the database.yml configured correctly. In principle, you don't even need docker/redis and other production-only stuff.

These two posts were super helpful to get me started, since this was my 1st time doing ruby:
- https://www.howtoforge.com/tutorial/ubuntu-ruby-on-rails/
- https://gorails.com/setup/windows/10
- and one more for good measure: https://medium.com/koaandco/rails-5-postgresql-on-wsl-9b1315ac0679

I've tried both setups, Ubuntu and Windows, and they both work just fine.

public/js/logix.js
```diff
@@ -1102,10 +1102,8 @@ CircuitElement.prototype.fixDirection = function() {
 // NOT OVERIDABLE
 CircuitElement.prototype.isHover = function() {
 
-    var mouseX = simulationArea.mouseX;
-    var mouseY = simulationArea.mouseY;
-
-    if (Math.abs(mouseX - this.x > Math.max(this.leftDimensionX, this.rightDimensionX)) || Math.abs(mouseY - this.y > Math.max(this.upDimensionY, this.downDimensionY))) return false;
```
satu0king: I wonder what this snippet of code was doing. @fpintos did it do anything useful?

satu0king:
Great! everything checks out. If there are any bugs, please report them

public/js/logix.js
fpintos: I believe the intention was to shortcut the calculation if the mouse was not inside a bounding rectangle around the element; but the expression was in fact incorrect, since the closing parenthesis for Math.abs was in the wrong place.
In the end, the final expression does the same with the correct values, so this was not needed at all.

Issue #127, opened 2019-02-26
Add pry-rails for easy debugging


ankitkataria:
`pry-rails` - https://github.com/rweng/pry-rails is a very useful plugin for live debugging the backend. 
Addition of pry will make the whole development process very easy. 

`binding.pry` - can be inserted anywhere inside the code, rails will open a console and stop execution when this statement will be encountered, then the state of various variables can be checked. 

For example: Adding `binding.pry` to `show` of `groups_controller` will show - 
![image](https://user-images.githubusercontent.com/22021150/53451037-d3ac1200-3a43-11e9-9721-7afa80a01d31.png)

Then the various variables can be inspected.

PR #128, opened 2019-02-26
Add pry-rails dependency to Gemfile

ankitkataria:
Solves #127 
`pry-rails` dependency has been added to the Gemfile. More details have been included in the issue.


ankitkataria:
Thank you for the suggestions @tachyons! I've implemented the changes.

PR #149, opened 2019-02-28
Additional ubuntu instructions

Ankitanvesh:
I experienced some errors while setting up the project in my local machine so I have added a few lines 
under the **Additional Setup instructions for Ubuntu users** in the README.md file in order to assist other developers in smooth setting up of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ Email: admin@circuitverse.org
 Password: password
 ```
 
+For debugging include `binding.pry` anywhere inside to code to open the `pry` console.
+
 ## Additional setup instructions for Ubuntu
 Additional instructions can be found [here](https://www.howtoforge.com/tutorial/ubuntu-ruby-on-rails/) and there are some extra notes for single user installations:
 - If setting up Postgres with these instructions, use your user name instead of 'rails_dev'.
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 ## Setup Instructions
 
+**Note:** You might want to use the docker instructions if you do not want to setup your own environment. 
+
 * Install ruby using RVM, use ruby-2.5.1
 * Install Dependencies: `bundle install --without production`
 * Configure your DB in config/database.yml, copy config/database.yml.example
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -70,6 +70,13 @@ production:
   password: <%= ENV['circuitverse_DATABASE_PASSWORD'] %>
 ```
 
+- If you are facing difficulties installing RVM (which can be the    case with most users on ubuntu trying to install rvm) try using
```
armudgal: Hi, can the bracket message be trimmed to `(mostly on Ubuntu machine)` or something of that sort cause `trying to install rvm` seems a little repetitive. 
Also, in one line can you also explain how this is solving the problem.

README.md
```diff
@@ -70,6 +70,13 @@ production:
...
+ this command ` \curl -L https://get.rvm.io | 
+  bash -s stable --ruby --autolibs=enable --auto-dotfiles`
+- If you are facing errors running the `rails db:create` ensure       that the socket file(i.e mysql.sock) is present in that location.
+  Some possible locations where it might be present is /run/mysqld/mysqld.sock  or /var/lib/mysql/mysql.sock and mention
```
armudgal: Please highlight the path using grave accent to distinguish them from the text.

README.md
Ankitanvesh: Ok I will do it

README.md
```diff
@@ -70,6 +70,13 @@ production:
   password: <%= ENV['circuitverse_DATABASE_PASSWORD'] %>
 ```
 
+- If you are facing difficulties installing RVM which is primarly     because Ubuntu's desktop edition ships with a version of rvm which is not the exact rvm version hence before installing it(RVM) you need to get rid of this inbuilt version
```
ankitkataria: What is meant by "exact version" of rvm? 
Also is it possible to break this line into smaller ones?

Ankitanvesh: By that I meant to imply that it is an older version

Ankitanvesh: I will try to shorten the line thanks for the feedback

Issue #141, opened 2019-02-27
SHOWING-"This page isn’t working"


akoopraba4:
I am not sure if this is correct or not.If we press the back button in new groups without entering the group name it shows 
This page isn’t working 
If the problem continues, contact the site owner.
HTTP ERROR 406
https://circuitverse.org/groups/new
Please check this out

satu0king:
Yes it is a bug.

akoopraba4:
I would like to work on this issue.However @satu0king  can you provide some information on how to go about it?

PR #153, opened 2019-02-28
fixed SHOWING-'This page isn’t working'

shubhankarsharma00:
with respect to issue #141

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -345,4 +345,4 @@ DEPENDENCIES
   will_paginate-bootstrap
 
 BUNDLED WITH
-   1.17.2
+   1.17.3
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -345,4 +345,4 @@ DEPENDENCIES
   will_paginate-bootstrap
 
 BUNDLED WITH
-   1.17.2
+   1.17.3
```
ankitkataria: Should there be a particular version of `bundler` to be used mentioned in the readme? Like we have versions mentioned for `rails`, `ruby` and `postgres`. Changing the bundler version in the gemfile might create problems for other users.

tachyons: Bundler versions hardly cause any issues.

tachyons: Except when it is major release. Like bundler 1.x and bundler 2.x

shubhankarsharma00:
Is the PR okay? Why isn't it approved @tachyons is there something wrong?

PR #158, opened 2019-03-01
Additional ubuntu instructions

Ankitanvesh:
I experienced some errors while setting up the project in my local machine so I have added a few lines
under the **Additional Setup instructions for Ubuntu users** in the README.md file in order to assist other developers in smooth setting up of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ production:
   password: <%= ENV['circuitverse_DATABASE_PASSWORD'] %>
 ```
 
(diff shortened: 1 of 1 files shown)


tachyons:
Better keep these stuff in trouble shooting wiki page

Ankitanvesh:
I have included it in the readme cause I think that for those setting up the project it would then be a decent head start to have their issues sorted just by reading the Readme.

Ankitanvesh:
Besides the readme should be as informative as it can be I guess

Issue #142, opened 2019-02-27
Migrate tooltips


satu0king:
Currently, tooltips information are stored at UX.js, migrate the same to individual circuit elements. Check out how Power and Ground implement Tooltips.

rkpattnaik780:
I would like to work on this issue.

armudgal:
Hi @rkpattnaik780, can you please if you are working on this issue

sarthak77:
Hi I would like to take up this issue if not taken

sanctophagus:
Hey @satu0king !
If no one is working on this issue, than I am willing to work on it!

armudgal:
@sarthak77 Would you be taking up this issue?

sarthak77:
ok i'll be working on it

armudgal:
Hi @satu0king, the issues are based on first come first serve, so sorry for that, please look through other issues. Cheers

agarwalrounak:
@armudgal I think you tagged the wrong person :sweat_smile:

sarthak77:
Yeah I'm working on it

sarthak77:
> Currently, tooltips information are stored at UX.js, migrate the same to individual circuit elements. Check out how Power and Ground implement Tooltips.

Hey I could not relate to what do you mean by Power and Ground
Can you elaborate?

satu0king:
> > Currently, tooltips information are stored at UX.js, migrate the same to individual circuit elements. Check out how Power and Ground implement Tooltips.
> 
> Hey I could not relate to what do you mean by Power and Ground
> Can you elaborate?

Look for this line in modules.js: 
`Power.prototype.tooltipText = "Power: All bits are High(1).";`

This is how we want all modules to implement tooltips now.

satu0king:
@sarthak77 since we haven't heard from you, I am releasing this issue. Please feel free to claim the issue again if you are still working on this.

sarthak77:
I was having trouble in setting up the system and had exams in between so was out for a while

mohamedar97:
i want to give this issue a try, can i take it?

prajapatgautam1999:
@satu0king @armudgal Is anyone working on this issue as it is not marked "taken" but there are few who seem to be working for the same.

satu0king:
This is a very easy issue, therefore the first PR will be merged.

armudgal:
@prajapatgautam1999 You can go ahead since we didn't receive any PR till now

[...]

PR #170, opened 2019-03-03
Add view and logic for Sixteen Segment display

### LINKED ISSUE #101: 16 Segment Display
Add a 16 Segment Display element

A 16 segment display element would be very useful for displaying complex characters with very little ROM space

sanctophagus:
Closes: #101 

This sixteen segment still does not look that aesthetically pleasing, I am currently working to improve it's looks

#### Describe the changes you have made in this pr
Add svg and logic for Sixteen Segment display

#### This is how it look on simulator
screenshot from 2019-03-06 04-39-41


satu0king:
@blane1257 @fpintos, do you think if this is implemented as a 16 bit input instead of 16 separate inputs, it will be better?

fpintos:
Indeed, using a 16bit input will make it simpler to use in most cases. The design with individual pins makes it virtually impossible to put these side by side without large gaps and lots of wires.

blane1257:
Yes, I think a 16 bit input is the way to go.

blane1257:
I also think the display should be made horizontally thinner, to more closely resemble its 7 segment counterparts

satu0king:
@sanctophagus can you replace the 16 individual pins and replace it one 16 bit input?

fpintos:
If we do go with the single input we will need to have it clearly documented so people know which bit corresponds to what pin.

For the HexDisplay this is not much of an issue, since one expects the hex mapping anyway, but in this case it would not be immediately clear.

sanctophagus:
@blane1257 Sure I will work to make it thinner.
@satu0king  @fpintos  I will make it one 16 bit input and document it, So that users find it easy to use.

sanctophagus:
I am thinking of converting it to sixteen+one bit input box (segment + dot). Currently I have 17 bit input, how should I proceed?
Also now it resembles its 7 Segment Counterpart!
sixteenbit

blane1257:
that looks good! you should have the 16 bit input on the top, similar to the HEX display, with the decimal input on the bottom

sanctophagus:
Okay!! 
I will make it that way!

fpintos:
It is looking very nice.

public/img/SixteenSegDisplay.svg
```diff
@@ -0,0 +1 @@
...7v34h-7zm-197,36h34v7h-34zm46,0h34v7h-34z"/><path d="M163,250h34v7h-34zm46,0h34v7h-34zm46,0h11v7h-11z"/></g><g style="fill:rgb(6,149,24);fill-rule:evenodd;stroke:rgb(6,149,24);stroke-width:0.1;" id="l4"><path d="M62,17h6l2,1l1,1l1,3v6l-3,3h-8l-3-3v-6l1-3l1-1zm46,0h6l2,1l1,1l1,3v6l-3,3h-8l-3-3v-6l1-3l1-1zm46,0h6l2,1l1,1l1,3v6l-3,3h-8l-3-3v-6l1-3l1-1zm46,0h6l2,1l1,1l1,3v6l-3,3h-8l-3-3v-6l1-3l1-1z"/><path d="M246,17h6l2,1l1,1l1,3v6l-3,3h-8l-3-3v-6l1-3l1-1zm-206,46h6l3,3v8l-3,3h-6l-3-1l-1-1l-1-2v-6l1-2l1-1zm230,0h4l3,1l1,1l1,2v6l-1,2l-1,1l-3,1h-4l-3-1l-2-2v-8l2-2zm-230,69h6l3,3v8l-3,3h-6l-3-1l-1-1l-1-2v-6l1-2l1-1z"/><path d="M270,132h4l3,1l1,1l1,2v6l-1,2l-1,1l-3,1h-4l-3-1l-2-2v-8l2-2zm-230,69h6l3,3v8l-3,3h-6l-3-1l-1-1l-1-2v-6l1-2l1-1zm230,0h4l3,1l1,1l1,2v6l-1,2l-1,1l-3,1h-4l-3-1l-2-2v-8l2-2zm-209,46h8l2,2l1,3v4l-1,3l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3z"/><path d="M107,247h8l2,2l1,3v4l-1,3l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3zm46,0h8l2,2l1,3v4l-1,3l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3zm46,0h8l2,2l1,3v4l-1,3l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3zm46,0h8l2,2l1,3v4l-1,3l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3z"/><path d="M270,247h4l3,1l1,1l1,2v6l-1,2l-1,1l-2,1h-6l-2-1l-1-1l-1-3v-4l1-3l1-1z"/></g></svg>
```
fpintos: Remember to update the svg icon to reflect the new layout.
If appropriate, feel free to use rect and circle; they make the svg smaller and more understandable as well.

public/js/UX.js
```diff
@@ -84,6 +84,7 @@ var help = {
     "VariableLed": "Variable Led ToolTip: Variable LED inputs an 8 bit value and glows with a proportional intensity.",
     "HexDisplay": "Hex Display ToolTip: Inputs a 4 Bit Hex number and displays it.",
     "SevenSegDisplay": "Seven Display ToolTip: Consists of 7+1 single bit inputs.",
+    "SixteenSegDisplay": "Sixteen Display ToolTip: Consists of 16+1 single bit inputs.",
```
fpintos: There is an open issue (#142) to move these strings out of this file.
You can make this string part of the element's prototype (in the element's file itself), for example:
```js
SixteenSegDisplay.prototype.tooltipText = "Consists of 16+1 single bit inputs.";
```

public/js/modules.js
```diff
@@ -576,6 +576,117 @@ SevenSegDisplay.prototype.customDraw = function() {
     ctx.stroke();
 }
 
+function SixteenSegDisplay(x, y, scope = globalScope) {
```
fpintos: Let's put this into its own file, in which case you'll update a couple of other files to include it.
See #199

public/img/SixteenSegDisplay.svg
sanctophagus: Sure! I'll do that.

sanctophagus:
@satu0king @fpintos @blane1257 I've updated the PR, please review

public/js/UX.js
fpintos: You can revert the change in this file.

public/js/modules.js
```diff
@@ -576,6 +576,112 @@ SevenSegDisplay.prototype.customDraw = function() {
...
+    ctx.lineWidth = correctWidth(3);
+    xx = this.x;
+    yy = this.y;
+    moveTo(ctx, x1, y1, xx, yy, this.direction);
+    lineTo(ctx, x2, y2, xx, yy, this.direction);
+    ctx.closePath();
+    ctx.stroke();
+}
+
+SixteenSegDisplay.prototype.customDraw = function() {
```
fpintos: One optimization idea, just to keep the code more compact and slightly faster, would be to define an array with the coordinates and just extract the bits using `this.input.value & (1<<n)` (undefined coerces to zero automatically). This would avoid converting the number to string, then splitting it, then mapping to number, then assigning to multiple variables.

I might look like this:
```js
var value = this.input.value;
var color = ["lightgrey", "red"];
[
    [-20, -38, 0, -38],
    [20, -38, 0, -38],
    ... (and the others as you have in the calls to customDrawSegment & customDrawSegmentSlant)
].forEach(function (c,i){
if (i<12) {
     this.customDrawSegment(c[0],c[1],c[2],c[3], color[value & (1 << i)]);
} else {
    this.customDrawSegmentSlant(...same here...)
}
});
```

public/js/modules.js
satu0king: @fpintos let us not do this as of yet for non-critical modules like these. I know it is better to have these in separate files and then bundle them together, wouldn't it be better to separate them once we setup webpack and other tools?

satu0king:
> It looks good to me; I would still suggest moving the new element to a new file, and you can revert the change in UX.js.

I think we can delay this for now. We can structure the codebase and refactor everything in a separate issue. What do you say @fpintos ?

public/js/modules.js
satu0king: If you do not want to use a loop it is fine, the code is still readable. I think it is a good idea to use a bitwise operation to perform slight optimization. However considering the amount of rendering we are doing, this hardly seems an overhead. Moreover, dec2bin internally uses `.toString()` which probably uses bitwise operations. However, this line seems unnecessary as you can directly index on the string and use that instead. 
`var arr = (binNum).toString(10).split("").map(Number);` as `binNum` is a string and can already be  accessed via index.

satu0king:
Make the optimization changes @fpintos suggested and we are good to go. Lets keep the function in `modules.js` for now. We will have to undergo a complete refactor soon, we will handle it then. After the PR is merged, @sanctophagus raise an issue on https://github.com/CircuitVerse/CircuitVerseDocs regarding documenting the module.

fpintos:
Sounds good.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

public/js/modules.js
sanctophagus: @fpintos Actually the expression `[value & (1 << i)]` returns a value in decimal format and we need either a 0 or 1 to  get the correct results. Can you please guide me, how should I proceed.

fpintos: Sorry...make it `color[value & (1<<i) != 0]` and it should work.

satu0king: Or `color[ (value>>i) & 1 ]` could work as well ?

sanctophagus:
@satu0king @fpintos @armudgal Please review this PR. I've made the requested changes.

armudgal:
Please update this branch with our master.

sanctophagus:
@armudgal I have rebased the branch on master.

armudgal:
@satu0king Hey, this is ready for merge. I would want you to take the final look and merge it. Thanks

sanctophagus:
@satu0king @fpintos may I know your views on this one?

PR #172, opened 2019-03-03
New Element: Counter

### LINKED ISSUE #67: "Counter" component
See this project for an example: https://circuitverse.org/users/1577/projects/4036

Essentially, the counter would take in a clock signal, and output a signal which is incremented by 1 every time it receives a clock pulse.

fpintos:
Fixes #67

# Counter

The **Counter** element produces a sequence of binary numbers, from zero to a specified maximum value. 

## Properties
- **bitWidth**: controls the number of bits of the `Value` and `MaxValue` pins.

## Inputs
- **MaxValue**: if connected, defines the inclusive maximum `Value`; otherwise the maximum `Value` is determined by `bitWidth`.
- **Clock**: on the _transition_ from 0 to 1, `Value` is either incremented or set to 0.
- **Reset**: when set to 1, `Value` is set to 0.

## Outputs
- **Value**: the current value of the counter.
- **Zero**: a flag that indicates when `Value` is zero and `Clock` is 1. This can be used to chain Counters together, much like a ripple counter built from flip-flops.

## Screenshots:
This diagram illustrates the pins in the Counter component (highlighted):
image

This is an example of building some sort of "digital clock" using the Counter component:
image

The first two counters on top are counting "minutes", the first goes from 0 to 9, the second from 0 to 5, so together they count from 0 to 59, and their outputs directly drive the HexDisplays.
The bottom counters are counting the hours; the 3rd counts from 0 to 9 when the 4th is zero or 1, and from 0 to 3 when the 4th is 2, so together they count from 0 to 23.

## Review Notes

The Counter can be found under the "Inputs" group.

> One can make an argument that it should be in the Sequential Elements group, since the corresponding circuitry would be mostly based on flip-flops. I chose Inputs because the Counter ends up working more like a source of numbers, like other elements in Inputs do, in particular the Random number generator.

A corresponding 3-bit, 0-7 counter looks like this:
image
Note I'm not including the MaxValue circuitry in the example above.
Samples with MaxValue can be seen in this project: https://circuitverse.org/users/1935/projects/4801

### DIFF
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -116,6 +116,7 @@
           <div class="icon logixModules" id="ConstantVal" ></div>
           <div class="icon logixModules" id="Stepper" ></div>
           <div class="icon logixModules" id="Random" ></div>
+          <div class="icon logixModules" id="Counter" ></div>
         </div>
 
         <div class="panelHeader">Output</div>
@@ -389,6 +390,7 @@
   <script src="/js/modules.js"></script>
   <script src="/js/sequential.js"></script>
   <script src="/js/ram.js"></script>
+  <script src="/js/Counter.js"></script>
   <script src="/js/canvasApi.js"></script>
   <script src="/js/data.js"></script>
   <script src="/js/listeners.js"></script>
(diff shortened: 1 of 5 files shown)


armudgal:
This PR is amazing, thanks @fpintos. I would wait for @satu0king to review this :)

jbox144:
It's perfect! Thanks!

satu0king:
Awesome! Thanks.

PR #179, opened 2019-03-03
Modify bootstrap alerts for a better UX

### LINKED ISSUE #180: Auto hide bootstrap alerts and change alert width
When creating a new group or adding a new assignment, an alert showing **Group was successfully created.** OR **Assignment was successfully created.** is displayed above the group name. This alert stays there on the page.
For a better UX, this alert should auto-hide after a certain amount of time. 
The alert would also look more aesthetically pleasing if it stretched across the entire container.
new group
_I've worked on this issue and I linked this issue number in the pull request._

ahmedkrmn:
Fixes #180
When creating a new group or a new assignment:
- Auto hide alert after 3 seconds.
- Stretch the width of the alert to fill the container
---
**Before**:
new group
**After**:
after
Note: Alert will auto hide after 3 seconds.


app/views/groups/show.html.erb
```diff
@@ -184,6 +184,7 @@
               this.innerHTML = formatDate(new Date(this.innerHTML), "dddd hh:mmTT zz, dd MMM yyyy")
           })
       })
+      setTimeout(()=>document.querySelector(".alert").remove(), 3000);
```
satu0king: I think these alerts are used everywhere, login, project dashboard etc. Could you confirm the same? If yes, maybe move this to `/app/views/layouts/application.html.erb`

armudgal: Please use `hide` instead of `remove`, we don't want to reload the element in DOM again if needed.
And also what @satu0king said.

ahmedkrmn: @satu0king, yeah I've just noticed that, working on it.

ahmedkrmn: @armudgal nice remark, will change it now.

ahmedkrmn:
@satu0king @armudgal I made the proposed changes, but why does it show a **Build Error** ?

armudgal:
I will test it in a few hours.

The build error is because we have CI now but the configuration file is here #186 , I believe when we will merge that PR, this build error should go.

armudgal:
Hi, can you update your branch with our master. Just want to make sure nothing breaks in the CI :) 
We can merge it then

ahmedkrmn:
> Hi, can you update your branch with our master. Just want to make sure nothing breaks in the CI :)
> We can merge it then

Done!

PR #181, opened 2019-03-03
Keep the connection point fixed when changing bitWidth of INPUT and OUTPUT elements

### LINKED ISSUE #97: Keep connection point fixed when changing bit width of input and output elements
When you change the bit width of an input or output element its position is adjusted so that the middle of the element stays the same.

This causes a disruption in the circuit layout when the connection point is to the right or the left of the element and usually requires one to go and move the element back into a better position. It can even happen that the connection point ends up on top of another line, in which case when we move it back we end up with an undesired connection.

My suggestion would be to anchor the connection point when changing the bitwidth, not the middle of the element. This way, increasing the bitwidth of an element whose connection is on the right will make it grow to the left and vice-versa; this way there will be no change in where the element is connected and therefore no chance of introducing an undesired connection.

fpintos:
Fixes #97

This PR updates the INPUT and OUTPUT elements to move themselves left or right when bitWidth is changed, so that the connection point remains fixed. This prevents the layout of the circuit from being accidentally disrupted when one changes the bitWidth of these elements.

### Current Behavior
- Connect an INPUT and an OUTPUT element.
- Change the bitWidth of these elements.
- Their central point remains fixed and wires begin to appear in undesirable places.

### New Behavior
- Connect an INPUT and an OUTPUT element.
- Change the bitWidth of these elements.
- Their connection point remains fixed; no new wires appear.

### To Test:
- increase and decrease the bitWidth, sequentially (1,2,3,...) or in jumps (1, 5, 8, etc..)
- test with the connection point on the right and on the left of the element
- note there is no change in behavior when the connection point is up or down - in this case it will remain in the middle.

### DIFF
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -1782,14 +1782,17 @@ Input.prototype.resolve = function() {
 // Check if override is necessary!!
 Input.prototype.newBitWidth = function(bitWidth) {
     if (bitWidth < 1) return;
+    var diffBitWidth = bitWidth - this.bitWidth;
     this.bitWidth = bitWidth; //||parseInt(prompt("Enter bitWidth"),10);
     this.setWidth(this.bitWidth * 10);
     this.state = 0;
     this.output1.bitWidth = bitWidth;
     if (this.direction == "RIGHT") {
+        this.x -= 10 * diffBitWidth;
         this.output1.x = 10 * this.bitWidth;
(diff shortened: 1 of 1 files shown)


satu0king:
Nice! It is amazing how such a small piece of simple elegant code can have a huge impact on the experience.

PR #187, opened 2019-03-04
adding watermark to iframes

### LINKED ISSUE #11: Add watermark with link in embed mode when embedded in external tools
If bloggers/teachers embed circuits from CircuitVerse in external places like powerpoint/blogs. A watermarks or banner such as "Made with CircuitVerse" or "View on CircuitVerse" needs to be displayed.

Amansingla97:
Fixes #11 

A watermarks "Made with CircuitVerse" is displayed on iframes on clicking to which directs to a new page in simulator.
screenshot 2019-03-04 at 3 46 19 pm


blane1257:
Maybe it should direct to the main page, instead of the simulator.

Amansingla97:
what do you mean by main page?

blane1257:
the CircuitVerse home page (https://circuitverse.org/) rather than simulator page (https://circuitverse.org/simulator)

Amansingla97:
what I thought was the person like to interact with the circuit in simulator but yes homepage seems more sensible.

satu0king:
> what I thought was the person would like to interact with the circuit in simulator but yes homepage seems more sensible.

No it should not direct to the home page either. It should direct to the project page: example https://circuitverse.org/users/126/projects/252. This way users will know it was made on CV and also they will be able to fork to the circuit, read description etc.

blane1257:
Yes, I think the project page is the best option.

satu0king:
Also the watermark should appear only when embedded outside CircuitVerse, that is when accessed via this URL signature: https://circuitverse.org/simulator/embed/704 and not this URL signature: https://circuitverse.org/simulator/704

Amansingla97:
@satu0king please review the changes

app/controllers/simulator_controller.rb
```diff
@@ -22,6 +23,9 @@ def embed
       render plain: "Project has been moved or deleted. If you are the owner of the project, Please check your project access privileges." and return
     end
     @logix_project_id = params[:id]
+    @project = Project.find(params[:id])
+    @author = @project.author_id
+    @show_value = 1
```
satu0king: This is the correct logic but I would prefer the variable to be a boolean instead of integer and the name to be @external_embed = true/false

Amansingla97: cool

satu0king:
Nice! thanks

PR #189, opened 2019-03-04
Support bitWidth inference in Output and Gates

### LINKED ISSUE #98: Automatic inference of bitwidths
Would it be possible for element to automatically infer the bitwidth of their inputs?
Perhaps the bitwidth value could have an "auto" value.

For some elements this should be simple:

- An Output element would simply pick up the bitwidth of the line it is connected to.
- Gates would automatically infer the bitwidth if all input lines are the same width and error out otherwise.
- Similar for the data input/output of decoders/multiplexers/flip-flops and others.
  - Maybe the same inference could be applied to the control signal, although this could cause trouble since a change in control signal size ends up changing the layout of decoders/multiplexers.
 - Input elements obviously need to specify their bitwidth since they themselves have no input.
  - That being said, suppose an input could be set to automatic as well; in this case an input exposed by a sub-circuit would inherit the bitwidth from the parent circuit.

This would allow us to create circuits that are more generic and reusable, since each instance could take a different set of bitwidth values.

fpintos:
Fixes #98 

# Automatic Inference of bitWidth in Output and Gate elements

This PR implements inference of bitWidth in CircuitVerse, making it easier to build circuits that use multi-bit signals.

## Example of current behavior

- Drop a couple of 4-bit Input elements and connect them to a default AND gate.
  - You are greeted with a message saying there is a bitWidth error.
- Fix the bitWidth of the AND gate.
- Connect an Output element to its output.
  - By default, Output comes as 1-bit and again you are greeted with another error message. 
- Now decide that you need the Input elements to be 8-bits.
  - You need to select each input, gate and output and fix their bitWidth to 8-bits.

## Example of new behavior

- Drop a couple of 4-bit Input elements and connect them to a default AND gate.
  - The AND gate automatically infers it should be 4-bits and fixes itself.
- Connect an Output element to the AND gate output.
  - In a similar manner, the Output realizes its input is a 4-bit signal and it fixes itself.
- Now decide that you need the Input elements to be 8-bits.
  - As you change both Inputs, all connected gates and outputs automatically change to 8-bits as well.
    > This might depend on the complexity of the circuit. See below.
- If a particular gate has a conflict, like one input asking for 4-bits and another asking for 5-bits, the circuit will report an error as usual and has to be manually fixed.

### Screen-capture of new behavior:
[Screen-capture of new behavior](https://www.youtube.com/watch?v=fYoIJ0YPUeI)
[CircuitVerse - Digital Circuit Simulator online - Google Chrome 2019-03-04 02-41-16.zip](https://github.com/CircuitVerse/CircuitVerse/files/2926025/CircuitVerse.-.Digital.Circuit.Simulator.online.-.Google.Chrome.2019-03-04.02-41-16.zip)

## Known issues

- At this point the fix targets Output and Gate elements.
  Other elements can be updated in future PRs to benefit from this mechanism using similar principles.

- Inference is currently limited to elements in a single circuit, ie, bitWidth does not flow into a sub-circuit.

- In complex circuits, not all gates might update in response to the change in bitWidth of an Input.
  - I've been trying to get the circuit to fix itself in most/all cases.
  - However, in some cases the bitWidth propagation in the wires does not seem to go all the way from one element to another. In those cases, changing an existing Input bitWidth might not cause all elements to fix themselves.
  - The error is still reported and can be fixed as usual.
  - In some cases, it is sufficient to change the value of an Input to cause its changes to propagate completely.
  - In other cases, it is a matter of fixing one offending wire and then all others fix themselves.
  - I'm looking for ways to eliminate this problem and have inference consider all nodes consistently.

### DIFF
--- a/public/js/Node.js
+++ b/public/js/Node.js
@@ -283,6 +283,16 @@ Node.prototype.resolve = function() {
 
         let node = this.connections[i];
 
+        // If we detect a bitWidth mismatch, give the parent element a chance to correct itself.
+        if (node.bitWidth != this.bitWidth) {
+            if (node.parent.onBitWidthError) {
+                node.parent.onBitWidthError(node, this.bitWidth);
+            }
+
+            // And then force recalculation of this particular node
+            node.value = undefined; 
+        }
+
         if (node.value != this.value) {
 
             if (node.type == 1 && node.value != undefined && node.parent.objectType != "TriState" && !(node.subcircuitOverride && node.scope != this.scope)) {
@@ -300,6 +310,8 @@ Node.prototype.resolve = function() {
 
                 node.bitWidth = this.bitWidth;
                 node.value = this.value;
+                this.highlighted = false;
+                node.highlighted = false;
                 simulationArea.simulationQueue.add(node);
             } else {
                 this.highlighted = true;
@@ -380,17 +392,14 @@ Node.prototype.draw = function() {
                 x: this.absX(),
                 y: this.absY() - 15
             }
-            if (this.type == 2) {
-                var v = "X";
-                if (this.value !== undefined)
-                    v = this.value.toString(16);
-                if (this.label.length) {
-                    canvasMessageData.string = this.label + " : " + v;
-                } else {
-                    canvasMessageData.string = v;
-                }
-            } else if (this.label.length) {
-                canvasMessageData.string = this.label;
+
+            // Show value, bitWidth and id on the tooltip of every node.
+            var v = (this.value !== undefined) ? this.value.toString(16) : "X";
+            var nodeTooltip = v + " (" + this.bitWidth + " bit" + (this.bitWidth == 1 ? "" : "s") + ", " + this.id + ")";
+            if (this.label.length) {
+                canvasMessageData.string = this.label + " : " + nodeTooltip;
+            } else {
+                canvasMessageData.string = nodeTooltip;
             }
         } else {
             setTimeout(function() {
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -12,6 +12,28 @@ function changeInputSize(size) {
 
 }
 
+/**
+ * Handler to update a gate's bitWidth when all connected input nodes have the same expected bitWidth.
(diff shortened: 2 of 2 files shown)


fpintos:
I haven't implemented the global/per-project switch to enable this feature yet.

satu0king:
I think this is a much-needed feature, However, I am not very optimistic about this yet. I think this would be highly error-prone and can cause a lot of issues. For example, if I add input of 3 bits directly to an output of 4 bits, what should change the input or the output? I guess in the method currently implemented there is no way for the input to know that it has to change so only output will change, that is those receiving the inputs will change. Also, I think there can be more trickier cases, like assume that there is a cyclic circuit which can trigger an infinite loop of bitwidth changes. 

I suppose maybe we can implement this with non-cyclic elements single node elements like outputs, some inputs and maybe some gates.

@fpintos what do you say?  I think we should definitely implement this with more discussion.

fpintos:
> @fpintos what do you say? I think we should definitely implement this with more discussion.
 
Oh, absolutely. This one deserves O LOT of attention. :-)
I should probably have clarified that this was more of a "look, it is not as hard as we imagined" PR.

I believe any solution we come up with should be unidirectional to avoid cycle-problems. From inputs to outputs seems like the natural direction to follow. So, in your examples above, if you add a 3-bit input to a 4-bit output, the output will change, never the input, by design.

This PR is likely doing that, but I need to investigate the cycles. There is a level of convergence to it - it will only ask an element to change its bitWidth if it detects a difference and the contract is that if an element allows to have its bitWidth changed then it must do so as requested and it will automatically propagate the change to its connected elements. With this in mind, I don't _think_ there would be a way for bitWidths to circle back with a different value, otherwise it would be a circuit design error with or without this PR. But indeed, we need to validate this; it would be nice to add a test suite for the simulator itself in order to enforce these behaviors.

I already found some cases where not everything updates automatically all the time; in one case I had a cascade of gates with one input feeding several gates and for some reason some nodes would not be scheduled to be updated - I wonder if the simulation is stopping as soon as it hits the first non-resolvable error and then depending on the order in which they evaluate one is left-over, maybe that's the case.

PR #197, opened 2019-03-05
Bug fix

### LINKED ISSUE #96: Stepper does not work in Preview/Embed mode
Steppers do not seem to be responding to keyboard events in the preview window, even thou they appear selected when you click on them.

Repro:
- Navigate to a project that uses steppers, ex: https://circuitverse.org/users/1935/projects/5524
- In the preview pane, click one of the steppers (in this case, one of the two squares inside the pixel art box)
  - Stepper becomes yellow, indicating it is selected.
- Press up/down

Expected:
- For steppers to change value, as it happens in design mode.

Actual:
- No response from stepper.

shubhankarsharma00:
Fixes #96 

added keydown2 and keydown3 eventListeners in embedListener.js


public/js/embedListeners.js
```diff
@@ -96,6 +96,21 @@ function startListeners() {
...
         }
+        if (simulationArea.lastSelected && simulationArea.lastSelected.keyDown2) {
+            if (e.key.toString().length == 1) {
+                simulationArea.lastSelected.keyDown2(e.key.toString());
+                return;
+            }
+
+        }
+
+        // if (simulationArea.lastSelected && simulationArea.lastSelected.keyDown3) {
```
ankitkataria: Is it possible to remove this comment if not required?

shubhankarsharma00: This was actually discussed last time it isn't required right now but might be required in future.

satu0king:
Neat! Thanks. 
@fpintos can you confirm that this is working as you expected?

PR #194, opened 2019-03-04
favicon added

sarthak77:
Fixes #125

PR #203, opened 2019-03-06
new favicon added

sarthak77:
Fixes #194

####new favicon added

newfavicon


blane1257:
It still seems to me a bit cluttered, and the "C" and "V" aren't particularly visible..

blane1257:
cv_logo_01

What about something more similar to this? Just something simple, with big, bold, and obvious letters, which can be seen clearly when the image is scaled down significantly. 

Of course, this was just thrown together quickly, and I don't intend for it to be the final product, as there are jagged edges, and imperfect angles.

satu0king:
I like @blane1257 logo, maybe we should have a poll after we shortlist some of them.

Issue #48, opened 2018-11-09
Subcircuit Layout upgrade


satu0king:
Referencing #36. 

Until recently subcircuits had very strict dimensions and positions and little could be done to move the node positions. As of now, subcircuits have flexible dimensions and node positions can be changed. 
Moving forward, the design pattern has to be considered to ensure extensibility and ease for both development and the users. 

Some of the features are to be considered to be implemented on the subcircuit layout.

- [ ] Output elements like LEDs (Digital+analog+RGB)
- [ ] Output elements like seven segment display, display matrix
- [ ] Input elements like button 
- [ ] Input elements like clock, stepper 
- [ ] Custom text elements in between

This would open a whole range of exciting possibilities. If the initial design is good, implementing these features will be easy. However, it is tricky as all LEDs and other output elements need not be part of the subcircuit. Providing a feature for the user to easily pick and chose his features will be difficult. Storing and manipulating the data from the development point of view is tricky too. 

@blane1257 @HypheX and thoughts? comments?

satu0king:
> This all sounds pretty great. Also, I love the text idea! perhaps it could have a few options for color, font, and size, too.

Yes this can be done too

HypheX:
I agree with blane, it sounds great! Can't wait for these features, I am "hyped".

jbox144:
I'll look into making some assets

blane1257:
surface elements such as buttons and LEDs should be individually togglable if possible

fpintos:
Could the input/output elements just have an extra property that tells if they show up in the layout or not?

We also need to consider if their labels show up or not; I can think of cases where we would want to and in cases where we would not want to.

The size of the leds might also be important; in a project like this  https://circuitverse.org/users/1935/projects/5010, I would love to have the LEDs in the "4-pixel cell" circuit, but they would need to be big enough to fill the layout, since the number of inputs dictates the size of the layout.

satu0king:
@fpintos These are valid points. This particular issue will take at least 3 months to implement correctly!

SubhajitCode:
@satu0king Hi I am interested in this project and I intend to be part of this one through GSoC .Can You Guide me please What is your primary Idea to do this project . I mentioned one of My own in the proposal But It will really help me a lot if you guide me here.....Thanks

fpintos:
Hi @SubhajitCode ,

[...]

PR #155, opened 2019-02-28
Added issue and pull request for the ease of creating one

creyente1897:
**Issue Number** #154 

#### Describe the changes you have made in this pr -
* Added issue template
* Added pull request template

Fixes #154

Issue #115, opened 2019-02-20
Subcircuit Layout height automatically increasing by 1 unit. 


satu0king:
Eventhough, in the edit mode, it has been configured to have this height, The height increases by one unit. 

Configured Height
![image](https://user-images.githubusercontent.com/2092958/53066168-a1a32900-34f4-11e9-81a9-e773e835ad0e.png)

Resultant Height.
![image](https://user-images.githubusercontent.com/2092958/53066146-8c2dff00-34f4-11e9-82a0-ba263b4858e2.png)

blane1257:
Yes, this happens all the time, and is really annoying!

satu0king:
> Yes, this happens all the time, and is really annoying!

Please report such things, even if it is small! I didn't know it till today when I was trying to use the SquareLED

SubhajitCode:
Hi I created a pull req to fix this bug kindly review it and let me know 
https://github.com/CircuitVerse/CircuitVerse/pull/220

PR #210, opened 2019-03-07
Subcircuit layout upgrade #48

### LINKED ISSUE #36: SubCircuit Surface LED's
When customizing the layout of a SubCircuit, any LED's that have been placed should appear on the SubCircuit similar to inputs or outputs, but on the surface. A similar idea I just had is to do the same with buttons, allowing for a SubCircuit button panel. 
Here's a graphical representation of what I mean about the surface LED's:
screenshot 2

SubhajitCode:
Fixes #36 

implements the the surface led (circular can be changed to rectangular..)
label is not visible yet and the layout size some time unstable ...
need more thorough work  and guidance from the core team ...
check before merging because all leds will be surface leds for now needed to work on that also
I request to be assigned on this issue for GSOC 2019

toggling led still pic


.gitignore
```diff
@@ -7,7 +7,11 @@
 # Ignore bundler config.
 /.bundle
 /.idea
-
+#my ignore specific to my environment
+/public/js/*.json
+/docker-compose.yml
+/Dockerfile
+#end ofmy ignore specific to my environment
```
satu0king: /docker-compose.yml and /Dockerfile cannot be in .gitignore!

public/js/logix.js
```diff
@@ -39,14 +39,14 @@ forceResetNodes = true; // FLag to reset all Nodes
 circuitElementList = [
     "Input", "Output", "NotGate", "OrGate", "AndGate", "NorGate", "NandGate", "XorGate", "XnorGate", "SevenSegDisplay", "HexDisplay",
     "Multiplexer", "BitSelector", "Splitter", "Power", "Ground", "ConstantVal", "ControlledInverter", "TriState", "Adder", "Rom", "RAM", "TflipFlop",
-    "JKflipFlop", "SRflipFlop", "DflipFlop", "TTY", "Keyboard", "Clock", "DigitalLed", "Stepper", "VariableLed", "RGBLed", "SquareRGBLed", "Button", "Demultiplexer",
+    "JKflipFlop", "SRflipFlop", "DflipFlop", "TTY", "Keyboard", "Clock", "Stepper", "VariableLed", "RGBLed", "SquareRGBLed", "Button", "Demultiplexer",
```
satu0king: You cannot just remove "DigitalLed from here!'

satu0king:
This looks like a good start and you are right about more thorough work and guidance from the core team. This is quite a critical module and we would want it to be done only during the GSoC period. You should try and tackle easier issues like #155 for now.

public/js/logix.js
jbox144: He can, and he did. Not that he should have... but he did.

SubhajitCode: thanks for the feedback I will correct the mistake and create another pull req and definitely look into issue #115

satu0king:
As previously discussed, this is a big project, and we would love that you tackle this as a GSoC project. You are of course open to experiments and write your findings in the proposal. We can then discuss which is the best approach based on your thoughts.

PR #217, opened 2019-03-08
fix bootstrap class of alert dialog

### LINKED ISSUE #213: Wrong use of bootstrap class in Delete CIrcuit
We are currently using the wrong class of bootstrap alert to give the warning when there are less than 2 circuits.

image

**Solution**: Use `alert-warning` or any other class if you think it would make more sense. Look into `showError` and `showMessage`

sanctophagus:
Fixes #213 

data.js:318: changed `showMessage` to `showError`

alert-fix


sanctophagus:
@armudgal Is it fine?

PR #219, opened 2019-03-09
Add ColorTTY.js

### LINKED ISSUE #190: ASCII capable display
#### I am submitting a -
**Feature Request**

#### Describe your feature -
Currently, we have an ASCII TTY display, and we are able to make displays using LEDs. This is all great, but at the moment there isn't much of a way to display ASCII on top of a pixel display, and displaying ASCII characters on a display is a bit impractical. So, why don't we make a component able to display both ASCII characters, as well as with a color background, similar to how many early terminal computers used.

#### ASCII Capable Display
The display should have an X coordinate and Y coordinate (with customization bitwidth, up to 8 bit), a background RGB input, foreground (ASCII) RGB input, ASCII character input, write input, clear, and 4 "shift" inputs which will shift the whole image in one of 4 directions (up, down, left, right). There should also be a DMA pin (if/when they are added) to be able to asynchronously fill the display.
#### Image
terminal_control colored_text

#### Are you working on this issue? (Yes/No)
No.

abdullah-mohamed:
[WIP]
issue #190 

I have added documentation for the element, created base class and model.
I have added most of the code needed, and still working on `customDraw()` function
I just want to be sure that I am on the right path and not missing anything before I go on.
my code is in ColorTTY.js file.

none yet


abdullah-mohamed:
@fpintos  @armudgal  This is my PR
@satu0king 
In TTY constructor function, on determining the width/height of the element.
When we have 32 cols and 3 rows, The width will be 320 units and the width will be about 50 units.
Assuming we don't divide by 2 before setting the dimensions, Is the unit pixels ?
Does the width and height include the border ?

blane1257:
There shouldn't be width and height settings, as this will become confusing to address. The width and height should be determined by the bitwidth of the X and Y coordinates. For example: An X bitwidth of 4 and Y bitwidth of 5 will result in 16 columns and 32 rows.

abdullah-mohamed:
@blane1257 
Good point. 
To make it not confusing, I think we can determine a specific height and width for each screen element.
Then the height can be `(2^yBitWidth) * specificHeightOfScreenElement` and so is the width.
 
Right now this won't be a problem as the user can't determine the width/height/columns/rows. All these values are all defined inside the code. There will always be 32 rows and 32 columns.
5 bits for both X and Y bitwidths allowing user to access any of the 32 rows/cols.

Thanks for the feedback, I will be adding more code tomorrow. Please let me know if you have any other notes.

blane1257:
please share screenshots as soon as possible, as I don't know the language, and sometimes words arent descriptive enough.

I'm excited to see more from this!

blane1257:
And just to be clear, each "pixel" shouldn't be square, instead it should be a bit taller than it is wide, as to match up with the ASCII characters better, without leaving many large spaces between letters

abdullah-mohamed:
@blane1257 you mean screen element, right ?
right now there are 32 * 32 elements, you can change each element 's back/fore color but each element will have the same back/fore color across all its height/width.

blane1257:
I'm sorry, I don't understand what you're saying. Are all the elements locked to the same background and foreground color?

abdullah-mohamed:
@blane1257 no, This is what I am aiming for.
Untitled

fpintos:
To answer the question at the beginning....

The "dimension" is not pixels, as the circuit can be zoomed in/out.

When you look at the circuit, notice the grid; each square is 5 units in size.
When you call setDimensions(dx,dy) you end up with an element that is twice as big - at first this seems weird and made me scratch my head when I started to contribute, but if you read the source you'll see that the CircuitElement has 4 dimension properties: left/right/top/bottom, so setDimensions(5,10) is saying "5 to the left, 5 to the right, 10 to the top and 10 to the bottom"; if needed, you can set the properties individually if the "center" of the element is not really centered. Not sure why it was done like this, maybe it is about where the element rotates (when it supports rotation).

You want to keep dimensions as multiple of 5 most of the time; in particular, the nodes need to land in multiples of 5, so they will connect properly.

satu0king:
> To answer the question at the beginning....
> 
> The "dimension" is not pixels, as the circuit can be zoomed in/out.
> 
> When you look at the circuit, notice the grid; each square is 5 units in size.
> When you call setDimensions(dx,dy) you end up with an element that is twice as big - at first this seems weird and made me scratch my head when I started to contribute, but if you read the source you'll see that the CircuitElement has 4 dimension properties: left/right/top/bottom, so setDimensions(5,10) is saying "5 to the left, 5 to the right, 10 to the top and 10 to the bottom"; if needed, you can set the properties individually if the "center" of the element is not really centered. Not sure why it was done like this, maybe it is about where the element rotates (when it supports rotation).
> 
> You want to keep dimensions as multiple of 5 most of the time; in particular, the nodes need to land in multiples of 5, so they will connect properly.

You are perfectly correct. I think each square is 10 pixels, not 5?

fpintos:
> You are perfectly correct. I think each square is 10 pixels, not 5?

Indeed, I stand corrected: each square is 10 units.

For example, the flip-flops call setDimensions(20,20) and end up as a 4x4 square.

satu0king:
@abdullah-mohamed, can you give us a status update on this?

abdullah-mohamed:
@satu0king 
I am still working on the `customDraw()` fn.
I had some issues concerning the difference of sizes between capital and small letters so I searched for a specific font and found a monospaced font that I hope will solve the problem.
According to my timeline, I will update the final pull request in 1 or 2 days max. 

To be noted: 
I only started local development a couple of days ago as I had some trouble setting my env either on my windows or using Docker(I asked for help on the community chat a couple of times) . I ended up using a centos Linux on a VM which finally worked great but just slower.

blane1257:
looks promising so far!

blane1257:
And regarding the font, I think a terminal font will be most aesthetically pleasing.

"Terminal is a family of monospaced raster typefaces. It is relatively small compared to Courier. It uses crossed zeros, and is designed to approximate the font normally used in MS-DOS or other text-based consoles such as on Linux. ... Fixedsys fonts of different code pages have different point sizes."

abdullah-mohamed:
Here is a new screenshot, font is perfect, background color working great, still need to correctly position the character and its background on the screen relative to the center of the element and no. of rows and cols 
Untitled

abdullah-mohamed:
This is taken now, Font is great. background color also, still need to correctly position the backgrounds and characters according to the number of rows and cols relative to the center.
Untitled

blane1257:
Are you displaying the letters as a single or a few strings, or are you displaying each letter or "pixel" as a seperate string, independent of the letters around it?

It seems best to me to use the latter method, as it replicates the eventual behavior, and the spacing should work well.

abdullah-mohamed:
@satu0king @blane1257 @fpintos 
Almost Done !!
The background is the same for all elements in the following screenshots because it is just for testing purposes but it was drawn one by one not a whole rectangle so when I need to specifically change one of the elements' background color it will be easy.
Width and height of the screen are now variables as multiples of 2.

A 16 * 64 screen.


An 8 * 32 screen.
Untitled

I still need to do :
    1- make svg more beautiful.
    2- find out why zooming in/out doesn't update the ctx.rect() method.
    3- find the bug inside the char positioning. (missing characters in some lines)
    4- make the color positioning approach the same as the char positioning.
    5- refactor some of the large functions.
    6- add mutable width and height/ reset button to properties UI.
    7- add a very small spacing between characters.

@blane1257 Indeed, I am targeting pixels, one by one.

blane1257:
It's coming along nicely!

I think it would be more aesthetically pleasing if the font were to mimic the font of early computers. Like for example, the font of the Commodore 64 computer.
c64-truetype-v1 0--style
Or the MS-DOS font.
Ms_dos_font

Of course, those are just examples. 

One concern with this is that these fonts are "pixelated" and the borders of the background elements might not line up with the "pixels," though I'm sure there are fonts that aren't pixelated but still mimic the early Commodore 64 or MS-DOS style.

abdullah-mohamed:
@blane1257 I used Lekton font. I can change the font as easy as changing 2 lines of code only.
If you have any suggestions of a specific FREE font, please give me its name or even better its API link directly such as : google fonts.
It is finishing up!
I fixed all the rendering bugs. I still need to clean up the code a bit so the code will pass all the tests.
Untitled

satu0king:
Looks nice!

blane1257:
A few fonts I think would look well...


https://www.fontspace.com/johan-winge/svi-basic-manual

https://www.fontspace.com/grudnuk-creations/modenine

ww.fontspace.com/i-shot-the-serif/white-rabbit

blane1257:
perhaps there should be a setting where the user can select a font from a dropdown list.

abdullah-mohamed:
@satu0king What do I need to get this PR merged? 
I have locally tested it by filling the `screenCharacters` array with values ready and there was no error rendering the characters/background.
Only when I zoomed in/out, the background gets outside the boundries(don't know why yet but I think I will either look into `zoom()` fn in `canvas.js`  or try make a function to adapt to zooming changes in my file).
Also the last commit, It makes user able to select a font to use. The code is almost complete but for some reason the dropdown doesn't appear in Properties sidetab of my element.
please I need a review so I can get this merged to add to my gsoc proposal ASAP.
maybe you can give me a circuit to test it and upload a GIF in this PR?

satu0king:
Please resolve the conflicts, then we can test if the element is working as per expectations.

abdullah-mohamed:
Done

abdullah-mohamed:
@satu0king should I fix those 274 issues ?

satu0king:
> @satu0king should I fix those 274 issues ? most of them aren't even issues critical!

We have a lot of JS linting issues as we only recently started using it. We will fix them later.

satu0king:
Not sure what is happening but I am getting these errors. 
image

The below error is probably because you didn't consider DPR in computing something. 
image

abdullah-mohamed:
@satu0king `node.value` returned an integer and I was using a string-type function.
I noticed you made Yposition bits 16, but I know the correct number of bits is only 5. This is another issue which I fixed. 
This error is gone now.
I will look into the DPR issue now.

abdullah-mohamed:
Fixed the errors with filling the buffer properly, now only the DPR left.
Untitled

abdullah-mohamed:
Fixed it! please review furthur
Untitled

blane1257:
It looks really good!

app/views/simulator/embed.html.erb
```diff
@@ -95,6 +95,7 @@
  <script src="/js/modules.js"></script>
  <script src="/js/sequential.js"></script>
  <script src="/js/ram.js"></script>
+ <script src="/js/ColorTTY.js"></script>
```
satu0king: You have to add this in edit.html.erb as well?

satu0king:
I tried testing it out. 
image

I am getting this error.
image

I think there are some cases you have missed out. Overall great work! You should consider project1 for gson.

abdullah-mohamed:
@satu0king The error can be fixed by checking if the value of the pins is undefined in `bufferFill`  fn then the function shouldn't execute. I will fix it and make a commit.
Indeed I think project 1 is really interesting, I already submitted my draft. I received the comments on the parts that I should give more details about and I will update my draft ASAP.

abdullah-mohamed:
@satu0king Are there any more issues to fix?

satu0king:
> @satu0king Are there any more issues to fix?

I am not sure what exactly is the problem. I just connected ascii and clock and it crashed. 

image

satu0king:
I will take a look into this and fix any remaining issues and merge the PR.

PR #222, opened 2019-03-09
Updated CodeofConduct

creyente1897:
Added ground rules and expectations in the code of conduct guide

### DIFF
--- a/CODE OF CONDUCT.md
+++ b/CODE OF CONDUCT.md
@@ -6,6 +6,13 @@ We are committed to making participation in this project a harassment-free exper
 
 Diversity and inclusion make the CircuitVerse community strong. We encourage participation from the most varied and diverse backgrounds possible and want to be very clear about where we stand.
 
+## Ground rules & expectations
+
+Before we get started, here are a few things we expect from you (and that you should expect from others):
+
+* Be kind and thoughtful in your conversations around this project. We all come from different backgrounds and projects, which means we likely have different perspectives on "how open source is done." Try to listen to others rather than convince them that your way is correct.
+* When adding content, please consider if it is widely valuable. Please don't add references or links to things you have created as others will do so if they appreciate it.
+
 ## Expected Behaviour
 
 - Be professional.


CODE OF CONDUCT.md
```diff
@@ -6,6 +6,13 @@ We are committed to making participation in this project a harassment-free exper
 
 Diversity and inclusion make the CircuitVerse community strong. We encourage participation from the most varied and diverse backgrounds possible and want to be very clear about where we stand.
 
+## Ground rules & expectations
+
+Before we get started, here are a few things we expect from you (and that you should expect from others):
+
+* Be kind and thoughtful in your conversations around this project. We all come from different backgrounds and projects, which means we likely have different perspectives on "how open source is done." Try to listen to others rather than convince them that your way is correct.
+* When adding content, please consider if it is widely valuable. Please don't add references or links to things you have created as others will do so if they appreciate it.
```
armudgal: Can you explain this point?

armudgal:
Also pull from our `master` branch, since we renamed the code of conduct file.

CODE OF CONDUCT.md
creyente1897: @armudgal I added it to let people know that there will always be different opinions for the same issue and we should respect each other opinions and users should add content to the guide if and only if it is valuable

armudgal:
Hi, for the first point we have something already mentioned here https://github.com/CircuitVerse/CircuitVerse/blame/master/code-of-conduct.md#L9-L16 and for the second point it is mentioned here in a different language https://github.com/CircuitVerse/CircuitVerse/blame/master/CONTRIBUTING.md#L6-L17.

Worst case scenario: Someone makes a PR with the doc changes, if we think that the mention is worth it, the PR will get merged otherwise it will be closed with the satisfaction of the author.

Hence I think this PR is not needed. Thoughts?

creyente1897:
@armudgal Yes you are right :smile:

PR #226, opened 2019-03-10
Add support for mysql in CI

tachyons:
Support for mysql in CI to ensure that changes are compatible with both postgres and PG


tachyons:
Seems like group spec is failing for mysql, @ankitkataria have a look

ankitkataria:
Looking into it.

ankitkataria:
@tachyons @armudgal Is it possible to rerun this possible workflow? All the specs seem to be passing for me locally. Although, they are taking a little longer than postgres.

Issue #117, opened 2019-02-20
ROM update


blane1257:
My idea is to make the ROM writable. Not necessarily persistent between saves, I mean like have an origin state, but instead of being locked in that state throughout the simulation, add the ability to write new values, and whenever the simulation is saved, the current state of the ROM component becomes the origin. This would make it so computers can much more easily build and run programs, instead of using RAM and potentially a slow program to run RAM programs.

Another feature I would love to see is customizable address bitwidth for the ROM components, but I understand this can cause some issues with the displaying of values on the surface of the ROM, but this could possibly be overcome by a DMA pin much like is described in #113

fpintos:
Sounds like you want an [EPROM](https://en.wikipedia.org/wiki/Eprom).

We can derive from the RAM and just include the memory array in the serializer.

The only issue I can think is that a large array will generate a large JSON; even the memory addresses that were not populated (and thus make the array sparse) will get converted to 'null' in the generated JSON, so maybe EPROMs should be limited in size.

blane1257:
Well, kind of more similar to an EEPROM, but yes. I basically want it to behave lile the RAM, but with an origin state, which can be updated to the current  by saving the simulation

blane1257:
If there's a "reset" pin, should this clear the entire ROM, or revert it to it's origin state?

Or maybe two reset pins, a "full reset" or maybe "format" which clears everything, and a "factory reset" which reverts to the origin state.

satu0king:
> Well, kind of more similar to an EEPROM, but yes. I basically want it to behave lile the RAM, but with an origin state, which can be updated to the current state by saving the simulation

Yes, what you need is a RAM which starts which an origin state which we can do by using a DMA module which can put data into the RAM.

blane1257:
well, i would also like the ability to update the origin state when the simulation is saved

blane1257:
really, the reason is, programming a 65KB ROM by hand sounds absolutely horrible to me. what i'd love to be able to do is build small programs to help me program the computer faster (like IDE's, interpreters, and other programs) so I can much more quickly fill the ROM with programs and files, then just save the end product and be done.

fpintos:
I'll write the EPROM this weekend.

blane1257:
Awesome!

PR #229, opened 2019-03-11
Add support for an EEPROM

### LINKED ISSUE #212: RAM should save it's values
RAM components are really handy for storing a lot of data, which is why I think they should also be able to save it's values with the project. Hopefully this isn't a big ask.

fpintos:
Fixes #212 , #117 

# New Element: EEPROM

The [EEPROM ](https://en.wikipedia.org/wiki/EEPROM) is a non-volatile memory element.

Its content is persisted within the CircuitVerse project and its pins and usage are identical to the RAM element.

EEPROMs are more "expensive" than RAMs, so the maximum address space of the EEPROM is smaller than the RAM, limited to 10-bits (1024 addresses). For CircuitVerse, this helps keep the serialized project small.

Erasing the EEPROM will make all its values zero. There is no way to recover the original data once you erase it, just like it happens on a real EEPROM. That being said, you can always reload the project and the saved data will be there. If you erase the EEPROM and _save_ the project, then, you know...you asked for it. :-) 

EEPROM data survives cut-and-paste operations, so you can do interesting things, like duplicating EEPROMs, creating an EEPROM in one circuit and then cut-and-pasting the element in another circuit, or you can have a library of ready-to-use EEPROMs with interesting data.

## Programming EEPROMs manually
Users might want to manually program the contents of an EEPROMs instead of building a circuit to do so.

An easy technique for this is to edit the JSON representation of the element directly:
- select-and-copy an EEPROM
- paste the contents of the clipboard into your preferred editor
- look for the data array in the constructorParamaters of the corresponding EEPROM
- update the array with the numbers you desire
- select the whole JSON and paste it back into CircuitVerse.

> The same technique can be used to pre-populate a ROM element.

> One can argue that we should have a built-in way to populate ROM, RAM and EEPROM in the UI. We have it for the ROM, which is limited to 16 bytes, but RAMs and EEPROMs can be larger, so the ROM UI does not scale for them. Maybe a simplistic solution would be to just ask for a string then parse it as a JSON array. This can be added in a separate issue/PR.

# Screenshots
image


public/js/eeprom.js
```diff
@@ -0,0 +1,46 @@
...
+    // Normalize this.data to use zeroes instead of null when serialized.
+    var data = this.data;
+    for (var i = 0; i < data.length; i++) {
+        data[i] = data[i] || 0;
+    }
+
+    saveInfo.constructorParamaters.push(data);
+    return saveInfo;
+}
+
```
satu0king: I love that EEPROM is able to inherit from RAM so well.

satu0king:
Did a quick test, looks good.

blane1257:
I'm a little bit disappointed with this..

The reason I wanted this component was because the idea of having a computer with a file system and that gave the end user the ability to create their own programs made me very excited. However, in the current state of this component (with only 10 bits for the address) such a computer would be almost impossible, especially when considering my planned ROM size of 64KB (16 bit address.) I could still have a larger RAM which on startup would load values from a ROM, maybe using DMA pins, but this still does not allow for saving the RAM values onto the ROM, meaning I would be unable to save any programs made in the computers software, the only programs that could be saved are those which were programmed by hand onto the ROM.

fpintos:
Hi @blane1257 , let me address the different issues you pointed out:

> The reason I wanted this component was because the idea of having a computer with a file system and that gave the end user the ability to create their own programs made me very excited. However, in the current state of this component (with only 10 bits for the address) such a computer would be almost impossible, 

You shouldn't feel limited by the address space of an element. You can always increase the address space by multiplexing them. This is how circuits in the real world create address spaces from smaller parts. 

As an example, here is a 256-bytes ROM built from the built-in 16-bytes ROM: https://circuitverse.org/users/1935/projects/4872. You can do the same with the RAM and EEPROM and you can then multiplex that circuit to build even bigger ones. Before we made the RAM built-in I was building large RAMs like this: https://circuitverse.org/users/1935/projects/4774, but then eventually we ended up hitting a limit on the size of the simulation queue.

> especially when considering my planned ROM size of 64KB (16 bit address.) 

64KB of ROM (or EEPROM) in this simulator sounds fairly ambitious. :-) Back in the days, Z80 computers managed to get away with 16KB of ROM and that was handling graphics and BASIC interpreters.

We can certainly bump the address space limit to reduce the need to create multiplexing circuitry, but keep in mind the simulation is slow (on purpose, I suppose). The min clock time is 50ms for each tick in the simulator, meaning you need 100ms for a full up/down cycle, which means you can read (or write) about 10 addresses/second on a single address line. Reading 64KB addresses would take 6554 seconds, just shy of 2 hours.

All that being said, choosing a smaller address space for the EEPROM was arbitrary, so we can always change it. The intention was to give a sense that you don't design "real" circuits around a lot of EEPROM, because they cost more per byte in the real world.

> I could still have a larger RAM which on startup would load values from a ROM, maybe using DMA pins, 

My suggestion would be to avoid copying - you can split the address space of the computer so that certain addresses target the EEPROM and other addresses target the RAM or even a ROM. This way you don't need to spend time copying values from the EEPROM to the RAM. Since the EEPROM is programmable you can always write directly to it, without having to write to the RAM and then copying to the EEPROM.

If you look at the larger RAMs in my example, https://circuitverse.org/users/1935/projects/4774, you'll notice they take an address line, split its bits and use some bits to select the smaller components and passing the remaining bits as address to those smaller components. That is how you divide the address space between ROMs, EEPROMs, RAMs, graphic cards, etc.

> but this still does not allow for saving the RAM values onto the ROM, meaning I would be unable to save any programs made in the computers software, the only programs that could be saved are those which were programmed by hand onto the ROM.

The ROM is indeed read-only, but the EEPROM you can write to it like you do with the RAM. Once you save the project, the values in the EEPROM will persist. Values in RAM will NOT persist.

blane1257:
The reason I would rather not multiplex is because it's much less practical, to achieve the same 64K space I would need 64 individual EEPROM components. Though I suppose you are right about such large address spaces, until now I haven't really put much thought into the time it takes to read/write, as "It's only one clock cycle, which is as fast as it could possibly get, so it couldn't possibly be an issue." But now I see, such a large ROM would be very impractical given the speed, so I should probably go down to 32K or 16K. This still would require 16 EEPROMs (or 8 for 16K) but that is significantly less..

The idea of one big address space with separate sections for ROM, RAM, and EEPROM sounds like a massive pain to make programs around, unless i misunderstand what you are saying.

PR #230, opened 2019-03-11
[FIX] Bootstrap

### LINKED ISSUE #215: Using Bootstrap CDN instead of hosting on our server
Hi,

So the issue title is pretty self-explanatory. It would be wise to use the static bootstrap files using CDN so that the loading time is faster.
Bonus: Cleans our code a little bit. (~22% reduction in our repo size)

sanctophagus:
Fixes #215 

Delete `public/bootstrap-3.3.7`
Delete `public/css/4.0.0_bootstrap.min.css` && `public/css/bootstrap-3.3.7.css`
Delete `public/js/1.13.0_popper.min.js` && `public/js/4.0.0_bootstrap.min.js`
replace hard links (within project) with  `CDN` links


app/views/simulator/edit.html.erb
```diff
@@ -11,12 +11,9 @@
 
 
 <!-- <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css"> -->
-<link rel="stylesheet" href="/css/bootstrap-3.3.7.css">
+<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
```
satu0king: We are using a custom build of CSS. Does the look of the simulator remain the same? I doubt it.

sanctophagus: Okay, I'll look into it.

satu0king: Looks like I was wrong, @sanctophagus can you confirm once, that all pages look same as before with no errors. I will merge then.

armudgal: @satu0king hey, all the bootstrap files are independent. We can go ahead and merging with this but I would wait for @sanctophagus to confirm once.

sanctophagus: @armudgal @satu0king  LGTM.

satu0king:
Looks good to me.

PR #232, opened 2019-03-12
Implement Subscriptions

### LINKED ISSUE #91: Implement Subscriptions for comments
https://github.com/lml/commontator

Check the documentation and add the functionality of subscriptions to comments.

Amansingla97:
Fixes #91


config/environments/development.rb
```diff
@@ -52,7 +52,7 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.active_job.queue_adapter = :sidekiq
-
+  config.action_mailer.default_url_options = { host: "www.circuitverse.org" }
```
tachyons: this should be localhost in development

tachyons:
Looks good, Few nitpicks
1. Add screenshots
2. Check the feasibility of moving logic out of the view using presenter pattern or decorator pattern
3. Check if we can simplify the frontend operations with https://stimulusjs.org/

config/environments/development.rb
Amansingla97: yes ill change it

Amansingla97:
Screenshot 2019-03-12 at 3 17 19 PM
Screenshot 2019-03-12 at 3 17 24 PM

satu0king:
Thanks for the PR. Can you summarize what is the current behavior? I noticed that you generated views from the default commentator plugin, can you summarize the changes.

Amansingla97:
I have made small changes to views of subscription which differ from default commentator plugin

->There were syntax error in the orignal repo (I raised a issue in their repo as well)
->We may need to customize the mailer views.So having app/views/commontator/subscriptions_mailer/ is be necessary.
->The UI of overall commontator does not match with our UI.so would have to change it at some point or other.
Rest the working as usual

satu0king:
> I have made small changes to views of subscription which differ from default commentator plugin
> 
> ->There were syntax error's in the orignal repo (I raised a issue in their repo as well)
> ->We may need to customize the mailer views.So having app/views/commontator/subscriptions_mailer/ is necessary.
> ->The UI of overall commontator does not match with our UI.So would have to change it at some point or other in future.
> Rest the working as usual

Thank you, could you also describe how subscriptions happen. Does it happen when the author creates the project, or when someone comments on the project. Or is it a manual subscruption. Also how are the emails sent? Are they sent for every comment? A screenshot for an email will be helpful.

Amansingla97:
->The email happens when someone else comments on the subscribed project.
->The User can opt to manually opt to subscribe.
->Emails are sent via smtp and would be using the same smtp settings.
->When a new comment is generated the email would be sent to all the subscribed users.it will enqueue the tasks in sidekiq which is then processed via smtp.

satu0king:
> ->The email happens when someone else comments on the subscribed project.
> ->The User can opt to manually opt to subscribe.
> ->Emails are sent via smtp and would be using the same smtp settings.
> ->When a new comment is generated the email would be sent to all the subscribed users.it will enqueue the tasks in sidekiq which is then processed via smtp.

Can you configure auto subscription? And manual unsubscription? I really doubt users will ever subscribe even if they would be interested.

Amansingla97:
Screenshot 2019-03-15 at 8 19 30 PM
These were available options , I changed it both automatic and manual.

satu0king:
Is the author of the project automatically subscribed even if he doesn't make a comment?

Amansingla97:
No,I just checked .Commontator does not provide this functionality.

satu0king:
> No,I just checked .Commontator does not provide this functionality.The author would also have to subscribe to comments or automatically when commenting.

I am sure Commontator has some function which we can call to subscribe/unsubscribe? After all, they need to have a way to change it themselves. Simply add a after_create hook?

Amansingla97:
ill check

satu0king:
> ill check

We can resolve this in a different issue.

satu0king:
undefined method `project_url'

This is the error that is coming in production, did you test the mail in development?

Amansingla97:
ya i tested with gem letter_opener
let me see again.I am in class right now.Will definitely look into it.

satu0king:
```
  config.comment_url_proc = ->(comment, app_routes) {
    app_routes.polymorphic_url(comment.thread.commontable, anchor: "comment_#{comment.id}_div")
  }
```
I think the answer is here.

PR #238, opened 2019-03-12
Track users in Google Analytics

### LINKED ISSUE #237: Track users in Google Analytics
Being able to track usage through Google Analytics ends up being cumbersome since many users will be counted twice when accessing from different devices or locations.

Send the `current_user` id along with the tracking token for GA. This would help converge all the activities by a single user across all devices and locations to a single user on GA natively.

**Describe alternatives you've considered**
None.

**Additional context**

Yes.

arkokoley:
Fixes #237

This patch sends the `current_user` 's id to the google analytics engine to provide deeper analytic insights.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -33,6 +33,9 @@
       gtag('js', new Date());
 
       gtag('config', 'UA-112678513-1');
+      <% if current_user %>
+      gtag('set', {'user_id': '<%= current_user.id %>'}); // Set the user ID using signed-in user_id.
+      <% end %>
   </script>
 
 </head>


tachyons:
Does this means we have to update privacy policy to reflect this tracking behaviour ?

arkokoley:
I don't think so. The privacy policy already states that we use google analytics to track usage. Not sureis this would require any changes. @satu0king might be able to explain this better.

satu0king:
@tachyons @arkokoley To be fair, when we created the TOS and Privacy page, we didn't put much thought as we were just getting kickstarted. Our TOS and Privacy page is copied from MIT Scratch, since they have a similar model. 

But yes as @arkokoley mentioned, we explicitly mentioned google analytics in our Privacy Policy.

PR #261, opened 2019-03-17
Add featured circuits management

### LINKED ISSUE #6: Featured Circuits/ Editor Picks
[Desmos](https://www.desmos.com/)
Refer to the way Desmos shows featured works for more details. 

Currently featured circuits is hardcoded temporarily. Will need a nice way for admins to selectively feature circuits. 

- [ ] Create Model FeaturedProjects - primary attribute - project_id
- [ ] Send email to the featured project's author 
- [ ] Monthly/ Weekly send email to everyone about past month's/week's featured works 
- [ ] Ensure FeaturedProjects cannot be made private or limited access
- [ ] Provide a way for people to submit their work to be featured

ankitkataria:
Fixes #6 

- Add a featured_circuits model
- Add mailers to inform the project owner about the project being featured on creation 
- Add featured_circuits_controller with `/feature` route

Todos - 
- [x]  ~Ensure FeaturedProjects cannot be made private or limited access~ Added warning for featured projects for changing access type. 
- [x]  ~Provide a way for people to submit their work to be featured~ Added featured checkbox at `users/:id/projects/:id/edit` for admins to feature projects
- [x] Add tests for models and mailer

Note - 
Task - `Monthly/ Weekly send email to everyone about past month's/week's featured works` can be better worked on once #7 is complete. 

## Screenshots of the changes (If any) -

`/featured` ( Edit: `/examples` page is now the same as before)
image

`/` (Edit: Now there'll be three sections, examples section will be seen in prod, the headings of `Featured Examples` and `Editor Picks` section are now clickable and point to `/examples` and `/featured` respectively)
image

`featured_circuit_email` 
- similar to `new_project_mail` with changes in subject and the `More Featured Examples` card
image

`project update page` - added a checkbox in project edit page to make it featured, which will only be visible to admins for public projects
image


ankitkataria:
@satu0king How do you propose to give users the facility to ask make their projects featured? Do we want to separate form for this and a link to this form at a few places? Or maybe a simple note to email the organization for getting their work featured?

satu0king:
For now, let us keep things simple. Let the examples page be fixed. Those are circuits that are hand-picked to illustrate the simulator features at large. 

I am thinking 3 sections on the home page. 

1. One is featured examples which are currently listed as featured circuits. This will basically be a shorter version of the examples page.
2. Second is Editor picks/featured circuits which you are working on - this will be the notable circuits by users the that are interesting, this will have a see more link where all featured circuits every can be seen.  
3. Recent circuits - current recent circuits 

@blane1257, @HypheX, @jbox144 what do you think about this? 

Right now the website is getting ~500 visitors per day. Once it grows, we might have to have an automated trending section based on user views/likes etc. 

For now, no need for a way for the user to request to be featured. Let there be a button in the project page "Feature This" which only admins can see. Admins can manually feature circuits. If there is a way that is simpler for you to implement this, go ahead. 

And you are right until #7 is completed, we cannot go forward with emails. Let the user whose circuit has been featured get notified, I see you have already implemented this.

ankitkataria:
@satu0king I've add a separate section for `Editor Picks`. As for the `Feature This` button, I had a few doubts. Some project are private, but they are visible to admins. Do we need to override the project_access_type if the admin chooses to make it featured? 
I was thinking that we should not give the admins the ability to override the access type if the project is private. Some users might not like this. :stuck_out_tongue: 
Also is the feature  `Ensure FeaturedProjects cannot be made private or limited access` required? 
Here I feel we should give the users the freedom to make their projects private if they don't wish their projects to be featured.

tachyons:
Instead of having separate table, you can just add `featured` boolean flag in the project and set it as false by default

ankitkataria:
Agreed. I was actually thinking of adding it as a `project_access_type`, but I had apprehensions of breaking the code at places where this value was not handled. 
Also, I felt that adding it as a separate boolean column is essentially Issue #267. What are your views on this? :roll_eyes:

tachyons:
Featured isn't an access type. It is semantically different. Better keep it as different flag

ankitkataria:
:gift: 
@tachyons  Added a separate column `featured` in `project` table itself , Also added tests. 
@satu0king Added a checkbox for making the projects featured which will only be visible to admins. Added a warning for featured projects on top of the project access type field. If the projects are featured and are made Private by the user, then they are no longer featured.

satu0king:
@ankitkataria @tachyons sorry for replying late! This might lead to repeat work. 

We would actually like it to be a separate table maybe? because we would want to store more information as to which admin featured it (later), when was it featured etc. Because the featured page will be ordered on the featured date. Later if we were to have other sections like trending which might take information from google analytics, view count, stars etc we would want that to be on another table right? 

@tachyons is right it cannot be added to project_access_type as the meaning is different. Access_type is mostly for authorization while featured is not. Of course we do have a constraint that featured cannot be private. 

@arkokoley @vik-y what do you say?

tachyons:
Agreed

app/controllers/logix_controller.rb
```diff
@@ -18,6 +18,10 @@ def gettingStarted
   def examples
   end
 
+  def featured
+    @projects = Project.where(featured: true, project_access_type: "Public")
```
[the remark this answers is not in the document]
satu0king: How does ruby ORM handle this query, @tachyons. Is the sql query of both the options equivalent?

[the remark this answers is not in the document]
tachyons: This was old review I forgot to submit. This is no longer relevant since we decided to go for separate table

app/controllers/logix_controller.rb
```diff
@@ -18,6 +18,10 @@ def gettingStarted
   def examples
   end
 
+  def featured
+    @projects = FeaturedCircuit.all.select(&:valid?).map(&:project)
```
tachyons: Why validating here ? This is expensive

tachyons:
Instead of having too many conditions to keep data consistent, you can have 

`FeaturedCircuit.joins(:project).where(projects: {access_type: 'Public'})`

app/controllers/logix_controller.rb
ankitkataria: I did this initially before I had implemented the hook to delete featured_circuit entry in project.rb. I'll remove it now.

app/controllers/featured_circuits_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class FeaturedCircuitsController < ApplicationController
+  before_action :authenticate_user!, only: [:feature]
+
+  def feature
```
tachyons: Since we are using a separate controller for this, can we make proper REST path here

Like
```ruby
class Circuits::FeatureController < AC
  def create
  end
end
```
You can read more about this pattern here
http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

ankitkataria: This blog is really informative!

tachyons:
Can you try removing N+1 queries? https://semaphoreci.com/blog/2017/08/09/faster-rails-eliminating-n-plus-one-queries.html

app/controllers/featured_circuits_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class FeaturedCircuitsController < ApplicationController
+  before_action :authenticate_user!, only: [:feature]
+
+  def index
+    @projects = FeaturedCircuit.includes(:project).map(&:project)
```
tachyons: `includes` is still expensive in terms of memory. Can you try
```ruby
class Project
  has_one :featured_circuit
end
```

```suggestion
    @projects = Project.joins(:featured_circuit)
```

app/controllers/featured_circuits_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def create
+    authorize FeaturedCircuit, :admin?
+    featured_circuit = FeaturedCircuit.find_by(project_id: featured_circuit_params[:project_id])
+
+    if featured_circuit
+      # if featured_circuit exist and checkbox is unmarked
+      featured_circuit.destroy if featured_circuit_params[:featured] == "0"
```
tachyons: this is should go to destroy action

app/controllers/logix_controller.rb
```diff
@@ -2,14 +2,17 @@ class LogixController < ApplicationController
...
+                       .where(project_access_type: "Public", forked_project_id: nil)
+                       .paginate(page: params[:page]).order("id desc").limit(Project.per_page)
+
+    respond_to do |format|
+      format.html # index.html.erb
+      format.json { render json: @projects }
+      format.js
+    end
 
+    @featured_circuits = FeaturedCircuit.includes(:project).take(4)
```
tachyons: use limit here so that Database will handle limiting records, also don't use magic numbers in the code 

https://www.eliotsykes.com/magic-numbers

app/views/user_mailer/featured_circuit_email.html.erb
```diff
@@ -0,0 +1,45 @@
+
```
tachyons: Can we use email frameworks like zurb(https://foundation.zurb.com/emails.html). Because it will be hard to maintain a file with inline css

app/controllers/featured_circuits_controller.rb
ankitkataria: Done, `joins` also saved us a query as compared to `includes`.

app/controllers/logix_controller.rb
ankitkataria: Working on this one.

app/views/user_mailer/featured_circuit_email.html.erb
ankitkataria: Will it be okay if I work on this one is separate PR? All the templates will be changed in that one.

ankitkataria:
@tachyons It made it a little easier for me to handle the form parameters that I was sending on the front end. The usual delete route is `/featured_circuit/:id` I didn't want to complicate the front end code and wanted to use the same route that I was using for creating i.e. `/featured_circuit/`, just with a different request type which is decided [here](https://github.com/CircuitVerse/CircuitVerse/pull/261/files#diff-f5031ccca95fa940b48661646ce56649R74)

ankitkataria:
@tachyons Added tests to the PR! Please review! :gift:

app/views/projects/_form.html.erb
```diff
@@ -1,4 +1,14 @@
 <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.0/jquery.min.js"></script>
+
+<% if policy(@project).admin? && @project.project_access_type == 'Public' %>
```
tachyons: Can we combine this policy to `can_feature?` or similar

ankitkataria: Done :heavy_check_mark:

tachyons:
Thanks

PR #265, opened 2019-03-18
Empty project

### LINKED ISSUE #262: Unable to save an empty project
Saving an empty project produces an error.

Create a new project, click 'save online', enter a project name, and see the error.

(Not working on this issue)

Amansingla97:
Fixes #262
Screenshot 2019-03-18 at 9 18 58 AM
Quick fix. I hope this solves the issue.let me know if something more is requires :)


app/controllers/simulator_controller.rb
```diff
@@ -60,12 +60,23 @@ def create
     @project.author = current_user
 
     data_url = params[:image]
-    jpeg      = Base64.decode64(data_url['data:image/jpeg;base64,'.length .. -1])
-    image_file = File.new("preview_#{Time.now()}.jpeg", "wb")
-    image_file.write(jpeg)
+
+    str =data_url['data:image/jpeg;base64,'.length .. -1];
```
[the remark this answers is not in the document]
Amansingla97: Done

ankitkataria:
@Amansingla97 I think we'll need a similar check in the `update` method as well. You can create a helper in `app/helpers/utils.rb` and use it in both `create` and `update`.

Amansingla97:
@ankitkataria  I think u mean to say app/helpers/simulator_helper.rb.

ankitkataria:
It's up to you to include it wherever you want to, my point is making a helper function out of it. :smile:

app/helpers/simulator_helper.rb
```diff
@@ -1,2 +1,25 @@
...
+      image_file = File.new("preview_#{Time.now()}.jpeg", "wb")
+      image_file.write(jpeg)
+    end
+
+    image_file
+  end
+
+  def check_to_delete (data_url)
+    str = data_url["data:image/jpeg;base64,".length .. -1]
+    if str.to_s.empty?
```
ankitkataria: This is same as `!str.to_s.empty?`

Amansingla97:
Sorry for empty space ,long lone and long method(corrected). Its the first time I am working with codeclimate :P

app/controllers/simulator_controller.rb
```diff
@@ -59,13 +62,14 @@ def create
     @project.name = params[:name]
     @project.author = current_user
 
-    data_url = params[:image]
-    jpeg      = Base64.decode64(data_url['data:image/jpeg;base64,'.length .. -1])
-    image_file = File.new("preview_#{Time.now()}.jpeg", "wb")
-    image_file.write(jpeg)
+    image_file = return_image_file(params[:image])
+
```
ankitkataria: Remove empty line

app/controllers/simulator_controller.rb
```diff
@@ -1,5 +1,5 @@
 class SimulatorController < ApplicationController
-
+  include SimulatorHelper
```
ankitkataria: Is it possible to implement this in a similar fashion as the `Utils` has been used in `collaborations_controller` and `group_members_controller`?

app/helpers/simulator_helper.rb
```diff
@@ -1,2 +1,25 @@
 module SimulatorHelper
+  def return_image_file(data_url)
```
ankitkataria: Look at how `app/helpers/utils.rb` has been implemented. Is it possible to implement it in a similar fashion for the sake of uniformity?

Amansingla97: Obviously it can be placed there but i think keeping it utils sacrifices the readability.

app/helpers/simulator_helper.rb
Amansingla97: But we are not declaring str in controller . So we can't use it directly

app/controllers/simulator_controller.rb
Amansingla97: Yes it is possible but i prefer to do it this way .

app/helpers/simulator_helper.rb
ankitkataria: I agree, I'm asking to implement `SimulatorHelper` is a similar fashion as `Utils` with the helper functions as static methods inside the `SimulatorHelper` and removing the `include` from the `SimulatorController`.

app/helpers/simulator_helper.rb
ankitkataria: That's fine, I was suggesting to change the implementation of the helper. :sweat_smile: to a one liner say - `data_url["data:image/jpeg;base64,".length .. -1].to_s.empty?`
Plus `File.delete(image_url) if data_url["data:image/jpeg;base64,".length .. -1].to_s.empty?` seems pretty readable.

app/controllers/simulator_controller.rb
ankitkataria: Okay, I felt it would have been better for the overall uniformity in the way helper functions were defined and used :slightly_smiling_face:

tachyons:
Add the tests to cover new behavior

app/helpers/simulator_helper.rb
Amansingla97: Bhai data_url bhi define nahi kiya controller mein.

ankitkataria: Does this work? 
```
def check_to_delete (data_url)
    !data_url["data:image/jpeg;base64,".length .. -1.to_s.empty
end
```

Amansingla97: yup this would work :). but i prefer to do it this way .

ankitkataria: Okay :+1: , I was suggesting for a one-line implementation for helper. Pardon me for not being more clear here. :smile:

tachyons: I agree with ankit here. This is a standard ruby idiom

app/helpers/simulator_helper.rb
tachyons: Helpers are supposed to be implemented as modules. But one downside is helpers have global scope. Since this logic is specific to simulator controller it is better to keep it in simulator controller itself as private methods

app/helpers/simulator_helper.rb
Amansingla97: cool

app/helpers/simulator_helper.rb
Amansingla97: cool

satu0king:
Instead of no preview available, can you simply have this image located at - https://circuitverse.org/img/default.png

Actually, this is carrierwave's default image and it should kick in automatically if there is no image associated.

In the case of updating a circuit when there is no image, we can simply not update the image, the previous image is fine. What do you think about this approach? 

image

Amansingla97:
@satu0king 
In the case of updating a circuit 
->When there is no image I don't find any reason to keep the previous image as the user have already decided to delete it.
->Keeping previous image creates unnecessary confusion to user about whether the project is updated (to empty)or not.

tachyons:
Enable rubocop linter in your editor so that you can check the style issues on the fly

satu0king:
> Rather when the project is updated to empty default.png should show up.

It is a matter of preference actually and no it doesn't need to mean the user has deleted the previous circuit, more often than not it could mean that the current circuit which is active is different from the previous circuit. Either approach is acceptable. 

I will test and merge this soon.

app/controllers/simulator_controller.rb
satu0king: We will need a major refactor of all these things probably during the GSoC period. We will need to have guidelines on these things so that there is no confusion.

satu0king:
Thanks!

PR #266, opened 2019-03-18
Add schema.rb to version control

### LINKED ISSUE #224: Handling schema.rb in version control
It is generally recommended to commit `schema.rb` file along with other file systems as mentioned [here]( https://stackoverflow.com/questions/6450263/what-is-the-right-approach-to-deal-with-rails-db-schema-rb-file-in-git)
Will it be a good idea to remove it from `.gitignore` and commit it for our project too?

ankitkataria:
Fixes #224 

Removed schema.rb from .gitignore and checked it in our version control

N.A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ jobs: # a collection of steps
 
       - run:
           name: Database setup
-          command: bin/rails db:migrate
+          command: bin/rails db:schema:load
 
       - run:
           name: Run rspec in parallel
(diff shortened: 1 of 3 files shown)


ankitkataria:
@tachyons If we check `schema.rb` in the version control will it also be a good option to use `db:schema:load` instead of `db:migrate` in our [circle.ci](https://github.com/CircuitVerse/CircuitVerse/blob/master/.circleci/config.yml#L54) config? 
It usually has faster run times that db:migrate and once our tests increase, it will be really useful.

tachyons:
Yes, we used db:migrate for compatibility with mysql and postgres. Now since mysql is removed it is safe to use schema load

PR #280, opened 2019-03-20
Update CONTRIBUTING.md

seannam1218:
- Add space before brackets for consistency. 
- Remove redundancy. "...and try to avoid private messages/conversations so whenever possible, do not take conversations to private channels." -> "Whenever possible, avoid taking conversations to private channels."


armudgal:
Thanks for these corrections

PR #285, opened 2019-03-20
removed the extra postgres gem from the Gemfile

Sadat97:
Fixes #

Removed the extra line 
gem "pg", "~> 1.1.4"

which was creating warnings when installing gems and could l cause errors if you change the version of one of them later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,7 +73,6 @@ gem 'pundit'
 gem "pg", "~> 1.1.4"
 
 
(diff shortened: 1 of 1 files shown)


ankitkataria:
I've also removed the extra pg gem [here](https://github.com/CircuitVerse/CircuitVerse/pull/261/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478) in #261.

PR #292, opened 2019-03-22
Fixes the height reverting issue #272

### LINKED ISSUE #272: Changes to layout height gets reverted after switching to another circuits
Changes to layout height gets reverted after switching to another circuits.

To Reproduce:
1. Create a circuit
2. Click "Edit layout", add a bunch of height, and save the change
3. Click "Edit layout" again, and confirm that the change is saved
4. Switch to another circuit, then switch back again
5. Click "Edit layout", and notice that the change of height is reverted. However, it shouldn't.

This bug only affects the height of a layout.

I'm not working on this issue.

NavonilDas:
Fixes #272

### Demo
converted

### DIFF
--- a/public/js/layout.js
+++ b/public/js/layout.js
@@ -1,7 +1,6 @@
 // Layout.js - all subcircuit layout related code is here
 
 // Function to toggle between layoutMode and normal Mode
-var allLayouts = {}; // Stores all layouts of SubCircuit
 function toggleLayoutMode() {
     if (layoutMode) {
         layoutMode = false;
@@ -93,7 +92,7 @@ function layout_buffer(scope = globalScope) {
     this.yy = yy;
 
     // Assign layout if exist or create new one
-    this.layout = (allLayouts[scope.id]) ? Object.assign({}, allLayouts[scope.id]) : Object.assign({}, scope.layout); //Object.create(scope.layout);
+    this.layout = Object.assign({}, scope.layout); //Object.create(scope.layout);
 
     // Push Input Nodes
     this.Input = [];
@@ -412,7 +411,6 @@ function saveLayout() {
             temp_buffer.Output[i].parent.layoutProperties.y = temp_buffer.Output[i].y;
         }
         globalScope.layout = Object.assign({}, temp_buffer.layout);
-        allLayouts[globalScope.id] = Object.assign({}, temp_buffer.layout);
         toggleLayoutMode();
     }
 }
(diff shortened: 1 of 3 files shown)


satu0king:
I checked the code, actually, it happens because we call `globalScope.fixLayout();` in `switchCircuit()` in data.js. This was actually intended as the user might create multiple inputs/outputs and then delete them but clearly, this restricts the flexibility for the user. 

@fpintos what do you advise, should we remove this feature?

fpintos:
> @fpintos what do you advise, should we remove this feature?

Yes, we should. I was hit more than once on by this one. :-) 
The layout should always be the size the user specifies it.

That being said, we should not lose the auto-grow ability when new pins are added.
In other words, perhaps the function that changes the layout size should grow, but never shrink the layout.

public/js/subcircuit.js
```diff
@@ -190,6 +197,11 @@ SubCircuit.prototype.reset = function() {
         return;
     }
 
+    if(allLayouts[subcircuitScope.id]){
+        // Set the width and height
+        subcircuitScope.layout = Object.assign(allLayouts[subcircuitScope.id]);
+    }
+
     subcircuitScope.layout.height = subcircuitScope.layout.height;
```
fpintos: These two lines seem like a no-op. I wonder if there was a typo in the original code.
Maybe we don't need the allLayouts copy if from here we could get a hold to the updated subcircuit template.
I was looking for a pointer to it but I'm not familiar with this part yet, @satu0king , do you know which variable would be holding all the loaded circuits here?

fpintos: Ah, I did some investigation:

By the end of saveLayout() the globalScope.layout is correct (and it is the same as scopeList[<id>].layout). As we switch circuits, by the time SubCircuit.reset is called the scopeList[<id>].layout.height was already reset.

Looking for where it happens I found switchCircuit() calling Scope.prototype.fixLayout(); this is the one resetting the height.

I think we can fix the issue by just removing Scope.prototype.fixLayout (in logix.js) altogether and its call from switchCircuit() (in data.js).

public/js/layout.js
```diff
@@ -1,6 +1,7 @@
 // Layout.js - all subcircuit layout related code is here
 
 // Function to toggle between layoutMode and normal Mode
+var allLayouts = {}; // Stores all layouts of SubCircuit
```
fpintos: It looks like we don't need this to fix this bug.
Just remove Scope.prototype.fixLayout() and its call from switchCircuit().

satu0king: Yes just removing the function call to `globalScope.fixLayout();` will fix this.

NavonilDas: It would be better to remove just one line instead of adding more lines.

Thanks, @satu0king and @fpintos

PR #297, opened 2019-03-23
sidekiq

### LINKED ISSUE #274: Sidekiq queue in docker
Have queues for sidekiq (not mails) in docker.

`bundle exec sidekiq -q default -q mailers -d` 

No

Amansingla97:
Fixes #274 

->I have dockerized sidekiq.
->Also added the sinatra gem which is needed for the Sidekiq web UI (the dashboard shown below)
Screenshot 2019-03-22 at 2 39 58 PM


satu0king:
Have you ensured that only admins can see this page? 
@vik-y can you check this PR

satu0king:
> @satu0king what's the exact requirement here? Are we planning to run only sidekiq in docker and run everything else outside docker or the requirement is to run sidekiq alongside other dockerized things.
> 
> The thing is the above PR will only work in case when everything is running is dockerized environment.

Yes, we are using docker env for development. In production we are not using it. We do not need the sidekiq dashboard at the moment right now but it would be a nice feature to have. However ofcourse it needs to be restricted.

Amansingla97:
I have added admin rights over sidekiq panel .Do let me know if you want to remove the sidekiq panel.

Gemfile
```diff
@@ -48,6 +48,10 @@ gem 'rails_admin', '~> 1.2'
 # gem 'cancancan', '~>2.0'
 
 gem 'sidekiq'
+gem 'sidekiq-scheduler'
+gem 'sidekiq-unique-jobs'
+gem 'sinatra', require: nil
```
tachyons: Curious why

Gemfile
```diff
@@ -48,6 +48,10 @@ gem 'rails_admin', '~> 1.2'
 # gem 'cancancan', '~>2.0'
 
 gem 'sidekiq'
+gem 'sidekiq-scheduler'
```
tachyons: Are we using this ?

config/routes.rb
```diff
@@ -76,7 +76,14 @@
   end
 
 
-
+  require 'sidekiq/web'
+  require 'sidekiq-scheduler/web'
+  Rails.application.routes.draw do
```
tachyons: No need to add this again, Just add in about draw block

satu0king:
@Amansingla97, please explain what are the dependencies you have included. To resolve the original issue, only the dockerfile needs to be updated. I understand to have the Sidekiq web UI, you need to add gems. Please elaborate.

Gemfile
Amansingla97: I was just playing with sinatra gem which required these . Ill remove sinatra and update it :)
Sorry for late reply

Gemfile
Amansingla97: No its not required ill update the pr soon

Amansingla97:
@satu0king u r right there is no need for extra gems. I was just playing with features of Sinatra gem which have their dependencies on scheduler gem. I'll remove the extra gem and update the pr. If we need Sinatra gem, later on, we can add it again :)

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,9 @@
+sidekiq_config = { url: "redis://redis:6379/0" }
+
+Sidekiq.configure_server do |config|
+  config.redis = sidekiq_config
+end
+
+Sidekiq.configure_client do |config|
+  config.redis = sidekiq_config
+end
```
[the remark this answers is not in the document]
Amansingla97: No, its not necessary in current scenario as we are using docker in development only.
If we decide to set up docker for production in future we can just 
sidekiq_config = { url: ENV['JOB_WORKER_URL'] }
Should I Remove it ?

Issue #288, opened 2019-03-21
Adding a custom context menu.


NavonilDas:
**Describe your feature.**
Adding a custom context menu for the simulation area.

The context menu will contain the following options in the same order as well:
- [ ] Copy
- [ ] Cut
- [ ] Paste
- [ ] Delete  (it deletes the selected circuit element)
- [ ] Undo
- [ ] New Circuit
- [ ] Insert Subcircuit
- [ ] Center Focus

This is the default context menu provided by Chrome.
![screenshot](https://user-images.githubusercontent.com/29132316/54744856-1188fa00-4bee-11e9-923e-07c3c0ac4f8b.jpg)

**Are you working on this? (Yes/No)**
Yes

satu0king:
Good idea. I have updated the issue to be more appropriate.

PR #305, opened 2019-03-23
Context menu feature

NavonilDas:
Added the feature of Context menu #288 

#### Tested on
Chrome ,Firefox , Edge

### Demo
sc

### DIFF
--- a/public/js/UX.js
+++ b/public/js/UX.js
@@ -1,8 +1,83 @@
 var smartDropXX = 50;
 var smartDropYY = 80;
 
+// Object stores the position of context menu
+var ctxPos = {
+    x: 0,
+    y: 0,
+    visible: false,
+};
 
-$(document).ready(function() {
+// Function hides the context menu
+function hideContextMenu() {
+    var el = document.getElementById('contextMenu');
+    el.style = 'opacity:0;';
+    setTimeout(() => {
+        el.style = 'visibility:hidden;';
+        ctxPos.visible = false;
+    }, 200); // Hide after 2 sec
+}
+
+// Function displays context menu
+function showContextMenu() {
+    if (layoutMode) return false; // Hide context menu when it is in Layout Mode
+    $('#contextMenu').css({
+        visibility: 'visible',
+        opacity: 1,
+        top: `${ctxPos.y}px`,
+        left: `${ctxPos.x}px`,
+    });
+    ctxPos.visible = true;
+    return false;
+}
+
+// Function is called when context item is clicked
+// eslint-disable-next-line no-unused-vars
+function menuItemClicked(id) {
+    hideContextMenu();
+    switch (id) {
+    case 0:
+        delete_selected();
+        break;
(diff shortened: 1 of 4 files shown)


satu0king:
Looks nice, please see the original issue, I had updated the requirements.

NavonilDas:
My Bad, I forgot to read the updated issue.

Anyway, the new commit has updated the context menu.
### Demo
sc

public/js/UX.js
```diff
@@ -1,8 +1,78 @@
...
+function menuItemClicked(id) {
+    hideContextMenu();
+
+    if (id === 0) {
+        document.execCommand('copy');
+    } else if (id === 1) {
+        document.execCommand('cut');
+    } else if (id === 2) {
+        // document.execCommand('paste'); it is restricted to sove this problem we use dataPasted variable
+        if (dataPasted) paste(dataPasted);
```
satu0king: Can you rename this to data copied? Also are you sure there is no way around this?

NavonilDas: These are some alternatives pasting:
* I can use the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) however it not compatible with all the browser. 
* I can save the copied data to local storage and then retrieve while pasting, this hack will help us to copy data from the different project.

Which method should i use??

jbox144:
Ok, now this is epic

public/js/UX.js
satu0king: I had a similar conclusion to use localStorage as at least within the browser it will work. However, I would like to confirm that CTR+C still works the same way as before correct? And that doesn't have any browser restrictions right. 

@jbox144 I agree, I expected the context menu to be quite complicated, I was impressed.

NavonilDas: "document.execCommand('paste') is not supported by browsers, because of security concerns as it might enable the script to read sensitive data (like passwords) from the clipboard."

Copy command is supported by every browser.

Cut command is supported by every browser except IE.

after adding the local storage CTR+C will work the same way.

satu0king: How about ctr+v?

NavonilDas: it will also work the same way.

satu0king:
Yes, please go ahead with localstorage.

NavonilDas:
## Here is the Demo

localstorage

public/js/listeners.js
```diff
@@ -272,8 +274,9 @@ function startListeners() {
         } else {
             data = e.clipboardData.getData('text/plain');
         }
-
-        paste(data);
+        // Check if data occurs on local storage else take data from clipboard
+        if (!(localStorage.getItem('clipboardData'))) localStorage.setItem('clipboardData', data);
+        paste(localStorage.getItem('clipboardData'));
```
satu0king: I don't think is a good idea. Once copied, it will persist on localstorage forever so basically clipboard is never used again.

PR #315, opened 2019-03-25
Fixes issue #312

### LINKED ISSUE #312: Preset not working in Sequential elements.
**Describe the bug**
https://circuitverse.org/users/4399/projects/12251
The Preset input does not seem to have any reaction to the sequential system. I have tried it on SR Flip Flop and D Flip Flop but it doesn't seem to bring any change in the output of the Flip Flops

pre1

preset

Though the preset is enabled but the output remains the same.
**To Reproduce**
https://circuitverse.org/users/4399/projects/12251
Here is a demo test. Try changing the input value given to preset. 

**Expected behavior**
Whenever the preset signal it enabled the output Q of the Flip Flop should have a value of 1 whatever the value of input and clock signal may be. 

 - OS: Ubuntu 18
 - Browser: Chrome

No

NavonilDas:
Fixes #312

When PRESET is active The output is set (High) and output doesn't depend on input.
When PRESET and RESET both are active the PRESET gets the priority.

### Working Demo for a T flip flop
tff

### DIFF
--- a/public/js/sequential.js
+++ b/public/js/sequential.js
@@ -57,7 +57,7 @@ TflipFlop.prototype.newBitWidth = function(bitWidth) {
     this.preset.bitWidth = bitWidth;
 }
 TflipFlop.prototype.resolve = function() {
-    if (this.reset.value == 1) {
+    if (this.reset.value == 1 || this.preset.value == 1) {
 
         this.masterState = this.slaveState = this.preset.value || 0;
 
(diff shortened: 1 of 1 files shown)


satu0king:
image

I think you misunderstood the purpose of preset. Preset is the value you want to reset to. So when reset is triggered by default it resets to 0 but if there is a preset value it presets to that value. 

I am not sure if regular flipflops have these things but this is the purpose we use preset for so far. @fpintos  might have some inputs on this. Nevertheless we cannot change this now as many circuits might already be using this feature,

PR #314, opened 2019-03-25
fix extra columns bug

sal2701:
# Fixes #291 

#### Describe the changes you have made in this pr -
I have changed the combinationalAnalysis.js file to take input by eliminating the whitespace rather than splitting after every space. I have changed the .split(' ') to .match(/\S+/g)

### Screenshots of the changes (If any) -

Inputs are "a b " and "1 2 "

## Before
circuitverse-bug
## After
circuitverse-bug solved

PR #318, opened 2019-03-25
Changed split parameter

sal2701:
Changed split parameter to "," from " ". Also used .trim() method to eliminate white space from the ends of entered strings

In reference to #314 

commas

### DIFF
--- a/public/js/combinationalAnalysis.js
+++ b/public/js/combinationalAnalysis.js
@@ -10,8 +10,8 @@ createCombinationalAnalysisPrompt=function(scope=globalScope){
     //console.log("Ya");
     scheduleBackup();
     $('#combinationalAnalysis').empty();
-    $('#combinationalAnalysis').append("<p>Enter Input names separated by spaces: <input id='inputNameList' type='text'  placeHolder='eg. A B C'></p>");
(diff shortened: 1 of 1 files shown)


sal2701:
@satu0king please review. I can't understand why the code is not passing the tests. I can't see why the code is considered as duplication. Thanks

public/js/combinationalAnalysis.js
```diff
@@ -31,16 +31,16 @@ createCombinationalAnalysisPrompt=function(scope=globalScope){
...
 
     var s='<table>';
     s+='<tbody style="display:block; max-height:70vh; overflow-y:scroll" >';
     s+='<tr>';
     for(var i=0;i<inputListNames.length;i++)
-        s+='<th>'+inputListNames[i]+'</th>';
+        s+='<th>'+inputListNames[i].trim()+'</th>';
     for(var i=0;i<outputListNames.length;i++)
-        s+='<th>'+outputListNames[i]+'</th>';
+        s+='<th>'+outputListNames[i].trim()+'</th>';
```
satu0king: Can you perform the trimming before itself on the actual list `inputListNames` and `outputListNames` because we are using the names in several places including the labels for the final input and output circuit elements!

sal2701: Oh I will get it done

satu0king:
If you are creating a new PR, please close the previous one. You can simply update your PR by pushing code to the same branch.

PR #319, opened 2019-03-26
Pgsearch(search on projects)

### LINKED ISSUE #21: Search Engine
Currently we are using google custom search which doesn't do a very good job. The search should search over project names and descriptions.

Amansingla97:
Fixes #21 

Implemented full text search with pg_search.Removed sql specific code with minor changes in UI.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -14,7 +14,7 @@ class Project < ApplicationRecord
   mount_uploader :image_preview, ImagePreviewUploader
 
   include PgSearch
-  pg_search_scope :text_search, against: [:name,:description]
+  pg_search_scope :text_search, against: [:name, :description]
 
   self.per_page = 8
 
@@ -24,7 +24,7 @@ class Project < ApplicationRecord
     if query.present?
       Project.text_search(query)
     else
-      return Project.all
+      Project.all
     end
   end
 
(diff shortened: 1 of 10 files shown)


app/models/project.rb
```diff
@@ -12,10 +13,21 @@ class Project < ApplicationRecord
...
+
   self.per_page = 8
 
   acts_as_commontable
   # after_commit :send_mail, on: :create
+  def self.search(query)
+    if query.present?
+      Project.text_search(query)
+    else
+      Project.all
```
tachyons: Can we have default limit for this as well

Amansingla97: what do you think default limit should be set to ? According to me 50 is good enough.

satu0king:
Looks good to me, will merge once @tachyons approves.

config/initializers/search_paginate_renderer.rb
```diff
@@ -0,0 +1,44 @@
...
+      link_wrapper = link(text, page, link_options.merge(class: 'page-link'))
+      tag :li, link_wrapper, class: 'page-item'
+    else
+      span_wrapper = tag(:span, text, class: 'page-link')
+      tag :li, span_wrapper, class: 'page-item disabled'
+    end
+  end
+
+  def gap
+    tag :li, tag(:i, ELLIPSIS, class: 'page-link'), class: 'page-item disabled'
```
satu0king: @Amansingla97 It says this: `ELLIPSIS` is not defined and hence giving an error in production

PR #324, opened 2019-03-27
Fix keyup and keydown listeners

### LINKED ISSUE #307: Keyboard doesn't work in preview mode.
https://circuitverse.org/users/2418/projects/10802 

This circuit's keyboard doesn't work in preview mode. This will mostly need copying of some listener code from listeners.js to embedListeners.js

No

ankitkataria:
Fixes #307 

A tricky but simple fix. The listeners were already present in `embedListeners.js` but, keyup and keydown events do not behave in the usual manner when they are embedded in the form of an iframe.  They need to be defined on the parent of the current window to capture the actual event.
References - 
- [simple fix](https://stackoverflow.com/a/10942878)
- [using jquery](https://stackoverflow.com/questions/1199075/keydown-event-on-a-iframe)

fix-keyboard-compress

### DIFF
--- a/public/js/embedListeners.js
+++ b/public/js/embedListeners.js
@@ -2,7 +2,7 @@
 // Refer listeners.js
 
 function startListeners() {
-    window.addEventListener('keyup', function(e) {
+    window.parent.addEventListener('keyup', function(e) {
         scheduleUpdate(1);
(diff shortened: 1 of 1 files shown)


satu0king:
Nice

satu0king:
Nice

PR #326, opened 2019-03-27
Fixes Backspace issue #322

### LINKED ISSUE #322: Backspace not working in Combinational Analysis inputs
**Describe the bug**
While entering data in the combinatinalAnalysis inputs, backspace doesn't work. It just stops the input as well. Same goes for delete as well.

**Expected behavior**
Backspace should work as normal

**GIF**

backspace

 - OS: Ubuntu 18.04
 - Browser Mozilla Firefox Quantum
 - Version 66.0.1

No

NavonilDas:
Fixes #322

Fixes the issue with backspace on combinational analysis prompt.

### Screenshots of the changes
backspace

### DIFF
--- a/public/js/listeners.js
+++ b/public/js/listeners.js
@@ -135,7 +135,7 @@ function startListeners() {
             }
         }
 
-        if (e.keyCode == 8 || e.key == "Delete") {
+        if (simulationArea.lastSelected && (e.keyCode == 8 || e.key == "Delete")) {
             delete_selected();
         }
 
--- a/public/js/combinationalAnalysis.js
+++ b/public/js/combinationalAnalysis.js
@@ -8,6 +8,9 @@ var sampleOutputListNames=["X"];
 
 createCombinationalAnalysisPrompt=function(scope=globalScope){
     //console.log("Ya");
+    // Deselect all
+    simulationArea.multipleObjectSelections = [];
+    simulationArea.lastSelected = undefined;
     scheduleBackup();
     $('#combinationalAnalysis').empty();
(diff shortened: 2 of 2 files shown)


public/js/listeners.js
```diff
@@ -135,7 +135,7 @@ function startListeners() {
             }
         }
 
-        if (e.keyCode == 8 || e.key == "Delete") {
+        if (simulationArea.lastSelected && (e.keyCode == 8 || e.key == "Delete")) {
```
satu0king: This is wrong because `simulationArea.lastSelected` could be undefined but `simulationArea.multipleObjectSelections` needn't be an empty list.

NavonilDas: @satu0king, I did this because :

* if the user wants to add a new circuit using combination analysis then there is no need to select previous elements.
* If Something is selected outside the simulation area then it should deselect the elements.

Should I remove the line ??

satu0king: Undo the changes to this line `if (simulationArea.lastSelected && (e.keyCode == 8 || e.key == "Delete")) {`

These changes are correct
```
    simulationArea.multipleObjectSelections = [];
    simulationArea.lastSelected = undefined;
```

NavonilDas: @satu0king if i change that line back to ` if (e.keyCode == 8 || e.key == "Delete") ` the backspace will not work due to the line `$("input").blur();` inside `delete_selected()` as the current iput will lose it focus.

satu0king:
This issue actually is affecting a lot of things including the property menu so we will have to handle it differently.

PR #328, opened 2019-03-27
Insert Subcircuit popup changes

hazemayman1:
### Changes:
+ Replaced the "insert subcircuit" button with a "create subcircuit" one in case no other circuits are already existing in the project.
+ Removed a no longer used reBuild function.
### Screenshots of the changes: 
insertsubcirc


satu0king:
I don't think this is necessary.

PR #333, opened 2019-03-29
unique tags only

### LINKED ISSUE #332: Duplication of tags is not handled
**Describe the bug**
If in the project update page, if the input for tags is `tag tag tag tag`, then this duplication is not removed and 4 taggins are created, while there should only be 1. 

**To Reproduce**
1. Go to project edit page
2. Input `tag tag tag tag` in tags input
2. Click `Update Project`
3. Go to project preview page

**Expected behavior**
Duplicate tags should be removed

**Screenshots**
`project_preview_page`
image

`projects/tags/tag`
image

Platform agnostic

**Additional context**

Yes. @shubhankarsharma00 added tags in PR #185 and might have some points to add. We can work on it together.

shubhankarsharma00:
Fixes #332 

Only added uniq tags when the input was changed to the tags list.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -88,7 +88,7 @@ class Project < ApplicationRecord
   end
 
   def tag_list=(names)
-    self.tags = names.split(",").map do |n|
+    self.tags = names.split(",").map(&:strip).uniq.map do |n|
       Tag.where(name: n.strip).first_or_create!
     end
   end
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -3,4 +3,5 @@
 class Tagging < ApplicationRecord
   belongs_to :project
   belongs_to :tag
+  validates_uniqueness_of :project_id, scope: [:tag_id]
 end
--- a/db/migrate/20190405160400_remove_duplicate_tags.rb
+++ b/db/migrate/20190405160400_remove_duplicate_tags.rb
@@ -1,0 +1,8 @@
+class RemoveDuplicateTags < ActiveRecord::Migration[5.1]
+  def change
+    duplicate_row_values = Tagging.select('project_id, tag_id, count(*)').group('project_id, tag_id').having('count(*) > 1').pluck(:project_id, :tag_id)
+    duplicate_row_values.each do |project_id, tag_id|
+      Tagging.where(project_id: project_id, tag_id: tag_id).order(id: :desc)[1..-1].map(&:destroy)
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -118,7 +118,23 @@ ActiveRecord::Schema.define(version: 20190628090620) do
   create_table "pending_invitations", force: :cascade do |t|
     t.bigint "group_id"
     t.string "email"
+    t.datetime "cr  create_table "subscriptions", force: :cascade do |t|
+    t.string "target_type", null: false
+    t.bigint "target_id", null: false
+    t.string "key", null: false
+    t.boolean "subscribing", default: true, null: false
+    t.boolean "subscribing_to_email", default: true, null: false
+    t.datetime "subscribed_at"
+    t.datetime "unsubscribed_at"
+    t.datetime "subscribed_to_email_at"
+    t.datetime "unsubscribed_to_email_at"
+    t.text "optional_targets"
     t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["key"], name: "index_subscriptions_on_key"
+    t.index ["target_type", "target_id", "key"], name: "index_subscriptions_on_target_type_and_target_id_and_key", unique: true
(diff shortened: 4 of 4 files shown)


app/models/tagging.rb
```diff
@@ -3,4 +3,5 @@
 class Tagging < ApplicationRecord
   belongs_to :project
   belongs_to :tag
+  validates_uniqueness_of :project_id, scope: [:tag_id]
```
satu0king: I think this is unnecessary and in fact might break projects in production as they might not have unique tags now. @tachyons might have more inputs.

tachyons: We can have a migration that deletes duplicate tags

shubhankarsharma00: I don't think it will break the production if there are duplicates already(I tested for that) but we can still have a migration. I can even remove it if you guys wish.

shubhankarsharma00: I manually added 2 duplicate tags and ran the code on my localhost and there was no problem so I think it will be alright if there are pre-existing duplicate tags.

app/models/project.rb
```diff
@@ -73,9 +73,10 @@ def tag_list
   end
 
   def tag_list=(names)
-    self.tags = names.split(",").map do |n|
+    self.tags = names.split(/\s*,\s*/).compact.reject(&:empty?).uniq.map do |n|
```
ankitkataria: Instead of using `compact.reject(&:empty?)` is it possible to use `.select(&:present?)` here?

app/models/project.rb
```diff
@@ -73,9 +73,10 @@ def tag_list
...
       Tag.where(name: n.strip).first_or_create!
     end
+    puts(self.tags)
```
ankitkataria: If this is not required here, can this be removed?

shubhankarsharma00: oops forgot by mistake, it's removed now

app/models/project.rb
shubhankarsharma00: I think we can. Should we?

tachyons: What was the reason for this change ?

ankitkataria: i felt `.compact.reject(&:empty?)` seemed a little repetitive for the task that it was trying to accomplish, so suggested doing `.select(&:present?)`. But it's not necessary, we can keep it the way it is.

tachyons: Not that, I meant splitting using regex

ankitkataria: `names.split(",").map(&:strip)` is one way to go about it.

shubhankarsharma00: > `names.split(",").map(&:strip)` is one way to go about it and `names.split(/\s*,\s*/)` also does the same thing. I think it is to remove the white space characters around `,`

@tachyons yeah this is the reason, earlier I was just stripping whitespace after I split across `,` but now ` tag` and `tag ` will be different if I split it that way.

shubhankarsharma00:
@satu0king @tachyons So what are your final thoughts on this. I'll change the PR accordingly.

app/models/tagging.rb
satu0king: Write the migration. Make sure you make changes only to tags and not anything else.

app/models/project.rb
satu0king: If `names.split(",").map(&:strip)` this can be used, this is preferable as it is definitely more readable.

db/migrate/20190405160400_remove_duplicate_tags.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveDuplicateTags < ActiveRecord::Migration[5.1]
+  def change
+    duplicate_row_values = Tagging.select('project_id, tag_id, count(*)').group('project_id, tag_id').having('count(*) > 1').pluck(:project_id, :tag_id)
+    duplicate_row_values.each do |project_id, tag_id|
+      Tagging.where(project_id: project_id, tag_id: tag_id).order(id: :desc)[1..-1].map(&:destroy)
+    end
+  end
```
satu0king: @vik-y @tachyons can you review if this migration is correct?

tachyons: Difficult to comment without testing it

shubhankarsharma00: @tachyons though there may be no way for you to know for sure whether it will delete the duplicates but from the code alone you can see that it surely isn't going to touch any other critical data (because we are only select and destroying taggings ). Also, if you have time you can run it on a test database (I have and it works fine).

app/models/project.rb
ankitkataria: @shubhankarsharma00  Can we change this to `names.split(",").map(&:strip)` :sweat_smile: ?

shubhankarsharma00: @ankitkataria, thanks for the suggestion this is definitely more readable.

shubhankarsharma00:
Created some mess in his PR :cry: creating new PR.

PR #339, opened 2019-03-30
Make project views unique

### LINKED ISSUE #246: Number of views for a circuit increases upon page reload
**Describe the bug**
Everytime a signed in user who is *not* the creator refreshes the page, the view count increases.

**To Reproduce**
1. Sign in to CircuitVerse. 
2. Go to any project. eg. https://circuitverse.org/users/3/projects/247
3. Keep refreshing the page
4. Views keep increasing

**Expected behavior**
Views should not change upon refresh.

**Screenshots**
rec

**Desktop**
 - OS: any
 - Browser: any
 - Version any

**Additional context**
@satu0king We can store a list of userids who have viewed the project and increment view only if the current user is not in that list. Let me know if this seems like a decent approach. We can periodically flush the list of viewers for each project to save space.

Yes,

animysore:
Fixes #246 

### Changes made :

1. Add `Ahoy` gem 
2. Modify project controller to:   
     - create an Ahoy event whenever a unique visit lands on /project/id as `Visited project <id>`
     - call `@project.increase_views` only if it is a unique visit


AniMysore74:
@satu0king @vik-y 
Circle CI fails on the table creation part. These migrations were generated by the Ahoy library and it throws no error on my environment. Any idea what's happening here?

-- create_table("ahoy_events", {:force=>:cascade, :options=>"ENGINE=InnoDB tDEFAULT CHARSET=utf8"})
rails aborted!
ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near "ENGINE"
LINE 1: ...ter varying, "properties" json, "time" timestamp) ENGINE=Inn...

Gemfile
```diff
@@ -60,6 +60,11 @@ gem 'bootstrap-typeahead-rails'
 # for authorization layer
 gem 'pundit'
 
+# for analytics
+gem 'ahoy_matey'
+
+gem 'mysql2'
```
satu0king: We are not using mysql anymore

app/controllers/projects_controller.rb
```diff
@@ -18,7 +18,10 @@ def get_projects
   # GET /projects/1
   # GET /projects/1.json
   def show
-    @project.increase_views(current_user)
+    if current_visit && !Ahoy::Event.find_by(visit_id: current_visit.id)
+      ahoy.track("Visited project " + @project.id.to_s)
```
satu0king: What does ahoy.track do ?

AniMysore74: Ahoy records **Visits** and **Events**. A **Visit** is recorded automatically whenever someone visits the site. It is persistent for 4 hours by default and recognises users. An **Event** corresponds to a user doing something on a site and is called with `ahoy.track`.  

An Event has a `name`, eg. "Viewed Home page" which is passed to  `ahoy.track`. The `visit_id` corresponding to the Visit on which the Event was created is one of the columns stored and I am using that to check if another Event was logged from the same visit, in which case, don't increment the counter.

Gemfile
AniMysore74: Rails raised an error and said I need to add this to Gemfile. Without this, the server would not start. Do I need to make some change to my local database.yml?

AniMysore74:
@satu0king I removed mysql gem and CircleCI passes as well now

Gemfile
```diff
@@ -60,6 +60,10 @@ gem 'bootstrap-typeahead-rails'
 # for authorization layer
 gem 'pundit'
 
+# for analytics
+gem 'ahoy_matey'
+
+
```
tachyons: ✂️ extra line

app/controllers/projects_controller.rb
tachyons:
```ruby
  Ahoy::Event.exists?(visit_id: current_visit.id)
```

app/controllers/projects_controller.rb
```diff
@@ -18,7 +18,10 @@ def get_projects
...
+      @project.increase_views(current_user)
```
tachyons: This is old logic, we can fetch this from db and set counter_cache for faster access

tachyons:
Also add test for new behaviour

animysore:
@satu0king I've resolved the changes you requested

satu0king:
@ankitkataria since you will be working on this, it might be a good idea for you to check this out.

animysore:
@satu0king The requested changes were made. Could you review once?

satu0king:
> @satu0king The requested changes were made. Could you review once?

Yes, it looks fine. @ankitkataria will be working on this and more so I would prefer if he checks this out in detail and then we can merge it. @ankitkataria is this PR good to merge?

app/controllers/projects_controller.rb
```diff
@@ -18,7 +18,10 @@ def get_projects
   # GET /projects/1
   # GET /projects/1.json
   def show
-    @project.increase_views(current_user)
+    if current_visit && Ahoy::Event.exists?(visit_id: current_visit.id)
```
ankitkataria: @animysore Correct me if I'm wrong, but should this check be `current_visit && !(Ahoy::Event.exists?(visit_id: current_visit.id))`. 

Moreover, it should be `current_visit && !(Ahoy::Event.exists?(visit_id: current_visit.id, name: "Visited project #{@project.id}"))` because if a user visits one project URL and after that open another different one then that is counted as the same visit for ahoy. I verified this locally. Could you also test this once?

animysore: @ankitkataria I see you made a PR with this. Shall I close this ?

ankitkataria: Let's keep this for reference until the other PR gets merged. We can close it after that. :slightly_smiling_face:

PR #343, opened 2019-04-01
A configurable RGB Led Matrix

### LINKED ISSUE #47: Display Matrix

fpintos:
Fixes #47 

### RGB LED Matrix
This PR implements a configurable RGB LED Matrix with the following characteristics:

#### MATRIX SIZE
- Number of rows and columns are independently configurable at design time.
  - Minimum is 1x1
  - Maximum is 128x128

#### LED SIZES
- There are 3 LED sizes, configurable at design time.
  - Small, 1x1.
  - Medium, 2x2.
  - Large, 3x3.

Examples of different matrix configurations and comparison with the size of LEDs:
image

#### PINS
- The matrix provides the following pins:
  - **LEFT**:  R<sub>y</sub>, 1-bit: one pin for each row.
  - **BOTTOM**: C<sub>x</sub>, 1-bit: one pin for each column.
  - **TOP**: CLR<sub>x</sub>, 24-bits (RGB): one pin of color for each column.
  - **RIGHT**:
    - COLOR, 24-bits (RGB): one pin of color. This is on the right side of the matrix.
    - ROW, variable bit width. This is on the right side of the matrix.
    - COLUMN, variable bit width. This is on the right side of the matrix. 

#### USAGE

Pixels can be set in three different ways:

- **Method 1**: Use inputs R and CLR.
  In this mode, pixel at (x,y) takes the color specified by CLR<sub>x</sub> if CLR<sub>x</sub> is connected and R<sub>y</sub> is 1.  
  
  This mode provides bulk editing with multiple colors inputs.  

  This is similar to how some real LED matrices work. A controlling circuit populates the column-color pins, possibly with different colors, and then turns on one or more of the R pins to light up those rows, then moves on to other rows. Inputs that are disconnected (no wires or via tri-state) do not affect pixels in their corresponding columns.

  The main difference is that LEDs in real matrices fade-out, while in this simulation we keep them on as if they had a memory attached to them.

  Example:
  image

- **Method 2**: Use inputs R, C and COLOR.
  In this mode, pixel at (x,y) takes the color specified by COLOR if COLOR is connected, R<sub>y</sub> is 1 and C<sub>x</sub> is 1.

  This mode provides bulk editing with a single color input. In particular, this makes it easy to draw rectangles in the matrix, by selecting the rows and columns that should be updated and turning on the COLOR pin. Using a tri-state in the COLOR pin allows one to control exactly when the color is imprinted, in case the selected rows and columns are changed by different clock cycles.

  Example:
  image

- **Method 3**: Use the ROW, COLUMN and COLOR inputs.
  In this mode, the pixel at (COLUMN,ROW) takes the color specified by COLOR if ROW, COLUMN and COLOR are connected.

  This mode provides single pixel editing.

  Using a tri-state in any of these pins allows one to control exactly when the color is imprinted, in case the pins are changed by different clock cycles.

  Example:
  image

> These methods are not exclusive. A single matrix can have all pins being used at any particular time, but if a pixel is affected by multiple methods then method 1 has lowest priority and method 3 has highest priority. In other words, COLOR has precedence over CLR<sub>x</sub>.

These mechanism allows a controlling circuit to set individual pixels one-by-one or an arbitrary number of pixels at once.

#### Additional Features
In the properties window, there is a button that toggles the matrix grid lines on and off. Turning them off allows the image in the matrix to appear as a normal bitmap.

Example:
image

 
#### Differences from real LED matrices
This element is different from real LED matrices in the following characteristics:
- Individual pixels do not need to be refreshed in order to prevent the light from fading out.
- The color of each pixel is persisted in the project file, so circuit preview can show the results of a previous calculation.

#### Equivalent Circuitry
The following project shows a circuit that is equivalent to a 4x4 RGB LED MATRIX:
https://circuitverse.org/users/1935/projects/14658


public/js/logix.js
```diff
@@ -1,1393 +1,1393 @@
-// Size of canvas
```
fpintos: I need to check why GitHub is having trouble showing this diff...all I did was add the extra component in the list.

blane1257:
The method of including the RGB value with the column inputs is a bit confusing. A single RGB input would be much easier to use.

I also think the pixels could be made smaller. They are already smaller than the square LEDs, but not bu much. Perhaps the pixel size could be configurable by the user.

satu0king:
Looks legit.

fpintos:
> The method of including the RGB value with the column inputs is a bit confusing. A single RGB input would be much easier to use.
> 
> I also think the pixels could be made smaller. They are already smaller than the square LEDs, but not by much. Perhaps the pixel size could be configurable by the user.

I have the LEDs configurable in 3 sizes. By default it comes as "2", which is the sample above (2 units in the grid). "1" is half than that (1 unit in the grid) and "3" is three units of the grid.

As for the pins, this is the usual setup in real display matrices, well, at least in several models, but granted, not all of them. The pins at the top of the matrix controls the color of the LEDs in each column and the pins on the side control which rows will be affected. This allows all pixels of a particular row to be refreshed all at once, and then the driving circuitry moves from one row to the next in quick succession. In real life only one row is light up at a time in most applications and we don't notice because they refresh quickly and our retinas retain the image longer than it takes for the refresh cycle to happen. If we had a single color input we would not be able to simulate those types of matrices; this way we can and we can always put a circuit to distribute the single color to all pins, like in the sample above. All that being said, in our simulator, we will cheat and let the LEDs remain lit even after their row signal goes down, so we can produce and save pixel art without requiring a constant refresh cycle.

public/js/RGBLedMatrix.js
```diff
@@ -0,0 +1,188 @@
+function RGBLedMatrix(x, y, scope = globalScope, rows = 8, columns = 8, ledSize = 2, colors = []) {
```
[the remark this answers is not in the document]
fpintos: needed by CV.

fpintos:
I've updated the rendering function so now the matrix can be bigger without lagging (at least not too much). Now we can do stuff like this, if you know what you are doing:

image

fpintos:
Here's a comparison of the three supported sizes of the LEDs in an 8x8 configuration:

image

Note: the larger padding in size 2 is needed to make the nodes fall in the right position in the grid and still be centered with their respective row & column.

blane1257:
Now I see how controlling the column RGB values independently can be very useful.

Are the X and Y sizes independent? There are many times when the number of columns to be different from the number of rows, like for example an ASCII character display.

public/js/logix.js
fpintos: looks like it was a whitespace change in all lines. I've reverted them.

fpintos:
@ankitkataria , I see you added the codeclimate checks but I seem to be unable to mark issues as won't fix (or anything else). Is that a permission issue in codeclimate itself? Thanks.

ankitkataria:
@fpintos Yes, I think only the member of the org in codeclimate can do so. Also, I think it would be okay to ignore the `Cognitive Complexity` and the `Lines of Code` errors.

fpintos:
@satu0king , the led matrix is ready. let me know if you have any other requests around it.

satu0king:
> I've updated the rendering function so now the matrix can be bigger without lagging (at least not too much). Now we can do stuff like this, if you know what you are doing:
> 
> image

This looks legit.

satu0king:
@fpintos, I will check in the weekend and get back to you but on first look, I don't think there is any.

satu0king:
I went through the implementation and just one word - Amazing! Everything is perfect and complete. Thank you very much. 

The only thing left is to document all this in the user documentation. I have raised an issue here: https://github.com/CircuitVerse/CircuitVerseDocs/issues/128

Issue #345, opened 2019-04-02
BUG in embeded Mode


SubhajitCode:
**Describe the bug**
The Bug appears when We embed the iframe generated in to another website .CircuitVerse Does not take any user input ,mouse click on the button is also broken
**To Reproduce**
1. open a project project window
2. Click on 'Embed' to generate embed iframe tag 
3. paste in your personal Website
4. Some interactivity is lost and circuit verse does not take any mouse input to change the state of button or any other input for that matter.
5.Javascript error

**Expected behavior**
It should take mouse input keyboard input properly
**Screenshots**
![Screenshot_20190402_181303](https://user-images.githubusercontent.com/41638148/55403234-03829400-5573-11e9-9bf4-67e869c36cb7.png)

**Reason**
Violation Of Cross-Origin Policy of Browsers
in the bellow commit https://github.com/CircuitVerse/CircuitVerse/commit/9d117395e369486feb7d8680819777549a8719b7#diff-eb48d01219533c6484ace1ac12e464bf the change had been made .
iframe is trying to access parent windows  keyboard  inputs that violets browsers  cross-origin policy. It can open door to attacks like XSS . So Browser (Checked on Chrome) blocks it.
 

**Additional context**
Reverting the commit will bring back issue https://github.com/CircuitVerse/CircuitVerse/issues/307

**Are you working on this issue? (Yes/No)** Yes

satu0king:
@ankitkataria , I think you worked on this, maybe you have some inputs.

PR #346, opened 2019-04-02
issue 307,345 Fix

### LINKED ISSUE #307: Keyboard doesn't work in preview mode.

SubhajitCode:
Fixes #307  and #345 

removed the "window.parent.addEventListener('keydown', function(e)" line as it was trying to access the parent which may be  in another domain so violets cross-origin policy of browsers 

And Implemented JavaScript Workaround so browser always force  focus on the inserted iframe and we can take our keyboard input from there 

Screenshot_20190402_184056


jbox144:
I read that as "issue 307345 Fix", and thought this was a late april fools day joke.

Idk about you, but I have a suspicion we haven't reached 307345 issues.

SubhajitCode:
Hmm comma in the title was  misleading indeed

satu0king:
Interesting, I will have to check it more closely and see if this is a good fix. The setInterval thing is not a good idea though because then it will affect the host site.

satu0king:
@ankitkataria, I think you worked on these listener changes. Can you take a look.

SubhajitCode:
@satu0king I changed the script tag implementation instead added a mouse event listener to capture users mouse event and focus on the iframe content so that we can take input

satu0king:
> @satu0king I changed the script tag implementation instead added a mouse event listener to capture users mouse event and focus on the iframe content so that we can take input

An iFrame embed should ideally not need any extra listener code injected into the parent. Maybe, what we are trying to do is not actually achievable without injecting extra code. @arkokoley might have more insights.

SubhajitCode:
@satu0king Sir I changed the code.Now no extra code in iframe is required .....

satu0king:
@ankitkataria @Amansingla97 Can you resolve this issue. This PR will solve the main issue. However, the keyboard input stops working. 

I would like to clarify the current code using `window.parent` works fine when embedded internally. Actually, there were issues with comments when we use `window.parent`. However, without `window.parent`, the keyboard wasn't working at all.

satu0king:
I am merging this now as it is quite critical for `docs.circuitverse.org` to work correctly.

PR #353, opened 2019-04-05
Fixes Output Disconnection on changing the Width of subCircuit.

### LINKED ISSUE #349: Output Disconnects on changing the Width of subCircuit.
**Describe the bug**
When we change the width of the subCircuit from layout mode and switch back to the circuit the output nodes are disconnected.

**To Reproduce**

demo

**Expected behavior**
Output Should remain connected.

Yes

NavonilDas:
Fixes #349 

## Demos
demo1
demo2

### DIFF
--- a/public/js/subcircuit.js
+++ b/public/js/subcircuit.js
@@ -263,7 +263,7 @@ SubCircuit.prototype.reset = function() {
             }
         }
     }
-
+    var prevConn = this.outputNodes;
     this.outputNodes = []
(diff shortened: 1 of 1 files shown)


satu0king:
While this might look desirable in most cases, it is undesirable in some cases. The user typically shouldn't change the layout so it is not much of a hindrance to connect the nodes. However, in some cases, not disconnected can cause the circuit wiring to end up in a bad state. So I will say this feature shouldn't be there.

PR #361, opened 2019-04-09
Delete Unnecessary SetTimeOut from application.html.erb

puskuruk:
Fixes #
Deleted unnecessary setTimeOut function.

This function gave error to user where there is no alert. If it's necessary it must be work only in where there is alert because it was hide alerts to user.

Screenshot from 2019-04-09 19-45-35

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -44,10 +44,6 @@
 <%= render 'logix_assets/nav'%>
 <%= yield %>
 
-<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


tachyons:
Can you handle the case when alert does not exist instead of removing this altogether. Or do you have better suggestions to hide alerts

puskuruk:
@tachyons  https://github.com/CircuitVerse/CircuitVerse/pull/362

tachyons:
Duplicates #362

PR #362, opened 2019-04-10
Patch 2

puskuruk:
#### Fixes

If there is a alert it will be hidden to user. Now we are checking there is alert or not.

I changed logic and now we are checking is there alert after I readed comment on https://github.com/CircuitVerse/CircuitVerse/pull/361


tachyons:
Remove unrelated commits

puskuruk:
@tachyons yeap! I changed branch and now there is only my commit.

satu0king:
Thanks

PR #380, opened 2019-04-30
Space split to Comma split in combinatinalAnalysis.js

### LINKED ISSUE #291: Extra columns in Combinational Analysis due to an extra space in input
**Description of the bug**
In the Combinational Analysis, while entering the input number of columns vary because of a space at the end of the input.

**Steps to reproduce the behavior:**
1. Go to Simulator.
2. Click on Combinational Analysis and Enter "a b c "
3. The generated table has 4 columns instead of the expected 3 columns.

**Expected behavior**
4 columns

**Screenshots**

Circuitverse_Error

4th column is the error.

Yes, I am working on this issue

sal2701:

I have changed the split parameter to "," from a " " so that names like "In A, In B" can be used.
In the code, a list was passed with the names as strings. This was causing a problem, so I mapped each string to an integer and sent the integer array forward and this caused no problems. i have used  random integer 7*i + 13 for each "i". The strings entered have been trimmed as well.

Comma_Input

### DIFF
--- a/public/js/combinationalAnalysis.js
+++ b/public/js/combinationalAnalysis.js
@@ -10,8 +10,8 @@ createCombinationalAnalysisPrompt=function(scope=globalScope){
     //console.log("Ya");
     scheduleBackup();
     $('#combinationalAnalysis').empty();
-    $('#combinationalAnalysis').append("<p>Enter Input names separated by spaces: <input id='inputNameList' type='text'  placeHolder='eg. A B C'></p>");
-    $('#combinationalAnalysis').append("<p>Enter Output names separated by spaces: <input id='outputNameList' type='text'  placeHolder='eg. X Y Z'></p>");
+    $('#combinationalAnalysis').append("<p>Enter Input names separated by commas: <input id='inputNameList' type='text'  placeHolder='eg. In A, In B'></p>");
+    $('#combinationalAnalysis').append("<p>Enter Output names separated by commas: <input id='outputNameList' type='text'  placeHolder='eg. Out X, Out Y'></p>");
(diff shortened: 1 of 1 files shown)


sal2701:
@satu0king please review

jbox144:
Use four tabs to indent code :P

sal2701:
@jbox144 what? :laughing:

public/js/combinationalAnalysis.js
```diff
@@ -32,8 +34,11 @@ createCombinationalAnalysisPrompt=function(scope=globalScope){
 }
 function createBooleanPrompt(inputListNames,outputListNames,scope=globalScope){
 
-    inputListNames=inputListNames||(prompt("Enter inputs separated by space").split(' '));
-    outputListNames=outputListNames||(prompt("Enter outputs separated by space").split(' '));
+    inputListNames=inputListNames||(prompt("Enter inputs separated by commas").split(','));
+    outputListNames=outputListNames||(prompt("Enter outputs separated by commas").split(','));
+    outputListNamesInteger=[];
+    for (var i = 0; i < outputListNames.length; i++)
+        outputListNamesInteger[i] = 7*i + 13;//assigning an integer to the value, 7*i + 13 is random
```
satu0king: Why use something like this why can it simply be the index i itself?

sal2701: Idk I just felt that the index i might just repeat somewhere. No good reason. Just a random feeling, so I put in this random expression

satu0king: Okay

PR #390, opened 2019-05-30
Init assignment grading feature

ankitkataria:
### Assignment grading feature. 

Feature description - 

- Go to assignment creation page for group. 
- Select grading scale - letter(A-F) or percent(1-100) or no scale(ungraded assignment), once the grading scale is set it cannot be changed

#### Assignment creation - 
image

#### Group member's view - 
image

- Once the deadline of assignment has been passed, the grading form is shown along side the submitted projects for the mentor

#### Mentors view
image

- Once the mentor is done grading assignments then the mentor can finalize grades, after which the grades will be visible to the group members. Grades can still be viewed by the mentor or exported in the form of a csv file. 

#### After finalization of grades - 
image

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -14,16 +14,18 @@ class GradesController < ApplicationController
     @grade.assignment_id = grade_params[:assignment_id]
     @grade.user_id = current_user.id
 
-    @grade.save
+    render json: { error: "Grade is invalid" },
+      status: 400 unless @grade.save
   end
 
   def destroy
+    project_id = @grade&.project_id
     if @grade.present?
       authorize @grade, :mentor?
       @grade.destroy
     end
 
-    render json: {}, status: 200
+    render json: { project_id: project_id }, status: 200
   end
 
   private
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -30,7 +30,7 @@ class GradesController < ApplicationController
 
   def to_csv
     respond_to do |format|
-      format.csv { send_data Grade.to_csv(params[:assignment_id]),
+      format.csv { send_data Grade.to_csv(params[:assignment_id].to_i),
         filename: "assignment-grades.csv" }
     end
   end
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -28,13 +28,16 @@ class GradesController < ApplicationController
     render json: { project_id: project_id }, status: 200
   end
 
+  def to_csv
+    respond_to do |format|
+      format.csv { send_data Grade.to_csv(params[:assignment_id]),
+        filename: "assignment-grades.csv" }
+    end
+  end
+
   private
     def grade_params
       params.require(:grade).permit(:project_id, :grade, :assignment_id)
-    end
-
-    def grade_udpate_params
-      params.require(:grade).permit(:grade, :user_id, :project_id)
     end
 
     def set_grade
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+class GradesController < ApplicationController
+  before_action :authenticate_user!
+  before_action :set_grade, only: [:create, :destroy]
+
+  def create
+    @grade = @grade.present? ? @grade : Grade.new(assignment_id: grade_params[:assignment_id])
+
+    authorize @grade, :mentor?
+
+    @grade.project_id = grade_params[:project_id]
+    @grade.grade = grade_params[:grade]
+    @grade.assignment_id = grade_params[:assignment_id]
+    @grade.user_id = current_user.id
+
+    @grade.save
+  end
+
+  def destroy
+    if @grade.present?
+      authorize @grade, :mentor?
+      @grade.destroy
+    end
+
+    render json: {}, status: 200
+  end
+
+  private
+    def grade_params
+      params.require(:grade).permit(:project_id, :grade, :assignment_id)
+    end
+
+    def grade_udpate_params
+      params.require(:grade).permit(:grade, :user_id, :project_id)
+    end
+
+    def set_grade
+      @grade = Grade.find_by(project_id: grade_params[:project_id])
+    end
+end
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -41,6 +41,7 @@ class GradesController < ApplicationController
     end
 
     def set_grade
-      @grade = Grade.find_by(project_id: grade_params[:project_id])
+      @grade = Grade.find_by(project_id: grade_params[:project_id],
+       assignment_id: grade_params[:assignment_id])
     end
 end
--- a/app/models/grade.rb
+++ b/app/models/grade.rb
@@ -33,8 +33,8 @@ class Grade < ApplicationRecord
 
     def self.to_csv(assignment_id)
       attributes = %w[email name grade]
-      group_members = User.where(id: GroupMember.where(group_id:
-        Assignment.find_by(id: assignment_id)&.group_id).pluck(:user_id))
+      group_members = User.joins(group_members: :assignments)
+        .where(group_members: { assignments: { id: assignment_id } })
       submissions = Project.where(assignment_id: assignment_id)&.includes(:grade, :author)
 
       CSV.generate(headers: true) do |csv|
--- a/app/models/grade.rb
+++ b/app/models/grade.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class Grade < ApplicationRecord
-  belongs_to :project, dependent: :destroy
+  belongs_to :project
   belongs_to :grader, class_name: "User", foreign_key: :user_id
-  belongs_to :assignment, dependent: :destroy
+  belongs_to :assignment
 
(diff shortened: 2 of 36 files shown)


app/controllers/grades_controller.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class GradesController < ApplicationController
+  before_action :authenticate_user!
+  before_action :set_grade, only: [:create, :destroy]
+
+  def create
+    @grade = @grade.present? ? @grade : Grade.new(assignment_id: grade_params[:assignment_id])
```
tachyons: If grade is already present, then it should got to update action, not create

ankitkataria: I chose to write one route for creating and updating grades so that in the front end it wasn't necessary to know whether the grade for the project had been created or not to decide which route to call. Also, I felt it was easier to check if the grade was created in the back end as compared to the front end. 😅

app/models/grade.rb
```diff
@@ -0,0 +1,49 @@
...
+    def assignment_project
+      if project&.assignment_id != assignment&.id
+        errors.add(:project, "is not a part of the assignment")
+      end
+    end
+
+    def self.to_csv(assignment_id)
+      attributes = %w[email name grade]
+      group_members = User.where(id: GroupMember.where(group_id:
+        Assignment.find_by(id: assignment_id)&.group_id).pluck(:user_id))
```
tachyons: Simplify the query

app/views/assignments/_form.html.erb
```diff
@@ -17,6 +17,18 @@
     <%= form.text_field :name, id: :assignment_name %>
   </div>
 
+  <% if allow_scale_change %>
```
tachyons: You can do something like `if @grade.new_record?`

app/views/assignments/show.html.erb
```diff
@@ -38,6 +47,10 @@
 
       <%end%>
 
+      <%if policy(@assignment).admin_access? && @assignment.graded? && @assignment.deadline - Time.now < 0 %>
```
tachyons: Move this logic to policy `can_export`

app/models/grade.rb
ankitkataria: Does this work? 
`User.where(id: Assignment.find_by(id: assignment_id)&.group&.group_members.pluck(:user_id))` ?

config/routes.rb
```diff
@@ -7,6 +7,10 @@
     resources :assignments
   end
 
+  resources :grades, only: [:create, :update]
+  delete '/grades', to: 'grades#destroy'
```
tachyons: Add destroy in above resource instead

ankitkataria: I had tackled the delete route in a similar fashion in featured circuit management as well -  https://github.com/CircuitVerse/CircuitVerse/pull/261#issuecomment-478684132
😅

config/routes.rb
```diff
@@ -7,6 +7,10 @@
...
+  get '/grades/to_csv/:assignment_id', to: 'grades#to_csv', as: 'grades_to_csv'
```
tachyons:
```ruby   
resources :grades, only: [:create, :update, :delete] do
  member do
    get :to_csv, param: :assignment_id
  end
end
```

db/migrate/20190526125442_add_grades.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :grades do |t|
+      t.string :grade
+      t.timestamps
+    end
+
+    add_reference :grades, :project, foreign_key: true, on_delete: :cascade
+    add_reference :grades, :user, foreign_key: true
+    add_reference :grades, :assignment, foreign_key: true, on_delete: :cascade
+    add_column :assignments, :grading_scale, :integer, default: 0
+    add_column :assignments, :grades_finalized, :boolean, default: false
```
tachyons: Add above two changes in separate migration

spec/models/grade_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  describe "validations" do
+    before do
+      @grade = FactoryBot.build(:grade, project: @assignment_project, grader: @mentor,
+          grade: "A", assignment: @assignment)
+    end
+
+    context "grading scale" do
+      it "valides grading scale" do
```
tachyons: Add more test cases for validity

app/models/grade.rb
```diff
@@ -0,0 +1,49 @@
...
+  belongs_to :project
+  belongs_to :grader, class_name: "User", foreign_key: :user_id
+  belongs_to :assignment
+
+  validates :grade, :user_id, :project_id, :assignment_id, presence: true
+  validate :grading_scale, :assignment_project
+
+  private
+    def grading_scale
+      valid = case assignment&.grading_scale
```
tachyons: We are validating assignment id, ie assignment record must be always present, so no need of `&` here

app/models/grade.rb
```diff
@@ -0,0 +1,49 @@
...
+              when "no_scale"
+                false
+              when "letter"
+                grade&.match(/^(A|B|C|D|E|F)$/).present?
```
tachyons: Make regex as constant

app/controllers/grades_controller.rb
tachyons: But logic will be messy in the future, We have to keep the code maintainable in long term

ankitkataria: I feel that the front end logic might get messy for checking in between post, put, delete requests for grades and changing them as and when we create/update/delete grades. Is there an easy way to handle it on the front end?

app/models/grade.rb
tachyons: I haven't tested, code should be something like this
```ruby
class Group
 has_many: :assignments
end
class GroupMember
 belongs_to: :group
 has_many: :assignments, through: :group
end
group_members = User.joins(group_members: {assignments: {id: assignment_id } )
```

So that we will get the result with single SQL query

config/routes.rb
tachyons: sure

app/models/grade.rb
ankitkataria: I'll try this.

tachyons:
@satu0king WDYT about the user interface ?

app/controllers/grades_controller.rb
satu0king: I think what @ankitkataria has done here is fine. Ideally, it should be separate but it is not a very big deal.

satu0king:
Everything looks good, I didn't go into details of the code. 
I don't see why there needs to be a different percentage/grade option. I tried entering A- as the grade and it didn't work as you expect the grade to be A, B, C only. Moreover, some colleges have other grades such as O grade.

Would it be possible to have the following: 

The grade is just a string, the mentor can write anything in it. Also, what is the purpose of the finalized grade thing, is it to lock the assignment to further changes? The export of CSV looks great. 

Also, what is the purpose of the user foreign key in the grade table? is it to record the mentor who graded?

ankitkataria:
> What is the purpose of the user foreign key in the grade table? is it to record the mentor who graded?

Yes, only when the grades are finalized, they'll be visible to the students.

app/policies/assignment_policy.rb
```diff
@@ -31,4 +31,12 @@ def reopen?
     raise CustomAuthError.new("Project is already open") if assignment.status == "open"
     true
   end
+
+  def can_be_graded?
+    admin_access? && assignment.graded? && assignment.deadline - Time.now < 0
```
tachyons: Time.current

app/views/assignments/show.html.erb
```diff
@@ -59,9 +68,8 @@
     <div class="col-3">
       <div class="overflow:scroll">
         <div class="list-group" id="list-tab" role="tablist" id="submissionTable" style="max-height:600px;overflow:hidden;overflow-y:scroll">
-		  <% @project_order = @assignment.projects.sort_by {|p| p.author.name} %>
-          <% @project_order.each do |project| %>
-              <a class="list-group-item list-group-item-action" id=<%= project.id%> ><%= project.author.name  %></a>
+          <% @assignment.project_order.each do |project| %>
+              <a class="list-group-item list-group-item-action" id=<%= project.id%> data-grade=<%= ProjectDecorator.new(project).grade_str %> ><%= project.author.name  %></a>
```
tachyons: declare decorators in controller

spec/models/assignment_spec.rb
```diff
@@ -46,5 +46,21 @@
...
+
+        (1..3).each do
+          @projects << FactoryBot.create(:project, assignment: @assignment,
+            author: FactoryBot.create(:user))
+        end
+      end
+
+      it "gives sorted assignment projects" do
+        expect(@assignment.project_order.map { |p| p.author.name })
+        .to eq(@projects.map { |p| p.author.name }.sort)
```
tachyons: 🤔 testing same thing ?

ankitkataria: `@assignment.project_order.map { |p| p.author.name }` - this will be a list of author names of the projects `@assignment.project_order`
`@projects.map { |p| p.author.name }.sort` - this will be a *sorted* list of author names from the projects associated with the assignment. So this checks if the projects from `@assignment.project_order`are sorted by author name 😅

ankitkataria:
Is this PR good to go? @satu0king 😅

PR #392, opened 2019-06-10
Restrict elements

ankitkataria:
### Feature description: 

- The mentor will add a new assignment for the group when he'll be asked to restrict elements if required. 
image

- Students, on starting their work, can view a warning when they try to use an element that has been restricted by the mentor.
image

- All the restricted elements used in the current scope is shown in the simulation area as so - 
image

It's updation on cut/paste/delete/addition has been handled. 

- On saving the circuit, backend check has been added [here](https://github.com/CircuitVerse/CircuitVerse/pull/392/commits/fcd4043ba4a0834932b60bb28b5218b445af4e1e#diff-8a6a2318d8b0a77e14beb2eaca8e6346R21) to ensure that `restrictedCircuitElementsUsed` key of data is populated correctly and the student has not tampered with the javascript code.  

- Once the project is saved, restricted circuit elements used is visible in the preview of the circuit. In this way, it is also visible to the mentor when he/she'll be grading the project. 
image

### Todo - 
- [x] Add tests for backend changes
- [x] Handle edit of restricted elements
- [x] Rebase and test changes when #390 is merged


db/migrate/20190609174303_add_circuit_elements.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :circuit_elements do |t|
+      t.string :name
+      t.string :image
+      t.integer :category
+      t.timestamps
+    end
+
+    create_table :assignments_circuit_elements, id: false do |t|
+      t.belongs_to :circuit_element, index: true
+      t.belongs_to :assignment, index: true
```
satu0king: So, I take it that if there is a record of a circuit element in `assignments_circuit_elements`, the usage of the circuit element will not be allowed. This way be default everything is allowed.

db/schema.rb
```diff
@@ -81,6 +98,18 @@
...
+    t.string "grade"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.bigint "project_id"
+    t.bigint "user_id"
+    t.bigint "assignment_id"
+    t.index ["assignment_id"], name: "index_grades_on_assignment_id"
+    t.index ["project_id"], name: "index_grades_on_project_id"
+    t.index ["user_id"], name: "index_grades_on_user_id"
+  end
```
satu0king: Why is grades table change here?

ankitkataria: Once the previous PR gets merged, I'll rebase this branch and this should go away then.

db/migrate/20190609174303_add_circuit_elements.rb
ankitkataria: Yeah, I was actually thinking of implementing it the other way around but I think this will be better. 😁

satu0king: Good, because then by default nothing should change .. 
Also please keep in mind that as we discussed after the database changes, all changes should happen in the front-end only. The server will not be doing any validations of the same.

ankitkataria: > The server will not be doing any validations of the same. That is the front-end can verify if there is any violation of the circuit elements usage

I think both the front end and backend should have validations on the circuits elements, so that if the front end validations fail, then there are still there in the backend. In the backend the validations will only be there when the user decides to save the circuits, and maybe when I have to show to the mentor which all restricted elements the student used.

app/views/simulator/edit.html.erb
```diff
@@ -100,93 +106,21 @@
...
+          <div class="panelHeader"><%= category %> </div>
+            <div class="panel" style="overflow-y:hidden">
+          <% elements.each do |element| %>
+              <div class="icon logixModules" id="<%= element.name %>" >
+                <img src=<%= element.image.url %>>
+                <p class="img__description"> <%= element.name %></p>
+              </div>
+          <% end %>
+            </div>
+        <% end %>
```
satu0king: This might actually hurt future attempts to decouple the simulator and the backend.

PR #401, opened 2019-07-07
WIP: Subcircuit2

abdullah-mohamed:
WIP:

Added layoutDraw() functions for Seven segments, Hex display, Sixteen segment, Counter, Random, Digital Leds, Variable Leds, RGBLed, SquareRGBLed.
Now All the elements shows inside layoutMode.

### DIFF
--- a/public/js/Counter.js
+++ b/public/js/Counter.js
@@ -100,7 +100,14 @@ Counter.prototype.customDraw = function () {
     lineTo(ctx, -20, 15, xx, yy, this.direction);
     ctx.stroke();
 }
+Counter.prototype.layoutProperties = {
+    rightDimensionX : 20,
+    leftDimensionX : 0,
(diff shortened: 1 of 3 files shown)


satu0king:
The renders look good! good job. Everything seems to be working fine. Only the squareRGB is not moving once placed. Please check that once. 

Here are some things to improve later, not immediate concern now
1. Hover is not being detected correctly, the bug will be in `layoutUpdate()` and might be tricky to debug
2. Hover needs to be implemented for all elements but no immediate requirement
3. The seven segment display need to be able to stack next to each other without overlay, we can fix this later as well. 
image

> Only the squareRGB is not moving once placed.
Once this is fixed, this PR can be merged

abdullah-mohamed:
SquareRGB Led is fixed.

PR #402, opened 2019-07-07
Search improvements

ankitkataria:
#### Improvement Description

-  search is now supported for both users and projects through pg_search
- projects can now also be searched against author names and associated tags
- users are searched against the name, educational institute and country

#### UI Changes

- Navigation bar

image

- Search pages

image

### Testing Solr locally

*Note:*  Make sure you have java installed locally. 

```
bundle install
bundle exec rake sunspot:solr:start
bundle exec rake sunspot:solr:reindex
export CIRCUITVERSE_USE_SOLR=true
```


satu0king:
image
This can be improved a bit ?

satu0king:
Once @tachyons reviews this PR, we canmerge

ankitkataria:
> This can be improved a bit ?

Yup, this was the default one. I'll add some custom styles to it.

app/controllers/users/logix_controller.rb
```diff
@@ -48,6 +52,10 @@ def groups
 
   private
 
+    def users_scope
+      UsersQuery.new.search_name(params[:q]).paginate(page: params[:page], per_page: 5)
```
tachyons: Move default per_page to a constant

app/views/users/logix/search.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<%= render "logix_assets/search_bar" %>
+
+<% if @users.empty? %>
+  <div class="container">
+    <h4 class="results-head">No Results Found</h4>
+  </div>
+<% else %>
+  <div class=""></div>
+  <div class="container">
+    <% @users.map { |user| ProfileDecorator.new(user) }.each do |profile| %>
```
tachyons: make @users as decorated users in controller itself

public/js/search.js
```diff
@@ -0,0 +1,24 @@
...
+    if (window.location.href.includes('search')) {
+        $('#nav-search').remove();
+    }
+
+    if (window.location.href.includes('users/search')) {
+        $('#search_type').val('Users');
+        $('#search-box').attr('action', '/users/search');
+    }
+
+    $('#search_type').change((e) => {
```
tachyons: Instead of dynamically changing endpoint, we can just pass resource as parameter

/search?query='Ankit'&resource=Project

spec/controllers/projects_controller_spec.rb
```diff
@@ -133,6 +133,61 @@
...
+            project_access_type: "Public")
+          get projects_search_path, params: { q: "basic gates" }
+          expect(response.status).to eq(200)
+          expect(response.body).to include "Full adder using basic gates"
+          expect(response.body).to include "Half adder using basic gates"
+          expect(response.body).not_to include "Full adder using half adder"
+        end
+      end
+
+      it "should get no results" do
```
tachyons: when ?

public/js/search.js
ankitkataria: Currently I have moved the search routes of users and projects to their respective controller(and have their own routes). They also have their own specific query implementations. If we move to the route you suggested, how are we planning to detect which query to call in the backend? Are we planning to have a generic query implementation for both users and projects?

tachyons: I think we can have generic implementation

ankitkataria: Currently, Users and Projects are the only two models that we have to search, will a generic query implementation be helpful? 🤔

tachyons: you can use dependency injection to specify the search query based on resource

ankitkataria: So should we have a separate search controller with this route and decide the query to use on the basis of the resource key passed?

app/views/users/logix/search.html.erb
ankitkataria: Doing this in the controller creates problems with the paginator 😓

public/js/search.js
ankitkataria: Moved search logic to the search controller. 😄

ankitkataria:
@tachyons Moved the when condition to a SearchHelper. Let me know if we can merge. 😁

PR #410, opened 2019-07-17
Subcircuit label

abdullah-mohamed:
WIP :

The user is no able to show the label of any element in the subcircuit

### DIFF
--- a/public/js/logix.js
+++ b/public/js/logix.js
@@ -1497,8 +1497,9 @@ CircuitElement.prototype.subcircuitMutableProperties = {
     }
 }
 
-CircuitElement.prototype.toggleLabelInLayoutMode = function() {
-    this.subcircuitMetadata.showLabelInSubcircuit = !this.subcircuitMetadata.showLabelInSubcircuit;
+CircuitElement.prototype.toggleLabelInLayoutMode = function(val) {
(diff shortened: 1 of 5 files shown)


public/js/logix.js
```diff
@@ -1481,6 +1482,26 @@ CircuitElement.prototype.layoutProperties = {
...
 };
 
+// Default mutable properties in layout mode 
+// Should we define it for subcircuit elements only??
+CircuitElement.prototype.subcircuitMutableProperties = {
+    "label": {
+        name: "Label: ",
+        type: "text",
+        func: "setLabel"
+    },
```
satu0king: No the label has to be common with the normal mode. The label direction however should be different. This makes sense as the circuit design is different from the subcircuit layout design but labels will not change.

abdullah-mohamed: As the function `setLabel` is the same. Whenever we change the label in either mode, it changes the original label. I can remove it from `CircuitElement.prototype.subcircuitMutableProperties` and this will remove it from showing in layout mode.

satu0king: Yes please remove it.

abdullah-mohamed: removed

PR #419, opened 2019-07-28
Add subscription of mails for users

ankitkataria:
### Feature description

- `/custom_mails/new` - this route is only available to admins and can be used to send custom emails to subscribed users
image

- Upon creation a preview of the email is shown with options to edit or send it right away
image

- Users can unsubscribe to e-mails from the profile settings page
image

#### Notes -

- Users are processed in batch sizes of 1000, currently, there are 7k users.
- User table export is only 2MB in size hence there is no significant load on memory while loading the user data to send mails.

#### TODO - 

- [x] Add tests to increase test coverage
- [x] Provide a way to unsubscribe from mails from with the sent mail

### DIFF
--- a/app/controllers/custom_mails_controller.rb
+++ b/app/controllers/custom_mails_controller.rb
@@ -46,6 +46,9 @@ class CustomMailsController < ApplicationController
   end
 
   def send_mail
+    @mail.sent = true
+    @mail.save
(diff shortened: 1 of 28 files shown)


satu0king:
Amazing job!

PR #420, opened 2019-07-29
checking clicks

abdullah-mohamed:
WIP : catch clicks on the elements on surface of the subcircuit

### DIFF
--- a/public/js/UX.js
+++ b/public/js/UX.js
@@ -143,7 +143,7 @@ function showProperties(obj) {
     if (layoutMode){ 
         $('#moduleProperty').show();
         $('#moduleProperty-inner').append("<div id='moduleProperty-header'>" + obj.objectType + "</div>");
-        if (obj.subcircuitMutableProperties) {
+        if (obj.subcircuitMutableProperties && obj.canShowInSubcircuit) {
             for (attr in obj.subcircuitMutableProperties) {
                 var prop = obj.subcircuitMutableProperties[attr];
                 if (obj.subcircuitMutableProperties[attr].type == "number") {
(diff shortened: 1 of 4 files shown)


public/js/subcircuit.js
```diff
@@ -290,6 +290,17 @@ SubCircuit.prototype.reset = function() {
...
 SubCircuit.prototype.click = function() {
     // this.id=prompt();
+    for(let i = 0; i < this.localScope["Button"].length; i++){
+        if (simulationArea.mouseDownX === this.x + this.localScope["Button"][i].subcircuitMetadata.x 
+            && simulationArea.mouseDownY === this.y + this.localScope["Button"][i].subcircuitMetadata.y){
+                this.lastClickedElement = i;
+                this.localScope["Button"][i].wasClicked = true;
+            }
+    }    
+}
```
satu0king: You cannot check if the mouse is hovering with by just checking if x and y match. You will need to do it similar to `CircuitElement.prototype.isHover()`

satu0king: Why hardcode with Button only? other elements can be clicked as well.

public/js/subcircuit.js
```diff
@@ -290,6 +290,17 @@ SubCircuit.prototype.reset = function() {
...
+
+SubCircuit.prototype.releaseClick = function(){
+    this.localScope["Button"][this.lastClickedElement].wasClicked = false;
```
satu0king: `this.lastClickedElement` can be undefined if no button was clicked.

public/js/UX.js
```diff
@@ -159,13 +159,13 @@ function showProperties(obj) {
...
-            s.val(obj.subcircuitMetadata.labelDirection);
-            $('#moduleProperty-inner').append("<p>Label Direction: " + $(s).prop('outerHTML') + "</p>");
-        }
+            if (!obj.labelDirectionFixed) {
+                if(!obj.subcircuitMetadata.labelDirection) obj.subcircuitMetadata.labelDirection = obj.labelDirection;
+                var s = $("<select class='objectPropertyAttribute' name='newLabelDirection'>" + "<option value='RIGHT' " + ["", "selected"][+(obj.subcircuitMetadata.labelDirection == "RIGHT")] + " >RIGHT</option><option value='DOWN' " + ["", "selected"][+(obj.subcircuitMetadata.labelDirection == "DOWN")] + " >DOWN</option><option value='LEFT' " + "<option value='RIGHT'" + ["", "selected"][+(obj.subcircuitMetadata.labelDirection == "LEFT")] + " >LEFT</option><option value='UP' " + "<option value='RIGHT'" + ["", "selected"][+(obj.subcircuitMetadata.labelDirection == "UP")] + " >UP</option>" + "</select>");
+                s.val(obj.subcircuitMetadata.labelDirection);
+                $('#moduleProperty-inner').append("<p>Label Direction: " + $(s).prop('outerHTML') + "</p>");
+            }
+        }
```
abdullah-mohamed: I didn't change any code here. I just added the second if statement inside the scope of the first one. The right side bar was printing "undefined" when a layout node is clicked, This will limit it.
The solution is elaborated in a new comment inside "layout.js" file just under the constructor function of "layout node" element.

public/js/layout.js
```diff
@@ -351,8 +351,34 @@ function layoutNode(x, y, id, label = "", type, parent) {
     this.prev = 'a';
     this.count = 0;
     this.parent = parent;
+    this.objectType = "Layout Node";    
+}
```
abdullah-mohamed: The next comment explains the reason behind the "undefined" bug in layout mode left side bar and how to properly fix it.

PR #605, opened 2019-11-11
Updated Gitpod para

ShreyaPrasad1209:
Fixes  

Added a heads up for beginners who'll work in production environment


tachyons:
- It is social sign in, not normal sign in flow
- One can configure social sign in by setting env vars, See the required env vars here
- What we need is a wiki documentation for setting env vars for beginners

Issue #507, opened 2019-11-01
Bootstrap Missing in Post comment and cancel button


sakshi1499:
**Is your feature request related to a problem? Please describe.**
Bootstrap missing in post and cancel buttons in the add comment section.

**Describe the solution you'd like**
Bootstrap can be added to make it look better and in sync with other buttons.

**Additional context**
Screenshot attached.

![comment](https://user-images.githubusercontent.com/42520146/68033211-42352400-fce5-11e9-94bc-1c1a5433a013.PNG)

Arun-42:
@sakshi1499  Where in particular is this section?

PR #613, opened 2019-11-15
Add Bootstrap Styling to Buttons

thiefie:
Fixes Multiple Related Issues

Some of the buttons did not have the bootstrap styling
nowbootstrap


sakshi1499:
You have changed this for all the buttons, right? I think this changes the create assignment and post comment buttons to buttons with bootstrap. I thought the cdn was missing but you have added inline btn property right? @fluctix

fluctix:
@sakshi1499 Yup,
Btw Bootstrap cdn was already linked in application.html , so I just styled them.

sakshi1499:
@fluctix  I have a doubt in the html file we have added command form.submit right? So how are we actually defining the placeholder/ name of the button? I have seen this being used <%= f.submit 'A' %> for name 'A', in which file here is this name being given? How so?

fluctix:
https://stackoverflow.com/a/3027233
https://apidock.com/rails/ActionView/Helpers/FormBuilder/submit
Check these out,
 mainly it is a method that creates a form submit button, the stackoverflow answer illustrates how it works

sakshi1499:
I was referring to exactly these two documents. What my doubt is, is that while declaring form.submit in html file there is no mention of what the placeholder/ name is, so in which file has it been mentioned? These two links define a different style of naming the buttons namely by mentioning form.submit "PlaceholderName". @fluctix

fluctix:
"When no value is given, it checks if the object is a new resource or not to create the proper label"(apidock)
The label is added by rails automatically if no value is specified , but in case we want a custom name using `f.submit 'ChosenName'`

fluctix:
@PragatiVerma18 can you please review?

fluctix:
I think the syntax is correct can you point where the problem is?

PragatiVerma18:
`<%= f.submit 'name of button here', :class => 'submit_class_name_here' %>`
I was talking about this.

fluctix:
I found out class: button_class does the same thing . 
Reference: https://stackoverflow.com/a/16481375
Should I change it?

PragatiVerma18:
> I found out class: button_class does the same thing .
> Reference: https://stackoverflow.com/a/16481375
> Should I change it?

No need @fluctix. It's alright. Thanks.

fluctix:
Anything else that needs to be done?

PragatiVerma18:
#613 also needs similar help.

fluctix:
@PragatiVerma18 I have fixed it along with #507

tachyons:
Next time you create a new branch, give meaningful name

fluctix:
i would keep that in mind next time.

app/views/custom_mails/_form.html.erb
```diff
@@ -11,7 +11,7 @@
   </div>
 
   <div class="actions" onclick="before_submit()">
-    <%= form.submit() %>
+     <%= form.submit() nil, class: "btn btn-primary" %>
```
tachyons: remove `()`

Gemfile.lock
```diff
@@ -192,8 +199,10 @@ GEM
     nio4r (2.5.2)
     nokogiri (1.10.5)
       mini_portile2 (~> 2.4.0)
+
     oauth2 (1.4.2)
       faraday (>= 0.8, < 2.0)
+
```
tachyons: ?

Gemfile.lock
```diff
@@ -192,8 +199,10 @@ GEM
     nio4r (2.5.2)
     nokogiri (1.10.5)
       mini_portile2 (~> 2.4.0)
+
```
tachyons: ?

Issue #518, opened 2019-11-05
Unable to login using the admin credentials mentioned.


sakshi1499:
**Is your feature request related to a problem? Please describe.**
I am unable to login through the login credentials mentioned on the readme.md file of CircuitVerse.

**Describe the solution you'd like**
Make these credentials work or create a new account and hence update the readme.

**Additional context**
Screenshots have been attached.
![readme](https://user-images.githubusercontent.com/42520146/68181636-1e602f80-ffbe-11e9-9267-b165d6550e49.PNG)
![invalid](https://user-images.githubusercontent.com/42520146/68181711-52d3eb80-ffbe-11e9-8bb6-fdee89391ccb.PNG)

ankitkataria:
Doing `bundle exec rake db:seed` is necessary to seed the database with the admin credentials. It would be great add a note regarding this in the README.

sakshi1499:
Yes, okay.

sakshi1499:
Should I add a note for the same in readme? It would be easier for beginners.

tachyons:
@sakshi1499 go ahead

PR #614, opened 2019-11-15
Update Readme.md to enable login credentials

sakshi1499:
Fixes #

This PR fixes issue #518 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,9 @@ docker-compose build --no-cache
 ```
 
 ## Developer Instructions
-Developers can quickly get started by setting up the dev environment using the instructions above. The database is seeded with the following admin account.
+Developers can quickly get started by setting up the dev environment using the instructions above. It is necessary to execute 
+'bundle exec rake db:seed' to seed the database with the admin account.
+The Admin Account credentials are as follows:  
 ```
 User: Admin
 Email: admin@circuitverse.org
--- a/README.md
+++ b/README.md
@@ -65,9 +65,7 @@ docker-compose build --no-cache
 ```
 
 ## Developer Instructions
-Developers can quickly get started by setting up the dev environment using the instructions above. It is necessary to execute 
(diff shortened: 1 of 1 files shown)


sakshi1499:
Do you want me to add this in the readme itself? Or the link to the file or wiki? I feel adding the whole thing in the readme will make it too long unnecessarily. Maybe a link would seem cleaner. What do you suggest?

tachyons:
just mention run rake db:seed in development environment to populate sample data. No need to mention what all data are populated

sakshi1499:
Okay, will update it that way :)

sakshi1499:
I was not getting what you wrote. I think what you mean to convey is that this command rake db:seed not only seeds the admin account but also other things. So instead of explicitly mentioning admin account I should mention that it populates the sample data and loads it, right? Correct me if I am wrong

tachyons:
Yes

sakshi1499:
@tachyons check if it is better now?

README.md
```diff
@@ -65,7 +65,9 @@ docker-compose build --no-cache
 ```
 
 ## Developer Instructions
-Developers can quickly get started by setting up the dev environment using the instructions above. The database is seeded with the following admin account.
+Developers can quickly get started by setting up the dev environment using the instructions above. It is necessary to execute 
+'bundle exec rake db:seed' to populate the sample data and access admin account.
```
ankitkataria: It's not necessary to seed the db for starting development, how does the following look? 

> To seed the database with some sample data, run `bundle exec rake db:seed`. The admin credentials after seeding will be as follows ...

sakshi1499: It seems good. I'll make these changes :)

sakshi1499: I think you already made the changes. Right? @ankitkataria

sakshi1499:
@ankitkataria This has to be closed too right?

PR #645, opened 2019-11-23
Redesign for About page

aayush-05:
Updated the About Us page 

**Screenshots**
1.
image
2.
image


aayush-05:
@tachyons Where am I supposed to put the styling code, should I make a separate stylesheet for about page or just keep it in the logix one ?
Also, I think now I can just replace the code for the about page in the existing html file since the controller has been shifted already ?

tachyons:
Yes, erb file path is now changed, put images in app/assets/images and styles in app/assets/stylesheets/about.scss

PR #684, opened 2019-12-04
[GCI] Hide deleted comments

### LINKED ISSUE #508: Comment should delete after being deleted.
As a user, its not likely that I would want my deleted comment to be marked as "comment deleted".

It should actually be deleted or removed for better user experience.

**Additional context**
Screenshot Attached.
delete

junron:
Fixes #508 

Deleted comments are hidden.

However, I still need some help with the flash messages part.
ping @shubhankarsharma00 

[GCI task](https://codein.withgoogle.com/dashboard/task-instances/6373363631521792/)

### DIFF
--- a/app/views/commontator/comments/_show.html.erb
+++ b/app/views/commontator/comments/_show.html.erb
@@ -9,6 +9,7 @@
    avatar = Commontator.commontator_avatar(creator, self) || ''
 %>
 
+<% unless comment.is_deleted? %>
 <div id="comment_<%= comment.id.to_s %>_div" class="comment">
   <div id="comment_<%= comment.id.to_s %>_top_div" class="comment_div">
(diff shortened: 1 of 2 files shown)


shubhankarsharma00:
What I meant with a flash message was actually a js function to show an undo option for 10 seconds or so. I think this can be done using a setTimeout function (which can trigger events after a set time) over here:  
https://github.com/CircuitVerse/CircuitVerse/blob/535b3c269b003aaa523492563af7be156a053b4d/app/views/commontator/comments/_actions.html.erb#L22-L28
But I think the PR satisfies the issue at hand and the issue of whether there should be a flash message or a confirm before deleting can be dealt with later in a new issue. This seems good enough.

junron:
Thanks for the swift response! I'll submit my task.

Issue #505, opened 2019-11-01
Invalid emails being allowed in 'add collaborator'


sakshi1499:
**Is your feature request related to a problem? Please describe.**
I tried to add invalid email and it showed it was valid.
Email that I added was test@rosy
**Describe the solution you'd like**
Invalid emails should get removed.

**Additional context**
Screenshot attached.
![invalid](https://user-images.githubusercontent.com/42520146/68023657-c9769d80-fccd-11e9-9b14-e2c3a357ef1f.PNG)

Shivansh2407:
Can I please work on this issue?

Shivansh2407:
Can I work on this @tachyons and @sakshi1499 ?

sakshi1499:
As all the PRs have been closed without merging, you can work on this. Ref the last PR, some comments have been made for improvement. Include those. @Shivansh2407

Shivansh2407:
Ok Cool @sakshi1499

PR #687, opened 2019-12-04
Update devise.rb

chambin0:
Fixes issue #505 
I fixed the email validation line in config/initializers/devise.rb

Changed line 173 from "config.email_regexp = /\A[^@\s]+@[^@\s]+\z/" to "config.email_regexp = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/"

### DIFF
--- a/app/helpers/utils.rb
+++ b/app/helpers/utils.rb
@@ -5,7 +5,7 @@ module Utils
   # @return array of valid emails
   def self.parse_mails(mails)
     mails.split(/[\s,\,]/).select do |email|
-      email.present? && /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.match(email)
+      email.present? && Devise::email_regexp.match(email)
     end.uniq
   end
 
--- a/app/helpers/utils.rb
+++ b/app/helpers/utils.rb
@@ -5,7 +5,7 @@ module Utils
   # @return array of valid emails
   def self.parse_mails(mails)
     mails.split(/[\s,\,]/).select do |email|
-      email.present? && URI::MailTo::EMAIL_REGEXP.match(email)
+      email.present? && /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.match(email)
     end.uniq
   end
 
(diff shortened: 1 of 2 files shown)


tachyons:
https://github.com/CircuitVerse/CircuitVerse/blob/master/app/controllers/group_members_controller.rb#L35 This is the line you need to check

chambin0:
Okay, now i also changed the utils.rb which is actually responsible for the email validation. I think it should work correctly now.

app/helpers/utils.rb
```diff
@@ -5,7 +5,7 @@ module Utils
   # @return array of valid emails
   def self.parse_mails(mails)
     mails.split(/[\s,\,]/).select do |email|
-      email.present? && URI::MailTo::EMAIL_REGEXP.match(email)
+      email.present? && /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.match(email)
```
tachyons: You can use Devise::email_regexp instead

config/initializers/devise.rb
```diff
@@ -170,7 +170,7 @@
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
+  config.email_regexp = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/
```
tachyons: remove this change

chambin0: Why? it's a correct regular expression for email validation from the link given in the task description.

app/helpers/utils.rb
chambin0: I did it but the bot says i shouldn't use "::".

app/helpers/utils.rb
```diff
@@ -5,7 +5,7 @@ module Utils
   # @return array of valid emails
   def self.parse_mails(mails)
     mails.split(/[\s,\,]/).select do |email|
-      email.present? && URI::MailTo::EMAIL_REGEXP.match(email)
+      email.present? && Devise::email_regexp.match(email)
```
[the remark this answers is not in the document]
tachyons: Use . instead of :: and match? instead of match

tachyons:
Update PR title, squash commits and provide meaningful commit message

PR #689, opened 2019-12-04
Update features.html.erb

YashviShahh:
Fixes #

I have removed the break line, and so the padding has became proper as compared to other elements.

After doing the changes

Screenshot (17)


tachyons:
Hi, Welcome to Circuitverse. Can you please add before screen shot as well? Also do not keep the code commented, remove it instead if not needed anymore

PR #696, opened 2019-12-05
Added fadeout to messages

### LINKED ISSUE #594: Messages should fade after sometime or upon clicking anywhere on window.

The messages should fade after some time or upon clicking anywhere else on the window. This would be great for user experience on phone as well as on laptop view.

Set the fade property for the message div using jQuery or JavaScript accordingly so that it fades after a certain duration or upon clicking anywhere else on the window.

**Additional context**

error

No, this is for GCI 2019.

BBroadbent:
Fixes #594

### DIFF
--- a/app/views/logix/index.html.erb
+++ b/app/views/logix/index.html.erb
@@ -1,6 +1,3 @@
-<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
-<script src="js/scripts.js"></script>
-
 <script>
   $(document).ready(function(){
     $("#alertdiv").delay(5000).fadeOut(600);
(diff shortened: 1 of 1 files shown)


app/views/logix/index.html.erb
```diff
@@ -1,3 +1,12 @@
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
```
vishnu-v-v: Did you check if jQuery is already present in the application?

app/views/logix/index.html.erb
```diff
@@ -1,3 +1,12 @@
...
+<script src="js/scripts.js"></script>
+
+<script>
+  $(document).ready(function(){
+    $("#alertdiv").delay(10000).fadeOut(600);
```
ankitkataria: Great job @BBroadbent ! 
The delay before the fadeOut seems a little too much, could we reduce it to something like 5s?

app/views/logix/index.html.erb
vishnu-v-v: Did you check if jQuery is already included in the application. If it is already included, then you don't need to include it here. Please recheck this.

Everything else is good to go. 👏

Issue #707, opened 2019-12-07
No blank option for the user's country when editing


JoshHeng:
**Describe the bug**
There is no blank option for the country when an account is edited.
When a new user account is created and the user edits their profile, the country is set to India as it is the top of the list.

**To Reproduce**
1. Register a new account
2. Go to the profile page and click edit

**Expected behavior**
There should be a blank option for the country

**Are you working on this issue? (Yes/No)**
No

PR #725, opened 2019-12-10
Blank Country

NumberC:
Fixes #
Fixed Error #707


tachyons:
Already resolved in #819

PR #739, opened 2019-12-15
several bug fixes and modifications including reducing file size by r…

oneLoveUser:
several bug fixes of **_functions.js;_** and reduced the file size by removing unnecessary images and linked them from the web


junron:
You may want to run `$ git config --global core.autocrlf true` in the future if you are using Windows to avoid the large number of files changed. See [documentation](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings)

PR #741, opened 2019-12-16
Added erb view that lists custom emails

nrs1729:
#### Describe the changes you have made in this PR:
I added a view to list the custom emails created so far (the view is only visible to admins). 

### Screenshots of the changes (If any):
image
image

### DIFF
--- a/app/views/custom_mails/index.html.erb
+++ b/app/views/custom_mails/index.html.erb
@@ -1,0 +1,30 @@
+</div>
+<div class="container">
+    <div class="row feature">
+        <div class="col">
+            <% if @mails.count!=0 %>
+                <h1 style="margin-bottom:15px" >Custom emails: </h1>
+                <table class="table">
+                    <thead>
+                        <tr>
+                            <th>Subject</th>
+                            <th>Sent?</th>
+                            <th colspan="2"></th>
+                        </tr>
+                    </thead>
+
+                <tbody>
+            <% @mails.each do |mail| %>
+                <tr>
+                  <td><%= mail.subject %></td>
+                  <td><%= mail.sent ? "Yes" : "No" %></td>
+                  <td><%= link_to 'Show', mail %></td>
+                  <td><%= link_to 'Edit', edit_custom_mail_url(mail) %></td>
+                </tr>
+            <% end %>
+            </tbody>
+          </table>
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -14,7 +14,7 @@
     resources :assignments
   end
 
-  resources :custom_mails, only: [:new, :create, :edit, :show, :update]
+  resources :custom_mails, only: [:index, :new, :create, :edit, :show, :update]
```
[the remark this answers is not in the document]
ankitkataria: Can we fix this error?

app/views/custom_mails/index.html.erb
```diff
@@ -0,0 +1,30 @@
...
+                <tr>
+                  <td><%= mail.subject %></td>
+                  <td><%= mail.sent ? "Yes" : "No" %></td>
+                  <td><%= link_to 'Show', mail %></td>
+                  <td><%= link_to 'Edit', edit_custom_mail_url(mail) %></td>
+                </tr>
+            <% end %>
+            </tbody>
+          </table>
+        <% end %>
```
ankitkataria: What if the mail count is 0? Could we have "No custom mails so far" is that case?

ankitkataria:
Also rebase this PR to latest master, only the commits relevant to the issue should be present in the branch.

nrs1729:
Sorry about the commits, fixed it. I added a case for if there are no emails and a test that checks there's a response and that it contains the mail subject, is that okay? Not sure if there's something else to cover for the coverage/coveralls check. The string literals violations hound is bringing up seem in line with the rest of the code?

spec/controllers/custom_mail_controller_spec.rb
```diff
@@ -68,6 +68,14 @@
         expect(response.body).to eq("The mails were queued for sending!")
       end
     end
+
+    describe "#index" do
+      it "shows the list of custom views" do
+        get custom_mail_path(@mail)
```
ankitkataria: You'll have to do `get custom_mails_path` to actually check the index route, this should check for the index route. 
This is also the reason why the test coverage is failing.

nrs1729: Thanks, should be fixed now

PR #744, opened 2019-12-18
Added Video Tutorial for Docker.

namantechlabs:


tachyons:
This can we added in wiki instead

PR #750, opened 2019-12-20
README System Tests Clarification

MrBartusek:
image
New system tests fail when chrome browser is not installed, I think adding a clarification will be helpfull


tachyons:
Thanks

Issue #369, opened 2019-04-15
Adding features for Test Bench


igaurab:
**Is your feature request related to a problem? Please describe.**
There is no feature for deleting particular input or output columns in [testbench](https://circuitverse.org/testbench)`This becomes a matter of issue when we accidentally click on add inputs or add outputs. If one have written a significant amount of test cases, in such case currently there is no option but to refresh the entire page and start fresh.
**Describe the solution you'd like**
I would like to add a button "Delete Column". The user first selects the column they want to delete and click on delete column to delete the column. Or may be other way around(Please suggest)
**Additional context**
I will also work on to improve the design of the page. 

**Are you working on this? (Yes/No)**
Yes

nrs1729:
Would it be okay to have a "-" button on each column that deletes it instead of one button at the top?

PR #754, opened 2019-12-21
Added ability to delete rows and columns

nrs1729:
#### Describe the changes you have made in this PR: 
Added buttons to delete input and output columns, and cases. 

### Screenshots of the changes (If any):
image

Issue #369

### DIFF
--- a/public/testbench.html
+++ b/public/testbench.html
@@ -63,6 +63,15 @@
     .inp{
         bgcolor:"red";
     }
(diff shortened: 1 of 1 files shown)


nrs1729:
I noticed there was a preexisting issue when the user set two column names to be the same (nothing to stop this happening) but its not really connected to this issue so I'll create a separate one.

PR #783, opened 2019-12-28
Added social media icons

### LINKED ISSUE #782: Add social media icons
We've just added Facebook and Twitter accounts. Icons/links to these should be shown on the header

Yes

JoshHeng:
Fixes #782 
Note that if we obtain a different Facebook username, the URL will have to be changed

Added Facebook and Twitter icons


nitin10s:
Could you post screenshot of a mobile screen as well. Thanks.

nitin10s:
A little padding on both sides vertically would make it better.

JoshHeng:
@nitin10s Vertical padding would make the top bar bigger vertically on a wider screen or shrink the social icons.
Media queries would be possible but would get messy if anything else in the top bar was changed

nitin10s:
I understand but the current design looks a little cluttered in the mobile view. Maybe changing padding in % would help.

JoshHeng:
I've changed the links to the new redirect routes (#796 #787) so we only need to update links in one place if they do change, and because the Facebook link needed to be updated anyway

nitin10s:
@tachyons

JoshHeng:
Note that the merged commit actually requires PR #796 to be merged as well as it redirects to links such as [circuitverse.org/facebook](circuitverse.org/facebook) which are only added in that PR.

Otherwise if that PR is rejected I can change the links back to the direct ones to Facebook and Twitter

MrBartusek:
@tachyons @JoshHeng new icons redirect to `https://circuitverse.org/facebook` instead of `/facebook` so when running locally they redirect to online page which throws 404

JoshHeng:
> new icons redirect to `https://circuitverse.org/facebook` instead of `/facebook` so when running locally they redirect to online page which throws 404

I think this is probably fine as once the `master` branch is deployed it will be a static link and won't change often

tachyons:
Oh wait, those shouldn't be hardcoded. Because this is an open source application and people can host their own instance of Circuitverse, domain name shouldn't be hardcoded in the codebase

PR #793, opened 2020-01-01
Readme installation

yummypasta:
## Describe the changes you have made in this PR -
Gem listener and the SCSS bootstrap package would not be installed if someone followed only the old instructions.

I made the bundle install command also include test dependencies, and added an npm install step.


MrBartusek:
we are using `yarn` which is listed in the readme, can be closed

PR #800, opened 2020-01-02
Delete Offline Projects Fix

### LINKED ISSUE #338: Deleting Offline Project
**Problem**
There is no way to delete an offline project either the user has to clear the complete local storage which will delete all his offline projects.

Adding Delete Button on the Project > Open Oflline.

open offline Project it will display a cross button on the right of the title.
when we click the on the button it will ask for confirmation from user. if the user confirms it will delete the project.

**Prototype**
look at the cross button.
proto

Yes

MrBartusek:
Fixes #338 
There is an old feature request that has been implemented but, the issue wasn't closed and it is not working now. I made it look better and work + this pr will finally close this issue

image

### DIFF
--- a/public/css/UX.css
+++ b/public/css/UX.css
@@ -1,5 +1,4 @@
-.deleteOff {
-    color: red;
+.deleteOfflineProject {
     float: right;
     cursor: pointer;
     padding: 2px;
--- a/public/js/data.js
+++ b/public/js/data.js
@@ -777,7 +777,7 @@ deleteOfflineProject = function (id) {
         localStorage.setItem("projectList", JSON.stringify(projectList));
         $('#openProjectDialog').empty();
         for (idd in projectList) {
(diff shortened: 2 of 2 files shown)


public/js/data.js
```diff
@@ -777,7 +777,7 @@ deleteOfflineProject = function (id) {
         localStorage.setItem("projectList", JSON.stringify(projectList));
         $('#openProjectDialog').empty();
         for (idd in projectList) {
```
JoshHeng: Use `id` instead of `idd`

public/css/UX.css
```diff
@@ -1,5 +1,4 @@
-.deleteOff {
-    color: red;
+.deleteOfflineProject {
```
JoshHeng: Maybe move to .scss files in `app/assets/stylesheets`

MrBartusek: that can be done when stylesheets of a simulator will be cleaned not a task for this pr I think

satu0king:
Good Job

PR #814, opened 2020-01-08
[GCI] Add group management system tests

junron:
Added system tests for group management system
Tests creating group, and editing group name, adding members and removing members!

[GCI task](https://codein.withgoogle.com/dashboard/task-instances/6398169206751232/)

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -67,6 +67,8 @@ RSpec.configure do |config|
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  config.include Warden::Test::Helpers
 end
 
 Shoulda::Matchers.configure do |config|
--- a/spec/system/group_spec.rb
+++ b/spec/system/group_spec.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Group management", type: :system do
+  before(:all) do
+    @user = FactoryBot.create(:user)
+    @user2 = FactoryBot.create(:user)
+    @group = FactoryBot.create(:group, mentor: @user)
+  end
+
+  before(:each) do
+    driven_by(:selenium)
+    login_as(@user, scope: :user)
+  end
+
+  after(:each) do
+    Warden.test_reset!
+  end
+
+  it "should create a group" do
+    visit "/groups/new"
+    fill_in "group[name]", with: "Test"
+    click_button "Create Group"
+
+    expect(page).to have_text("Group was successfully created.")
+  end
+
+
+  it "should add a member to the group" do
+    visit "/groups/#{@group.id}"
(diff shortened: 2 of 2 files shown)


junron:
The error that fails the test seems to be unrelated. Also the tests worm on circle ci for some reason.
Please have a look @tachyons

tachyons:
App veyor is windows CI, so environment can vary, we need to get this work on both platforms

spec/rails_helper.rb
```diff
@@ -66,6 +66,8 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  config.include Warden::Test::Helpers
```
tachyons: https://github.com/CircuitVerse/CircuitVerse/blob/bfe6a5884b65eca5b7de58ddd328dd39d4a13a73/spec/rails_helper.rb#L43 

Adding system here should work

junron: hmm it still doesn't work

junron: @tachyons

junron:
@tachyons is it ok if I just do the adding group test? the adding member part seems problematic due to the morals.

junron:
@tachyons Seems like it's more complicated than I thought, could the task be extended til Sunday?

spec/system/group_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Group management", type: :system, js: true do
+  before do
+    Capybara.raise_server_errors = false
```
Ananthavijay: Is using this fine? @junron
Instead, just move the ``time.js`` to ``public/js`` directory. That will solve the issue I guess.

junron: Great suggestion! I'll try it out

junron:
@tachyons @Ananthavijay Copying the file solved the issue, but it also caused a lot of duplication. Maybe another PR could solve the underlying issue of the file not being served properly, then I'll update my PR accordingly.

tachyons:
Only group creation is included in this PR so far, Need group managing group feature as well, like editing, deleting, adding members etc

junron:
@tachyons @ankitkataria All done!

spec/system/group_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    click_on "Remove"
+
+    accept_alert
+
+    expect(page).to have_text("Group member was successfully destroyed.")
+  end
+
+  it "should change the group name" do
+    visit "/groups/#{@group.id}"
+
```
alanjsph: Is there a particular reason why you are leaving spaces between each live ?

junron: Not really, except to separate actions from assertions. I'll remove all other unnecessary line breaks

junron:
@alanjsph Please have a look

Issue #398, opened 2019-07-02
Improve Assignment Restricted Element Reporting


satu0king:
1) Show restricted elements banner only if present
![image](https://user-images.githubusercontent.com/2092958/60487354-55939c80-9cbd-11e9-9a4c-ec89fa97e131.png)
2) Show restricted elements across all scopes in assignment preview (when mentor sees)
3) Improve restricted element detection
4) Show warning message when circuit is saved

ankitkataria:
I'll start working into this. 😄

satu0king:
Oh, I was going to assign this to myself xD

ankitkataria:
We can divide work on this one too. Which ones would you want to pick?

shubhankarsharma00:
Leaving some added information:
- [ ] To show Restricted Elements div only when applicable:
Just add an if statement  
`         <%if @assignment.restricted_circuit_elements != "None"%>
`
https://github.com/CircuitVerse/CircuitVerse/blob/2898a70401ae1d7521a9e8c51347ec9710a74bdf/app/views/assignments/show.html.erb#L38-L40

PR #818, opened 2020-01-10
Shows restricted elements if present

Ananthavijay:
Fixes https://github.com/CircuitVerse/CircuitVerse/issues/398

- Added an if condition.

### DIFF
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -30,10 +30,12 @@
         <%= @assignment.graded %>
       </p>
 
+     <% if @assignment.restricted_circuit_elements != "None" %>
       <p>
         <strong>Description:</strong>
         <p><%= @assignment.description&.html_safe%></p><br>
       </p>
+    <% end %>
 
       <p>
         <strong>Restricted Circuit Elements:</strong>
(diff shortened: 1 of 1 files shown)


app/views/assignments/show.html.erb
```diff
@@ -30,10 +30,12 @@
...
 
+     <% if @assignment.restricted_circuit_elements != "None" %>
       <p>
         <strong>Description:</strong>
         <p><%= @assignment.description&.html_safe%></p><br>
       </p>
+    <% end %>
 
       <p>
         <strong>Restricted Circuit Elements:</strong>
```
fluctix: Shouldnt this be within the if condition?

fluctix: It seems you have added the condition to the description...

MrBartusek: I haven't seen that too, good point

fluctix:
That's the reason appveyor has gone red

Ananthavijay:
My bad, didn't see it properly

MrBartusek:
@fluctix yes it is `Failure/Error: expect(response.body).to include(@assignment.description)`

Ananthavijay:
Pushed it in a hurry, thanks for spotting it out @fluctix

fluctix:
Looks good..

PR #819, opened 2020-01-11
Better Country Picker on User's Profile

### LINKED ISSUE #707: No blank option for the user's country when editing

MrBartusek:
Fixes #707

- Added option to do not select a country
- Added user's Geo-IP Country on top of the dropdown (priority country)

When user's country can be detected (via IP):
cv_server
When cannot bedetected  :
cv_local

### DIFF
--- a/app/views/users/logix/edit.html.erb
+++ b/app/views/users/logix/edit.html.erb
@@ -14,7 +14,7 @@
 
         <div class="field form-group">
           <%= f.label :country %><br/>
-          <%= f.country_select :country, { priority_countries: ["IN"] }, { class: "form-control" } %>
+          <%= f.country_select :country, { include_blank: 'Prefer Not To Enter', priority_countries: get_priority_countries() }, { class: "form-control" } %>
         </div>
 
         <div class="field form-group ">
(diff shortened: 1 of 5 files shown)


app/views/users/logix/edit.html.erb
```diff
@@ -14,7 +14,7 @@
 
         <div class="field form-group">
           <%= f.label :country %><br/>
-          <%= f.country_select :country, { priority_countries: ["IN"] }, { class: "form-control" } %>
+          <%= f.country_select :country, { include_blank: 'Prefer Not To Enter', priority_countries: get_priority_countries() }, { class: "form-control" } %>
```
JoshHeng: I would change `Prefer Not To Enter` to `-`

MrBartusek: Moved to [Slack](https://circuitverse-team.slack.com/archives/CQ17325A5/p1579456597001500)

PR #822, opened 2020-01-15
Delete package-lock.json

MrBartusek:
We recently switched to yarn but didn't deleted `package-lock.json` and that is causing some problems (Heroku for example) [Slack Conversation](https://circuitverse-team.slack.com/archives/CQFAC5L7P/p1579036024001300)


fluctix:
Should be merged ASAP , as having both of them would confused new users also on whether to use yarn or npm , also yarn does not detect npm dependencies so there could be a lot of problems. Maybe we could have a note in the readme to ask users to use yarn only?

PR #832, opened 2020-01-20
Fix navigation links in header

### LINKED ISSUE #765: Navbar not working in new ux
**Describe the bug**
On any page navbar buttons have `#` in href, `Learn` button is always active
Only `Login` and `Notifications` works properly
Page logo should work as an index button but it not

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to any page with navbar
2. See error

**Expected behaviour**
Clicking buttons should change page and the selected page should be marked as active

**Screenshots**
image

 - OS: Windows 10
 - Browser: Vivaldi (chromium)

[Beginner GCI Task](codein.withgoogle.com/tasks/5710665268854784)

JoshHeng:
Navigation links in the header have been fixed to anchor to the correct places, instead of to #

Fixes #765


app/views/layouts/_header.html.erb
```diff
@@ -10,25 +10,25 @@
   <div class="collapse navbar-collapse" id="collapsedNavbar">
     <ul class="navbar-nav ml-auto mt-2 mt-lg-0">
       <li class="nav-item px-2">
-        <button class="btn btn-outline-success my-2 my-sm-0 rounded-0 simulator text-dark" type="submit">Simulator</button>
+        <a class="btn btn-outline-success my-2 my-sm-0 rounded-0 simulator text-dark" href="/simulator">Simulator</button>
       </li>
-      <li class="nav-item px-2 active">
```
tachyons: Can we also make sure that active class applied properly based on the current page

JoshHeng: This should be done

aayush-05: I guess @tachyons meant to add the active class to an element only when that particular page is active; rather than removing it altogether.

PR #865, opened 2020-02-05
Adding ColorSelector [WIP]

### LINKED ISSUE #850: Color property selector
image

Colour text field at simulator should be replaced with the colour selector
_This issue is a part of my [Usability Test](https://drive.google.com/open?id=1igwLx2tc1dAU_75qY2F4u7ZnHf2crcp-)_

No

Shivansh2407:
Fixes #850 Color property selector

Here I have added a Color Selector Instead Of the text field . I guess we Do not require the colortorgba function anymore as this selector provides a Hexadecimal value automatically and hence I have commented them ( Will remove them once approved).

I will provide the screenshots soon

### DIFF
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -2296,8 +2296,8 @@ function DigitalLed(x, y, scope = globalScope, color = "Red") {
     this.directionFixed = true;
     this.fixedBitWidth = true;
     this.color = color;
-    var temp = colorToRGBA(this.color)
-    this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
+    //var temp = colorToRGBA(this.color)
+    //this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
 
 
 }
@@ -2316,15 +2316,15 @@ DigitalLed.prototype.customSave = function () {
 DigitalLed.prototype.mutableProperties = {
     "color": {
         name: "Color: ",
-        type: "text",
+        type: "color",
         func: "changeColor",
     },
 }
 DigitalLed.prototype.changeColor = function (value) {
     if (validColor(value)) {
         this.color = value;
-        var temp = colorToRGBA(this.color)
-        this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
+        //var temp = colorToRGBA(this.color)
+        //this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
     }
 
 }


thiefie:
The color option does not even appear now.
Does this work for you?
Screenshot_2020-02-05_17-29-12

Ananthavijay:
Change the title of PR too

Shivansh2407:
Working On It...Will push some changes soon @thiefie

MrBartusek:
@Shivansh2407 you can just save this pr as a draft or don't make pr at all if you are still working on it

nitin10s:
@Shivansh2407 Are you working on this?

Shivansh2407:
Yes @nitin10s I am working on this. I will open a PR soon

nitin10s:
@Shivansh2407 We should close this one then, right?

Shivansh2407:
I will be adding and squashing the commits in this patch only so I think we should let it be open.

MrBartusek:
@Shivansh2407 any update? are you still working on it?

Shivansh2407:
Yes @MrBartusek I am working on it. Some changes needs to be fixed and then I will update the PR.

public/js/modules.js
```diff
@@ -2296,8 +2296,8 @@ function DigitalLed(x, y, scope = globalScope, color = "Red") {
     this.directionFixed = true;
     this.fixedBitWidth = true;
     this.color = color;
-    var temp = colorToRGBA(this.color)
-    this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
+    //var temp = colorToRGBA(this.color)
+    //this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
```
satu0king: I don't think this change is necessary

public/js/modules.js
```diff
@@ -2316,15 +2316,15 @@ DigitalLed.prototype.customSave = function () {
...
         func: "changeColor",
     },
 }
 DigitalLed.prototype.changeColor = function (value) {
     if (validColor(value)) {
         this.color = value;
-        var temp = colorToRGBA(this.color)
-        this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
+        //var temp = colorToRGBA(this.color)
+        //this.actualColor = "rgba(" + temp[0] + "," + temp[1] + "," + temp[2] + "," + 0.8 + ")";
```
satu0king: I don't think this change is necessary

public/js/modules.js
```diff
@@ -2316,15 +2316,15 @@ DigitalLed.prototype.customSave = function () {
 DigitalLed.prototype.mutableProperties = {
     "color": {
         name: "Color: ",
-        type: "text",
+        type: "color",
```
satu0king: This change is necessary to make change to the UI

satu0king:
Note that, the color picker should pick color, convert to hex and send inside to the function. Therefore this should only be a UI change. I actually prefer having the current option of writing the color name/color hex code as its easier to say I want "red", "green", "blue" rather than picking it from a color picker. Of course some users might need to have the color chosen from a picker.

Shivansh2407:
Ok @satu0king. So then what needs to be done here?

satu0king:
> Ok @satu0king. So then what needs to be done here?

Sorry for delayed response. One way is to give text option just like how it is now and give a color picker version as well. I have seen some applications do this well not able to find a good example now.

abstrekt:
You're working on it right? @Shivansh2407

satu0king:
@Shivansh2407 @abstrekt shall we work on this now?

abstrekt:
> 
> 
> @Shivansh2407 @abstrekt can we work on this now?

what's the progress?

satu0king:
> > @Shivansh2407 @abstrekt can we work on this now?
> 
> what's the progress?

I think we need to decide how the UI will look. As previously discussed in this thread. I think having both color picker + text option is good. Because, it is easier to write "green", "red", "magenta" than picking from a color picker. (I personally find that harder). Also it is important to note that users won't be very picky on their colors as the colors are meant to differentiate rather than aesthetics. The rendering code itself uses any valid CSS color.

@nitin10s any ideas?

satu0king:
Closing this issue as its directed to old UI. Thanks!

PR #882, opened 2020-02-08
URL for User Projects and Profile Details

### LINKED ISSUE #713: Change the URLs for the profile details and a user projects
It would make more sense if the profile information of a user was found at https://circuitverse.org/users/00000/ and the projects found at https://circuitverse.org/users/00000/projects

Currently the projects are found at https://circuitverse.org/users/00000/ and profile at https://circuitverse.org/users/00000/profile

No

Shivansh2407:
Fixes #713 

Here I have Fixed the routes.rb files to change the Url for Project from https://circuitverse.org/users/id/ to https://circuitverse.org/users/id/projects and for details from https://circuitverse.org/users/id/profile to https://circuitverse.org/users/id/  . Similar changes take place for Editing profiles also.

BEFORE :
before1
before2

AFTER :
after1
after2

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -62,11 +62,11 @@ Rails.application.routes.draw do
 
   scope '/users' do
     get '/', to: 'welcome#all_user_index', as: 'all_users'
-    get '/:id/profile', to: 'users/logix#profile', as: 'profile'
-    get '/:id/profile/edit', to: 'users/logix#edit', as: 'profile_edit'
+    get '/:id/', to: 'users/logix#profile', as: 'profile'
+    get '/:id/edit', to: 'users/logix#edit', as: 'profile_edit'
     patch '/:id/update', to: 'users/logix#update', as: 'profile_update'
     get '/:id/groups', to: 'users/logix#groups', as: 'user_groups'
-    get '/:id/', to: 'users/logix#index', as: 'user_projects'
+    get '/:id/projects', to: 'users/logix#index', as: 'user_projects'
     get '/:id/favourites', to: 'users/logix#favourites', as: 'user_favourites'
     get '/educational_institute/typeahead/:query' => 'users/logix#typeahead_educational_institute'
     get '/:id/notifications', to: 'users/notifications#index', as: 'notifications'


Shivansh2407:
I think that the Profile Page should have a Link to the User Projects too Otherwise people will not be able to see the projects of others. Please give your views on this @Ananthavijay and @JoshHeng

tachyons:
This will break existing URLs people shared, any changes like this should back with proper reason and fallback machanism to keep previously shared URLs work

Shivansh2407:
> This will break existing URLs people shared, any changes like this should back with proper reason and fallback machanism to keep previously shared URLs work

So @tachyons what should be done here ?

Shivansh2407:
@tachyons what should we do here ? Shall we continue with the existing URL or this new approach ?

thiefie:
The problem would be with `/id/profile` as it does not exist anymore. 
You could have a 
1) redirect mechanism or
2) maybe display the same content as `/id` and `/id/profile`

@tachyons @Shivansh2407 what do you think?

Shivansh2407:
> The problem would be with `/id/profile` as it does not exist anymore.
> You could have a
> 
> 1. redirect mechanism or
> 2. maybe display the same content as `/id` and `/id/profile`
> 
> @tachyons @Shivansh2407 what do you think?

Yeah we have either of the methods preferably the second one. I would like to hear your views too on this @tachyons

tachyons:
Just realised that you are changing the content of the link as well, it will break lot of links which are shared in blogposts, assignments, documents etc. We should never change the functionality of URL once the application is live

Shivansh2407:
I will revert those changes @tachyons Sir

Shivansh2407:
Should I proceed with making the content redirected to that link or content on both links same?

thiefie:
Based on what @tachyons said , i think we should stick to how it was.

tachyons:
Yes

tachyons:
Thanks for your attempt

Shivansh2407:
> Thanks for your attempt

Glad to Contribute Sir 👍

PR #890, opened 2020-02-09
Bit width issue

### LINKED ISSUE #855: BidWidth Error Changes
**Describe the bug**
BidWidth error is hard to understand for beginners and its behaviour should be changed. Currently, the cable will light up red and it shows an error message (as on screenshot)

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Simulator
2. Insert `Input` with `BidWidth: 2` and `Digital Diode` `BidWidth: 1`
3. As excepted it produces an error

**Expected behaviour**
Better feedback will be to light up the target node to red when the user will be still dragging it and when the user tries to attach it (drop) the connect won't create and an error prompt (not that error like on screenshot) will show

**Screenshots**
image

No

jha11aditya:
Fixes #855

#### Describe the changes you have made in this pr - In Node.js I wrote to diisplay the obj type of the two elements between which the BitWidth error occurs . if the connection to the second element is made indirectly from a wire node then it displays string "Wire" else the name of the first object.
Next i wrote to delete the last wire added.

### DIFF
--- a/public/js/Node.js
+++ b/public/js/Node.js
@@ -212,6 +212,14 @@ Node.prototype.connect = function(n) {
     updateSimulation = true;
     scheduleUpdate();
 }
+Node.prototype.disconnectWire = function(n) {
+    //searches for the wire which connects this-obj and node-obj and deletes it
+    for (var i = this.parent.scope.wires.length - 1; i >= 0; i--){
+        if((this.parent.scope.wires[i].node1 === this) && (this.parent.scope.wires[i].node2 === n)){
+            this.parent.scope.wires[i].delete();
+        }
+    }
+}
 Node.prototype.connectWireLess = function(n) {
 
     if (n == this) return;
@@ -304,7 +312,15 @@ Node.prototype.resolve = function() {
             } else {
                 this.highlighted = true;
                 node.highlighted = true;
-                showError("BitWidth Error: " + this.bitWidth + " and " + node.bitWidth);
+                // if direct conn then show this.parent's obj type else it is a wire, show node's parent objtye ; And delete the latest wire connection made
+                obt = null;
+                if(this.parent.objectType == "CircuitElement"){
+                    obt = "Wire";
+                } else {
+                    obt = this.parent.objectType;
+                }
+                showError("BitWidth Error: " + obt + ": " + this.bitWidth + " and " + node.parent.objectType + ": " + node.bitWidth);
+                this.disconnectWire(node);
             }
         }
     }
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -125,6 +125,7 @@ GEM
       thor (>= 0.19.4, < 2.0)
       tins (~> 1.6)
     crass (1.0.5)
+    daemons (1.3.1)
     devise (4.7.1)
       bcrypt (~> 3.0)
       orm_adapter (~> 0.1)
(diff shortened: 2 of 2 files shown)


public/js/Node.js
```diff
@@ -212,6 +212,14 @@ Node.prototype.connect = function(n) {
...
     scheduleUpdate();
 }
+Node.prototype.disconnectWire = function(n) { 
+    //searches for the wire which connects this-obj and node-obj and deletes it
+    for (var i = 0; i < this.parent.scope.wires.length; i++){
+    if((this.parent.scope.wires[i].node1 === this) && (this.parent.scope.wires[i].node2 === n )){
+        this.parent.scope.wires[i].delete();
+        }
+    }
+}
```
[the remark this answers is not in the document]
jha11aditya: @tachyons In the previous commit i was only deleting the last pushed wire which could result in a fault, take the case --
    i connect two elements with same bit width and later i connect many more other elements ,
    Now if i change the bitwidth of the first element , the last pushed wire would be deleted . 
So I have instead added a disconnectWire(node)  function which searches for connections between the 2 objects which have BitWidth error and then deletes them. This provides the appropriate job.

jha11aditya:
This above push is because previously the branch had earlier commits form a different issue as I had committed in the master earlier. Now this pull request is clean and has commits only for this issue.

jha11aditya:
@MrBartusek 
1> If the wire is still being dragged how will i know which element is the target.
And yes in the case the mouse is above the target node in the dragging phase , the specific wire is being dropped after you attach it.
2> Why should the target node be lit red, it is not possible to determine whether the target or source node's bitwidth is at fault when seen as a whole. So the wire is lit RED and dropped thereafter.
3> The error prompt is necessary otherwise how will the user know what was the error he did.

MrBartusek:
@jha11aditya So if I understand correctly the current state of simulator code is not allowing to make these things work?

satu0king:
Good attempt! It partially solves the issue, though I am not very sure if this is the way to go. Pending for now.

satu0king:
@mahmodHammad can u take a look on this PR ?

public/js/Node.js
```diff
@@ -212,6 +212,14 @@ Node.prototype.connect = function(n) {
     updateSimulation = true;
     scheduleUpdate();
 }
+Node.prototype.disconnectWire = function(n) {
+    //searches for the wire which connects this-obj and node-obj and deletes it
+    for (var i = this.parent.scope.wires.length - 1; i >= 0; i--){
```
mahmodHammad, line 217: for better readability, please create a variable for this.parent.scope.wires

public/js/Node.js
```diff
@@ -212,6 +212,14 @@ Node.prototype.connect = function(n) {
     updateSimulation = true;
     scheduleUpdate();
 }
+Node.prototype.disconnectWire = function(n) {
```
mahmodHammad, line 215: This should work 👍

mahmodHammad:
This  PR it's just for displaying clear info about the Bit width issue, along with deleting the last connected wire.


so this can get merged @satu0king

satu0king:
Closing this issue because of the following reasons:
1. New UI is in a different codebase
2. Wiring needs to be completely redone as a separate task

PR #911, opened 2020-02-18
max bitwidth  cross fix

### LINKED ISSUE #910: Input BitWidth on increasing over 32 does not work .
**Describe the bug**
The input when below 32 we can easily toggle the Bits from 0 to 1 or vice-versa. But over 32, let us assume 33 then the 33rd bit does not allows toggle , In this case you only can toggle the first 32 bits and rest remain static. If we do not want the circuit to accept input more than 32 bits then we should provide any alert otherwise we should fix this.
The spinner up button does not allows going beyond 32 but when user enters manually then this bug occurs.

**To Reproduce**
1. Go to 'circuitverse.org'
2. Click on 'simulator'
3. Scroll down to 'inputs' and select input , then enter a bitwidth greater than 32 manually.

**Expected behavior**
It should not create the input box for bitwidth over 32 as it does not responds to it.

**Screenshots**
Issue3

 - OS: [Windows 10]
 - Browser [Chrome]

No

jha11aditya:
Fixes #910

#### Describe the changes you have made in this pr - I added a conditional block using Jquery that 
compares the value of netbitwidth to the defined max bitwith :
if greater Then makes the netbitwidth value max[i.e 32] & shows error
and if less/eq then does nothing

mxbw

### DIFF
--- a/public/js/UX.js
+++ b/public/js/UX.js
@@ -227,8 +227,8 @@ function showProperties(obj) {
     $(".objectPropertyAttribute").on("change keyup paste click", function () {
         // return;
         //////console.log(this.name+":"+this.value);
-        if(parseInt($('input[name=newBitWidth]').val()) > parseInt($('input[name=newBitWidth]').attr("max"))){
-            $('input[name=newBitWidth]').val(parseInt($('input[name=newBitWidth]').attr("max")));
+        if (parseInt($('input[name=newBitWidth]').val(), 10) > parseInt($('input[name=newBitWidth]').attr("max"), 10)) {
+            $('input[name=newBitWidth]').val(parseInt($('input[name=newBitWidth]').attr("max"), 10));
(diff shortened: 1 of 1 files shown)


Shivansh2407:
@jha11aditya please add a video/gif that describes the changes that you have done.

jha11aditya:
Is that error related to my modification[i.e addition of the condition block inside Jquery function]?

Shivansh2407:
> Is that error related to my modification[i.e addition of the condition block inside Jquery function]?

I have gone through my PRs Gitpod but they are working absolutely fine . I think so it might be some problem here. Please once ask it on slack channel 👍

satu0king:
I merged the PR as it looked right to me. However, I realized `input[name=newBitWidth]` need not always be present in all property menus. Please add a check to see if its present and then check the condition. @jha11aditya

Issue #916, opened 2020-02-20
Conflicting Integrations causing hinderance in development workflow


YashKumarVerma:
**Describe the bug**
As of now, multiple services are being used for code review, namely Hound, CodeCliamte, AccessLint.  These serivices were added to serve for segments of the application, i.e. the rails part, the frontend javascript engine etc, but are not configured properly, and therefore started reviewing all code changes.

The problem at hand is that these services have different configurations and hence contradict each other. This fails the auto PR review a significant number of times. Following the conventions of one service breaks the rules of the other.

**To Reproduce**
1. Edit some code for javascript
2. Make a pull request
3. OR refer #906 

**Expected behavior**
Only one service should be used for auto-code analysis.

**Screenshots**
Code Climate says
![image](https://user-images.githubusercontent.com/14032427/74906079-f2828180-53d5-11ea-8e68-4f1cfca44eff.png)
Hound says
![image](https://user-images.githubusercontent.com/14032427/74906096-ff9f7080-53d5-11ea-9943-6e608e148854.png)

**Desktop (please complete the following information):**
 - OS: [Ubuntu 19.10]
 - Browser [80.0.3987.106 (Official Build) (64-bit)]

**Are you working on this issue? (Yes/No)**
No.

tachyons:
Solution

Configure hound to not check any files other than erb

Shivansh2407:
@tachyons in the hound.yml file , it is only configured for ` erblint:  enabled: true` and others as false so then where should we look to fix this ?

tachyons:
get other engines which are tru by default and disable this

PR #917, opened 2020-02-20
Configure hound to check only erb files

Ananthavijay:
Fixes https://github.com/CircuitVerse/CircuitVerse/issues/916

- Configured hound to check only erb files.


tachyons:
Fixed in master

Issue #905, opened 2020-02-16
Code Refactor : Javascript Standards, Linting


YashKumarVerma:
**Is your feature request related to a problem? Please describe.**
- The current collection of frontend scripts is a beautiful piece of software, however, it does not follow the latest conventions and standards.
- The codebase does not follow any particular set of styles as well, which makes it look inconsistent and cluttered.   
- Linting throws nearly hundreds of errors in one file.
- Total 45k+ errors in `public/js` alone

**Describe the solution you'd like**
Completely acknowledging the fact that a lot of hard work and effort was put in to build the application logic, the primary focus was to get the simulator working. Now as the community has grown, we must enhance the code to make it more maintainable and well documented.

The javascript sources need to be refactored, to adapt any conventional styling guide, and must pass linting tests. 

**Describe alternatives you've considered**
- The current implementation breaks code into multiple files then creates a function for each item.
- A cleaner method would be to convert every component (currently a file) to a module and then use the module on the calling side.
- This will promote the modularity of code, also make it easier to reuse. 
- Also considering that converting code into modules will break the code, we must first convert the codebase into something like classes, and when all files are converted to modules, rewrite the way they are called.

**Are you working on this?**
Yes, very actively.

**References**
- [ES6 Standards](http://es6-features.org/)
- [ES Lint](https://eslint.org/)

**Progress**
- [ ] canvas2svg.js
- [ ] data.js
- [ ] engine.js
- [ ] layout.js
- [ ] miniMap.js
- [ ] Plot.js
- [ ] SAP_DATA.js
- [ ] subcircuit.js
- [ ] UX.js
- [x] canvasApi.js
- [ ] eeprom.js
- [ ] engine_old.js
- [ ] listeners.js
- [ ] modules.js
- [ ] Quin_McCluskey.js
- [ ] search.js
- [ ] testBench.js
- [ ] verilog.js
- [ ] combinationalAnalysis.js
- [ ] embed.js
- [ ] eventQueue.js
- [ ] logix.js
- [ ] Node.js
- [ ] ram.js
- [ ] sequential.js
- [ ] test.js        
- [ ] Counter.js
- [ ] embedListeners.js
- [ ] grading.js
- [ ] metadata.json
- [ ] pace.js
- [ ] RGBLedMatrix.js
- [x] signup.js
- [x] utils.js
- [x] wire.js

sal2701:
Hey @YashKumarVerma great initiative! Can I also help you with this task?

[...]

PR #945, opened 2020-02-24
Linting Fixes in Signup.js

YashKumarVerma:
Part of #905 

Fixes 62 Linting Issues
signup js

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,6 @@
 {
     "extends": "airbnb-base",
-    "globals": {
-    },
+    "globals": {},
     "rules": {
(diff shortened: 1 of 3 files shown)


rathod-sahaab:
It woul be better if you make a WIP request instead of making ~35 requests as they are all related

YashKumarVerma:
> It would be better if you make a WIP request instead of making ~35 requests as they are all related

The PR is not in the state of WIP. 
The idea was to have separate PR for each file as fixing all files would take a **very long time**

Also, citing the discussions on the slack channel, further PRs are put on hold.

PR #962, opened 2020-02-27
Revert "max bitwidth  cross fix"

satu0king:
Reverts CircuitVerse/CircuitVerse#911


jha11aditya:
The name attribute  name='newBitWidth' is of the input tag of the html used in User Xperience file UX.js which is common for all  element's properties input items . So using Jquery to get that element should'nt leave out any element I think!. If there is any kind of exception or I didn't understand the point, please let me know.

PR #981, opened 2020-02-29
Update edit group UI

Shweta0002:
Updated UI for edit group page.


nitin10s:
@tachyons Please close.

PR #1002, opened 2020-03-02
Improved installation instructions

falguniraina:
A minor change in docs to make it easy to copy one file to another. Thanks!

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -30,7 +30,7 @@ You will also need the required software, detailed above. `Redis` and `PostgreSQ
 1. Install bundler : `gem install bundler`
 2. Install dependencies: `bundle install`
 3. Install Yarn dependencies: `yarn`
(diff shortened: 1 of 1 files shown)


falguniraina:
> cp command is not available for windows

Actually this is my first pr so could you please suggest me the changes i am supposed to make for this to get merged!!

Ananthavijay:
I think this change is not required or add a line saying that ````If you are not a windows user this command can be used````

tachyons:
We support multiple operating systems. 'cp' command is only available in Unix like oses like Linux and Mac OSx. So having cp command can potentially confuse windows users. You can rephrase the paragraph to avoid that confusion

aayush-05:
@falguniraina are you still working on this ?

aayush-05:
@nitin10s can we close this one; inactive since a long time?

PR #1018, opened 2020-03-04
 Fix javascript teacher new assignment error

### LINKED ISSUE #103: "EventQueue size exceeded" exception
I'm trying to build a large RAM component from smaller building blocks and eventually the simulator hits an "EventQueue size exceeded" exception.

Once this happens the editor goes into a bad state, failing to draw new components and (perhaps) even save the existing work.

I suppose this limit might exist to keep the simulation contained; if that is the case, perhaps the error should be handled in a graceful way so the circuit doesn't get wasted.

Alternatively, we could increase the limit of this queue or make it configurable on a per project basis.

For a repro, see this circuit: https://circuitverse.org/users/1935/projects/4774
In the 16KB circuit (which is incomplete) I'm trying to add another 4KB circuit; doing so causes the exception.

### LINKED ISSUE #1017: Javascript Error at New Assigment Teacher
**Describe the bug**
When the teacher creates a new assignment the deadline time remaining showing error that says "something went wrong when loading script" instead of the remaining time.
The javascript doesn't load

**To Reproduce**
1. Login
2. Click on Your newly created group
3. Click on Create Assignment

**Expected behavior**
A time remaining should be showing, but the javascript doesn't load at all

**Screenshots**
image

 - OS: Linux Mint 18.1 Cinnamont 64-bit
 - Browser Chrome
 - Version 73

**Additional context**

Yes

Miftahunajat:
Fix Javascript teacher new assignment error

Fixes #1017 

In this PR I remove the quotes mark (') on Javascript file in _form.html.erb on line
`loadRestrictions(JSON.parse('<%= raw @assignment.restrictions %>'));`
Become
`loadRestrictions(JSON.parse(<%= raw @assignment.restrictions %>));`

Before
image

After
image

### DIFF
--- a/app/views/assignments/_form.html.erb
+++ b/app/views/assignments/_form.html.erb
@@ -61,7 +61,7 @@
 
 <script type="text/javascript">
   $(document).ready(function() {
-    loadRestrictions(JSON.parse('<%= raw @assignment.restrictions %>'));
+    loadRestrictions(JSON.parse(<%= raw @assignment.restrictions %>));
     loadDeadlinePicker();
   })
 
--- a/app/views/assignments/_form.html.erb
+++ b/app/views/assignments/_form.html.erb
@@ -61,7 +61,7 @@
 
 <script type="text/javascript">
   $(document).ready(function() {
-    loadRestrictions(JSON.parse(<%= raw @assignment.restrictions %>));
+    loadRestrictions(`JSON.parse(<%= raw @assignment.restrictions %>)`);
     loadDeadlinePicker();
   })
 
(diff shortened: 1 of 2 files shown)


tachyons:
CI is failing

Miftahunajat:
> CI is failing

I think it's because the spec file.
I will try to fix the spec file.

Miftahunajat:
What is the chrome and chromedriver version used in the CircleCI ?

aayush-05:
What if you don't remove the extra 0.1s wait time added in the spec file @Miftahunajat ?

Miftahunajat:
@aayush-05 When the `sleep(0.1)` not removed the CI has already had an error. 
So I create new commit just to add visible: true option so that the test will be waiting for the element to be visible.
It looks like the problem is visibility issue across the browser.

PR #1080, opened 2020-03-08
Fixed The Position of Navbar on top when Scroll

akansha-singh:
Fixes #
ezgif com-video-to-gif(1)


yash15601:
@akansha-singh  fix code climate issues

akansha-singh:
> A little shadow for navbar when scrolled would look good I'll suggest

ezgif com-video-to-gif(2)

PR #1182, opened 2020-03-14
convert all scss font-szie to rem from px

abstrekt:
Changed all ```font-size: xx px``` to rem


abstrekt:
but they are static, it's about the user more than the device i believe, if someone has browser font size to high or low , having rem will take that change, pixel will remain the same no matter the browser settings

abstrekt:
I thought px won't be affected by the browser font size settings, but it seems to work fine, nevermind @nitin10s, thanks for the review.

PR #1206, opened 2020-03-16
Circuits follow the cursor when pasting

### LINKED ISSUE #971: Allow circuits to follow the cursor when pasting
## Description

The current pasting behavior is to directly put down the circuits wherever the cursor is. This can lead to inconvenience and sometimes frustration.

* Often times the circuits don't land where the user wants them, so the user needs to move them manually.
* To make matters worse, the pasted circuits are not auto-selected, so the user needs to re-select whatever have been put down. This can be a nightmare when doing large-scale copy-pasting operations(as seen in #935, point 9 and 10).

## Proposed solution

Instead of putting down the circuits directly when the user hits Ctrl+V, let the circuits hover and follow the cursor until the user left clicks to place them down. This approach has several advantages:

* **Intuitive interface**. No need to figure out where to place the cursor before Ctrl+V.
* **Simple implementation**. Adding new circuits already behaves this way(newly added circuit will follow the cursor until left click). We only need to generalize that behavior to any group of circuits.

## Alternatives

Here is a less "radical" solution: After Ctrl+V we deselect whatever was selected before, and auto-selected the circuits that were put down. This makes it easier to move the circuits manually. But it's not as convenient.

I plan to work on this issue, and I would like to hear some feedback :)

KaustubhDamania:
Fixes #971 

When a circuit is copied, the circuit will follow the cursor until the user left clicks on the simulator area or when the user clicks Ctrl+Z.

ezgif com-video-to-gif

### DIFF
--- a/public/js/listeners.js
+++ b/public/js/listeners.js
@@ -290,7 +290,16 @@ function startListeners() {
         } else {
             e.clipboardData.setData('text/plain', textToPutOnClipboard);
         }
+        // since the 'textToPutOnClipboard' variable is a string, we need to
+        // parse it as a JSON
+        const clipboardDataObject = JSON.parse(textToPutOnClipboard);
+        const objectKeys = Object.keys(clipboardDataObject);
 
+        // the id of the element to be copied would be the key after the key
+        // 'name'
+        const elementId = objectKeys[objectKeys.indexOf('name') + 1]
+        const obj = new window[elementId]();
+        simulationArea.lastSelected = obj;
     });
 
     document.addEventListener('paste', function(e) {


KaustubhDamania:
@satu0king can you review this?

KaustubhDamania:
@satu0king can you review this please?

KaustubhDamania:
@satu0king can you review this please?

satu0king:
Nice Attempt. Please check the comment I have posted in the original issue. 
Your implementation though smart hack, it falls short. It doesn't work correctly when multiple elements are selected. You will have to make changes to paste function in `logix.js`

PR #1181, opened 2020-03-14
Changed no dp default

abstrekt:
temporary Fix 
#1167 

#### Describe the changes you have made in this PR -
* maybe just a temporary fix to issue #1167, about the checkbox
* changed the no dp default.jpg 
### Screenshots of the changes (If any) -
display

PR #1208, opened 2020-03-16
fix squished profile image

mahmodHammad:
Fixes #1204

#### Describe the changes you have made in this PR -
- fix squished profile image
- make it rounded 

### After:
img fix2

### Before:
img bef

PR #1266, opened 2020-03-21
changed no dp default.img

abstrekt:
New PR for #1181
Describe the changes you have made in this PR -

*   changed the no dp default.jpg

##Screenshot After:
Firefox_Screenshot_2020-03-21T10-39-32 046Z


abstrekt:
@nitin10s

nitin10s:
Please attach screenshots here as well.

abstrekt:
> 
> 
> Please attach screenshots here as well.

done.

abstrekt:
this would cause no problem, the purpose of this PR is that the default no dp image was pixelated, & looked bad, this no dp is actually square, looks circular because of transparent edges, if the css on #1208 is set right, this should look the same merging both the PRs, maybe a conflict may arise, which can be fixed.

abstrekt:
In short, this PR changes the default image when no dp is set, #1208 changes the DP shape to be circular @nitin10s

abstrekt:
## screenshot of old default.jpg


Screenshot_2020-03-22-22-31-35-61~01

abstrekt:
@nitin10s  it just changes:
image
to:
Default

aayush-05:
We cannot set the image height to auto, the user can upload big size images which will try to take up complete space for its dimensions then.

abstrekt:
No there is a width mention somewhere @aayush-05

aayush-05:
Imagine a user uploads an image of 300*150 dimensions

abstrekt:
> 
> 
> Imagine a user uploads an image of 300*150 dimensions

I made a rollback,  #1208 was addressing aspect ratio, this PR is intended only for the binary files.

aayush-05:
@abstrekt can you add the code alongwith, just a little condition to use these binary files in case of no profile picture

abstrekt:
> 
> 
> @abstrekt can you add the code alongwith, just a little condition to use these binary files in case of no profile picture

it is already there, since before!

aayush-05:
Ah, got it. My bad!

nitin10s:
@aayush-05

PR #1338, opened 2020-04-01
Reporting issues - prevent spaming with empty messages or one word

mahmodHammad:
Fixes #

- refactor reporting issues 
- adding a missing div
- prevent submitting spamming messages which is less than 2 words
- encapsulate the implementation details of the reporting issues using IIFE

Reporting issues - prevent spamming with empty messages or one word
spam

### DIFF
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -266,7 +266,7 @@
     </div>
       <div class="container">
         <center>
-     <button type="submit" id="report" class="btn btn-primary" >Report</button>
+     <button type="submit" id="report" disabled class="btn btn-primary " >Report</button>
    </center>
     </div>
   </div>
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -241,6 +241,39 @@
         <label class="option inline"> <input type="radio" name="resolution" value="4">4x</label>
       </p>
     </div>
+      
+    <!---issue reporting-system----->
+    <div class="report-sidebar">
+    <a type="button" class="btn btn-primary text-light" data-toggle="modal" data-target=".issue"> <span class="fa fa-bug" ></span>&nbsp;&nbsp;Report an issue</a>
+    </div>
+      <!---issue reporting system-modal---->
+<div class="modal fade issue" tabindex="-1" role="dialog" aria-labelledby="mySmallModalLabel" aria-hidden="true">
+  <div class="modal-dialog modal-sm">
+    <div class="modal-content"> 
+      <div class="container my-2">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      <div class="container  text-center">
+       <h4>Report an issue</h4>  
+    </div>
+    <hr>
+     <div class="container my-2 text-center" id="result">
+    </div>
+    <label id="report-label" style="font-weight: lighter" ><b>Describe your issue:</b></label>
+    <div class="form-group">
+    <textarea class="form-control border-primary"  id="issuetext" rows="3"></textarea>
+    </div>
+      <div class="container">
+        <center>
+     <button type="submit" id="report" class="btn btn-primary" >Report</button>
+   </center>
+    </div>
+  </div>
+  </div>
+</div> 
+</div>
+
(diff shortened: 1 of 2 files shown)


yash15601:
@mahmodHammad  change "input" spelling in popup

yash15601:
You can use 'require'  in input field  this will not send empty text I think

mahmodHammad:
> Rather than 2 words let's keep the error on not empty.
> Also, is it possible this error message is displayed under the text field [like a validation] rather than a separate alert?

actually I am thinking  of  making the button disabled if  the input message is less than 2 words instead of that alert message
@nitin10s  what do you think about that?

nitin10s:
Exactly solves the problem in a lot less code.

mahmodHammad:
> You can use 'require' in the input field this will not send an empty text I think
@yash15601  I have tried it, but for some reason, it doesn't work, also we want to prevent the spam messages which has one word, like "test" or "fiwhfhw"

yash15601:
Did you used require  or required According  to me it should work

yash15601:
Looks good great  work 😃

mahmodHammad:
I tried required not require

nitin10s:
I don't think we should have a minimum word limit. I can't decide what will be appropriate. I would prefer the Can't send an empty form.

mahmodHammad:
> I don't think we should have a minimum word limit. I can't decide what will be appropriate. I would prefer the Can't send an empty form.

I used just one-word limit, users will never provide us anything useful will one word.

public/js/logix.js
```diff
@@ -1493,3 +1493,30 @@ CircuitElement.prototype.generateVerilog = function() {
...
+     if(message.split(' ').length>1){
+     message += "\nURL: " + window.location.href;
+     message += `\nUser Id: <%= user_signed_in? ? " #{current_user.id.to_s} : #{current_user.name}" : "Guest user" %>`
+     postUserIssue(message)
+     $('#issuetext').hide();
+     $('#report').hide();
+     $('#report-label').hide();
+     }else{
+        //  actually i wanted the message to be  (don't spam) XD 
+         alert("your input message can not be less than two words")
```
nitin10s, line 1510: Write: The response should have at least 2 words.
Remove the comment

mahmodHammad: please check it again

mahmodHammad:
@nitin10s 
i removed the one word limit ,please check it again
issue

mahmodHammad:
@nitin10s  Please check it again after changes

nitin10s:
Can you please resolve the AccessLint check @mahmodHammad

app/views/simulator/edit.html.erb
```diff
@@ -241,6 +241,39 @@
...
+        </button>
+      <div class="container  text-center">
+       <h4>Report an issue</h4>  
+    </div>
+    <hr>
+     <div class="container my-2 text-center" id="result">
+    </div>
+    <label id="report-label" style="font-weight: lighter" ><b>Describe your issue:</b></label>
+    <div class="form-group">
+    <textarea class="form-control border-primary"  id="issuetext" rows="3"></textarea>
```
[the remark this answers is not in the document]
mahmodHammad: @nitin10s 
i don't know how to fix it because the label already exists!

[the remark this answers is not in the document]
mahmodHammad: I will try to put the label in the same div of the textarea, I hope this will fix it.
what do you think? @nitin10s @yash15601

public/js/logix.js
```diff
@@ -1493,3 +1493,37 @@ CircuitElement.prototype.generateVerilog = function() {
...
+    let valid = false
+    $('#issuetext').on('input', function(){
+        message = $('#issuetext').val().trim();
+        valid = message.length>0;
+        $('#report').attr("disabled",!valid);
+    })
+    
+    $('#report').click(function(){
+         message += "\nURL: " + window.location.href;
+         message += `\nUser Id: <%= user_signed_in? ? " #{current_user.id.to_s} : #{current_user.name}" : "Guest user" %>`
```
tachyons, line 1511: @mahmodHammad

mahmodHammad: I think i unconscious converted it to a string.
I don't remember how i did that, but i couldn't test on the development environment, thats why this happened

mahmodHammad: I will come back with a fix after a couple of weeks

PR #1343, opened 2020-04-02
Marquee effect for example title

harshgujrati:
Fixes #


harshgujrati:
@nitin10s  please review

harshgujrati:
> @harshgujrati I think so we should keep the fit titles in place and the longer titles display marquee effect on hover and remain as it is when not hovered. For example
> Circuit Name : Shivansh will remain as it is as it fits in the place.
> Circuit Name : Really Long Circuit Name . Now this will get converted to Really Lon... and then will display the complete circuit name on hover with a marquee effect.
> I would like to know the views of @nitin10s @satu0king and @tachyons on this.

I think this can create non uniformity and will not look good but i may be wrong, waiting for your views @nitin10s  @satu0king  @tachyons

harshgujrati:
Not an optimum solution for user with visual impairments, closing the PR

PR #1059, opened 2020-03-07
Add Recent Circuits

Shivansh2407:
Fixes #1058 

#### Describe the changes you have made in this PR -
I have added the recent circuits section to the Homepage.

### Screenshots of the changes (If any) -
Capture

Issue #982, opened 2020-03-01
No of Views can be easily increased by requesting page in a loop by curl or something else


ciphereck:
**Describe the bug**
I request to get this page in the browser https://circuitverse.org/users/18390/projects/56841 and saw no of views increase by one. But then I added a for loop in a shell script to make a curl request again and again, and it's increasing no of views. 
Suppose a person creates a project, he/she can easily increase their number of views using multiple iterations of making that request from command line or javascript.

**To Reproduce**
1. Go to 'Terminal
2. write command: "for ((i=0; i<=10;i++)); do curl https://circuitverse.org/users/18390/projects/56841; done"
3. Check that the number of views got increased by same amount by visiting https://circuitverse.org/users/18390/projects/56841

**Expected behavior**
A user must not be able to increase its number of views like this, not at least this easily.

**Screenshots**
I did the same to increase the number of views of a project from 25 to 296.
![1](https://user-images.githubusercontent.com/24265219/75621023-924dc580-5bb5-11ea-9c8d-69d4b91f882f.png)
![2](https://user-images.githubusercontent.com/24265219/75621024-9679e300-5bb5-11ea-87de-76848be00038.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Chrome
 - Version Version 76.0.3809.100 (Official Build) (64-bit)

**Additional context**
If we use a number of views to find trending projects, it's going to create a lot of problems there.

**Are you working on this issue? (Yes/No)**, 
Maybe, depending on the expertise needed.

KaustubhDamania:
Maybe this issue can be fixed by incrementing the views count only if the session id encountered is new :thinking:

tachyons:
https://github.com/CircuitVerse/CircuitVerse/pull/414

pranavmodx:
I would like try this issue. If someone could give me a head start on the section of code handling user views count, it would be great.

ciphereck:
Hi @pranavmodx 
Sorry for the late reply. Check #414 PR, it contains some work, maybe you can take it from there.

Ananthavijay:
@tachyons this can be closed.

ciphereck:
@Ananthavijay yeah sure. Now it can after Or #1346

PR #414, opened 2019-07-22
Add ahoy to improve project view detection

ankitkataria:
Fixes #246 , Work done on top of PR #339 by @animysore

#### Describe the changes you have made in this pr -

- Add Ahoy gem
- Modify project controller to:
create an Ahoy event whenever a unique visit lands on /project/id as Visited project <id>
call @project.increase_views only if it is a unique visit

### Screenshots of the changes (If any) -
N.A

### Todo - 
- [x] Fix project view test

PR #1346, opened 2020-04-02
Add ahoy to improve project view detection

### LINKED ISSUE #246: Number of views for a circuit increases upon page reload

ciphereck:
Fixes #246 and #982  , Work done on top of PR #414 by @ankitkataria @animysore

Describe the changes you have made in this pr -

- Add Ahoy gem

- Modify project controller to:
create an Ahoy event whenever a unique visit lands on /project/id as Visited project
call @project.increase_views only if it is a unique visit

To check the increment in the project views, you need to rebase it with PR #1059 

NA

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -15,6 +15,45 @@ ActiveRecord::Schema.define(version: 2019_12_08_045111) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "ahoy_events", force: :cascade do |t|
+    t.bigint "visit_id"
+    t.bigint "user_id"
+    t.string "name"
+    t.json "properties"
+    t.datetime "time"
+    t.index ["name", "time"], name: "index_ahoy_events_on_name_and_time"
+    t.index ["user_id"], name: "index_ahoy_events_on_user_id"
+    t.index ["visit_id"], name: "index_ahoy_events_on_visit_id"
+  end
+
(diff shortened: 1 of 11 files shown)


ciphereck:
@tachyons
Please review!

db/schema.rb
```diff
@@ -245,7 +284,7 @@
     t.string "uid"
     t.string "profile_picture_file_name"
     t.string "profile_picture_content_type"
-    t.bigint "profile_picture_file_size"
+    t.integer "profile_picture_file_size"
```
tachyons, line 287: related ?

ciphereck: I don't think so. I guess when original PR was created, it was an integer. That's why this change. I'll update it right away. @tachyons

PR #1391, opened 2020-04-11
Fix footer responsiveness for all displays

aayush-05:
**Screenshots**
image
image
image
image


aayush-05:
@tachyons @nitin10s please have a look at this

aayush-05:
Makes sense @nitin10s . Can you redesign the footer a little bit then, so that it don't feel empty on the left side (specially with desktop screen size) ?

PR #1428, opened 2020-05-10
deselect

abstrekt:
deselect feature on esc


abstrekt:
@satu0king  will this suffice?

PR #1436, opened 2020-05-15
Dec-bin-hex converter

ryuk156:
This is a new feature which was include in todo list  in one of the issue which is converter which convert the decimal values  into binary and hexadecimal values ,the PR which includes the function to convert the values and the script in "simulator/edit.html.erb" to create a dialog which has input fields and a reset button to reset the input fields,since the backspace is not woking i have to include this reset button

Video:
ezgif-7-44aedd9b2223


ryuk156:
@satu0king  @nitin10s  @tachyons  review please

public/js/dec-bin-hex-converter
```diff
@@ -0,0 +1,47 @@
+(function(){
```
mahmodHammad, line 1: that's great you used IFIE to encapsulate your code

app/views/simulator/edit.html.erb
```diff
@@ -329,4 +332,33 @@
       <script src="/js/combinationalAnalysis.js"></script>
       <script src="/js/Quin_McCluskey.js"></script>
     <% end %>
-    <script src="/js/canvas2svg.js"></script>
\ No newline at end of file
+    <script src="/js/canvas2svg.js"></script>
+    
+    <script type="text/javascript">
```
mahmodHammad, line 337: It would be better if you put this code into a separate js file or put it inside the logix.js .if you have some free time.
but it will be refactored any way later

public/js/dec-bin-hex-converter
```diff
@@ -0,0 +1,47 @@
...
+                    to : function (baseTo) {
+                        return parseInt(num, baseFrom).toString(baseTo);
+                    }
+                };
+            }
+        };
+    };
+        
+    // binary to decimal
+    ConvertBase.bin2dec = function (num) {
```
mahmodHammad, line 16: I think you should use a prototype for the inheritance that could happen later

ConvertBase.prototype.bin2dec

public/js/dec-bin-hex-converter
```diff
@@ -0,0 +1,47 @@
+(function(){
+
+    var ConvertBase = function (num) {
```
mahmodHammad, line 3: why did not you make to , from as parameters for this function. instead of this complicated implementation

`function (num ,from , to)`
`return  parseInt(num, baseFrom).toString(baseTo);`

mahmodHammad:
@ryuk156 
great job. I did not test it , but i believe it should work
this requested changes is only for best practices

ryuk156:
Will makes changes in code soon after this once merged will be optimizing code more

ryuk156:
@nitin10s @satu0king  @tachyons  review please

satu0king:
Merged in #1455

PR #1452, opened 2020-05-23
Allow blank issues

MrBartusek:
I think limiting possible issues to only two categories is a quite bad idea. I want to add an issue that is more like issue regarding code base and it doesn't fit a `bug` or `feature request`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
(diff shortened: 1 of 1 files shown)


MrBartusek:
@nitin10s please review

nitin10s:
We can make a new category for such issues to organize them better, what do you suggest be the category name?

MrBartusek:
@nitin10s i don't think so, there is no reason to create category for only one issue

nitin10s:
I meant for all the issues that people might have regarding the codebase.

MrBartusek:
@nitin10s Well, for that kind of issue you can create `Codebase` template but that would be very unpopular so there is not much sense in it

PR #1463, opened 2020-05-27
removed backspace to delete selection

abstrekt:
Fixes #

I have removed 'keyCode == 8'  condition from delete selected listener, keyCode 8 is backspace, but some browsers have default listener with backspace, making it useless on browsers such as firefox,
I also added a condition to fire delete selected only when something is selected.

### DIFF
--- a/public/js/listeners.js
+++ b/public/js/listeners.js
@@ -161,8 +161,8 @@ function startListeners() {
             }
         }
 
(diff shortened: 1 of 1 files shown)


abstrekt:
This was actually causing a bug in my #1438 PR

nitin10s:
@abstrekt Is it still valid?

abstrekt:
no i will close this

PR #1473, opened 2020-05-31
Add Collaborators API

Nitish145:
1. Adds the following collaborators endpoints
    - `/api/v1/projects/:project_id/collaborators` to GET & POST collaborators.
    - `/api/v1/projects/:project_id/collaborators/:id` to DELETE a collaborator.

2. Add specs for each controller action and possible test cases.

### The updated `docs` are available [here](https://nitish145.github.io/slate/)


Nitish145:
@tachyons please check!!

app/controllers/api/v1/collaborators_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    @added_mails = []
+
+    newly_added.each do |email|
+      user = User.find_by(email: email)
+      if user.present?
+        @added_mails.push(email)
+        Collaboration.where(project_id: @project.id, user_id: user.id).first_or_create
+      end
+    end
```
tachyons, line 33: We have same logic in one more place, right? Can we extract this ?

app/controllers/api/v1/collaborators_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def set_collaborator
+      @collaborator = @project.collaborators.find(params[:id])
+    end
+
+    def collaborator_params
+      params.require(:collaborator).permit(:project_id, :emails)
+    end
+
+    def parse_mails_except_current_user(mails)
```
tachyons, line 64: Worth extracting to a PORO(Plain old ruby object) ?

Nitish145: Sure!! Will do :)

app/controllers/api/v1/collaborators_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+class Api::V1::CollaboratorsController < Api::V1::BaseController
+  before_action :authenticate_user!
+  before_action :set_project
+  before_action :check_author_access, only: %i[create destroy]
+  before_action :set_collaborator, only: %i[destroy]
+
+  # /api/v1/projects/:project_id/collaborators
+  def index
+    @collaborators = paginate(@project.collaborators)
```
tachyons, line 11: Why we are exposing this endpoint to all, shoudn't this be restricted based on access to project?

app/controllers/api/v1/collaborators_controller.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class Api::V1::CollaboratorsController < Api::V1::BaseController
+  before_action :authenticate_user!
+  before_action :set_project
+  before_action :check_author_access, only: %i[create destroy]
```
tachyons, line 6: prefer except to avoid accidentally exposing a new point

app/controllers/api/v1/collaborators_controller.rb
Nitish145: I thought we can show the collaborators to any authenticated user, does that make sense ? @tachyons

Nitish145:
@tachyons any more changes here ?

app/controllers/api/v1/collaborators_controller.rb
tachyons: But that is public project alone, right? cc @satu0king

Nitish145: Yeah, i missed that part.. will add a check for this.. @tachyons

Nitish145: @tachyons made the required changes please check..

PR #1475, opened 2020-06-01
Hotkeys [Task 1.2]

abstrekt:
This feature will enable users to set custom key-bindings for the simulator's functionality.


public/js/hotkey_binder/defaultKeys.json
```diff
@@ -0,0 +1,30 @@
...
+  "Insert Label": "Ctrl + L",
+  "Label Direction Up": "Alt + Up",
+  "Label Direction Down": "Alt + Down",
+  "Label Direction Left": "Alt + Left",
+  "Label Direction Right": "Alt + Right",
+  "Move Element Up": "W",
+  "Move Element Down": "S",
+  "Move Element Left": "A",
+  "Move Element Right": "D",
+  "Hotkey Preference": "Ctrl + .",
```
satu0king, line 28: Ctrl + '/' is a better hotkey for preferences as Google Docs uses that.

public/js/hotkey_binder/model/addShortcut.js
```diff
@@ -0,0 +1,86 @@
+//Assign the callback func for the keymap here
+
+const addShortcut = (keys, action) => {
+	let callback;
+	switch (action) {
+		case "New Project":
+			callback = newProject;
+			break;
+		case "Save Online":
```
satu0king, line 9: Maybe have a JSON map instead of using switch case? 

like 
```
shortcutMap = {
    "New Project": newProject
};
// Then u can simply do 
if(action in shortcutMap) {
    return shortcutMap[action];
}
```

public/js/hotkey_binder/model/forElements.js
```diff
@@ -0,0 +1,11 @@
...
+// 	if (simulationArea.lastSelected && simulationArea.lastSelected.newElement)
+// 		simulationArea.lastSelected.delete();
+// 	const obj = new window[ele]();
+// 	simulationArea.lastSelected = obj;
+// 	smartDropXX += 70;
+// 	if (smartDropXX / globalScope.scale > width) {
+// 		smartDropXX = 50;
+// 		smartDropYY += 80;
+// 	}
+// };
```
satu0king, line 11: If this whole file is commented, can we remove it?

public/js/hotkey_binder/model/addShortcut.js
```diff
@@ -0,0 +1,86 @@
...
+		case "Move Element Right":
+			callback = moveElement('right');
+			break;
+		case "Hotkey Preference":
+			callback = openHotkey;
+			break;
+			
+		default:
+			callback = () => true;
+			break;
```
satu0king, line 78: Why default callback? Shouldn't it throw an error? or ignore.

public/js/hotkey_binder/defaultKeys.json
```diff
@@ -0,0 +1,30 @@
+/**Add more elements here, along with a valid value for key
+* Elements keys must have the same name as their ID
+**/
+
+defaultKeys = {
+  "New Project": "Shift + N",
```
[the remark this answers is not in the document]
abstrekt: Surely.

public/js/hotkey_binder/defaultKeys.json
abstrekt: Unfortunately the key '/' came under the unsupported keys, hence I had added  "Ctrl + ." for it temporarily, looking for a suitable alternative.

public/js/hotkey_binder/model/addShortcut.js
abstrekt: I'll try on implementing as suggested.

public/js/hotkey_binder/model/forElements.js
abstrekt: Of course, I kept it incase it might it is decided to bring in elements drop shortcut, I shall remove this file.

public/js/hotkey_binder/model/addShortcut.js
abstrekt: If new shortcuts are added correctly, it will never hit the default case, so far it doesn't, I just added it as code-climate suggested so. If you suggest I shall remove it.

public/js/data.js
```diff
@@ -1007,8 +990,8 @@ function generateImage(imgType, view, transparent, resolution, down = true) {
 
 }
 
-if (logix_project_id == 0)
-setTimeout(promptSave,120000);
+// if (logix_project_id == 0)
+// setTimeout(promptSave,120000);
```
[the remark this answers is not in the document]
abstrekt: > 
> 
> I tested, it is working on the whole. It is working on Mac as well. There are some issues:
> 
>     1. Delete - backspace key is no longer working. Note that backspace is delete on mac.
> 
>     2. CMD/Ctr + R is not refreshing page
> 
>     3. Copy/Paste is no longer working
> 
>     4. Check other regular shortcuts
> 
> 
> Don't do preventDefault or stopPropagate if that shortcut is not mapped. Looks like many browser shortcuts are disabled. Like Ctr + L should select URL of the current tab.
> 
> CMD is working on Mac as required. I found two files which have `value.replace("Ctrl", "Meta")`, do we need to have them in 2 places?
> 
> Also it will be helpful if you could tell which files are written by you and which files are from the library.

I'll rectify these bugs soon.

& I the plugin files are named as xxxx.plugin.js, there are currently 2 of them in the hotkey implementation, shortcuts.plugin.js under modal dir & normalize.plugin.js under modal/normalize dir, rests are written by me.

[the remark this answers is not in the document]
abstrekt: > 
> 
> Why is this commented out?

I comment them out every time I start working on the simulator for a smooth workflow, this time I forgot to uncomment it, I shall do it in the next commit.

public/js/hotkey_binder/model/addShortcut.js
abstrekt: > 
> 
> CMD is working on Mac as required. I found two files which have `value.replace("Ctrl", "Meta")`, do we need to have them in 2 places?
> 
value.replace("Ctrl", "Meta")`: one for the shortcuts, one for restricted shortcuts, I could have have an util func for that..didn't....

abstrekt:
> 
> 
> I tested, it is working on the whole. It is working on Mac as well. There are some issues:
> 
>     1. Delete - backspace key is no longer working. Note that backspace is delete on mac.
> 
>     2. CMD/Ctr + R is not refreshing page
> 
>     3. Copy/Paste is no longer working
> 
>     4. Check other regular shortcuts
> 
> 
> Don't do preventDefault or stopPropagate if that shortcut is not mapped. Looks like many browser shortcuts are disabled. Like Ctr + L should select URL of the current tab.
> 
> CMD is working on Mac as required. I found two files which have `value.replace("Ctrl", "Meta")`, do we need to have them in 2 places?
> 
> Also it will be helpful if you could tell which files are written by you and which files are from the library.

1. Delete - backspace fixed
2. browsers shortcuts now working
3. Copy/Paste working

@satu0king rectified everything, please check.

abstrekt:
> 
> 
> This doesn't happen normally, but this happened once, I don't know why I pressed N for the new shortcut and clicked Enter and this appeared.
> image

I'm unsure about the cause,
Does it get fixed if you hit reset to default?

nitin10s:
1. There is no indication that the dialogue box is scrollable - add a scrollbar
2. Align the shortcut titles with the heading "Command"
3. Bolden the headings
4. When an already used shortcut is entered. In the error message, replace "Override?" with "Press Enter to override."

nitin10s:
> > This doesn't happen normally, but this happened once, I don't know why I pressed N for the new shortcut and clicked Enter and this appeared.
> > image
> 
> I'm unsure about the cause,
> Does it get fixed if you hit reset to default?

I reloaded and it worked fine. Didn't try that button

abstrekt:
> 
> 
>     1. There is no indication that the dialogue box is scrollable - add a scrollbar
It actually has a scrollbar, not so visible in chrome browser, need to change the backdrop filter color, UI issue
>     2. Align the shortcut titles with the heading "Command"

>     3. Bolden the headings
Headings are already bold?
>     4. When an already used shortcut is entered. In the error message, replace "Override?" with "Press Enter to override."
Okay

nitin10s:
I am able to set shortcuts like Ctrl + C that anyway work in a default way. So on entering such shortcuts - An error message should be displayed something like "Looks like this is a default system shortcut and can't be applied! Try a new combination." and disable pressing Enter.

abstrekt:
> 
> 
> I am able to set shortcuts like Ctrl + C that anyway work in a default way. So on entering such shortcuts - An error message should be displayed something like "Looks like this is a default system shortcut and can't be applied! Try a new combination." and disable pressing Enter.

we can just add such combos to restricted array

public/js/UX.js
```diff
@@ -469,3 +464,14 @@ $("#octalInput").on('keyup', function () {
     var x = parseInt($("#octalInput").val(), 8);
     setBaseValues(x);
 })
+
+// some hack
```
satu0king, line 468: Please document what hack is this and why we need it

public/js/hotkey_binder/model/addShortcut.js
satu0king: @abstrekt throw an error, its better to know something went wrong than nothing happening.

satu0king:
Merged

PR #1477, opened 2020-06-02
Add project preview image API

Nitish145:
Add `/api/v1/project/:id/image_preview` to GET public project's full image url

```
{
  "project_preview": "http://localhost:3000/img/default.png"
}
```

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -4,7 +4,7 @@ class Api::V1::ProjectsController < Api::V1::BaseController
   include ActionView::Helpers::SanitizeHelper
 
   before_action :authenticate_user, only: %i[index show]
-  before_action :authenticate_user!, except: %i[index show]
+  before_action :authenticate_user!, except: %i[index show image_preview]
   before_action :load_index_projects, only: %i[index]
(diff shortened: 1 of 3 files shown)


Nitish145:
@satu0king @tachyons  please check!!

satu0king:
Looks good to me! Thanks 
@tachyons can you review and merge the same.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -34,6 +34,11 @@ def show
     render json: Api::V1::ProjectSerializer.new(@project, @options)
   end
 
+  def image_preview
+    @project = Project.open.find(params[:id])
```
tachyons, line 38: What about private projects? They don't have preview images ?

Nitish145: They do have but @satu0king asked me for public projects only, is that right @satu0king ?

satu0king: Yes idea is that we don't want private project's preview image to get leaked this way. @tachyons is there a better way or its fine to do that?

tachyons: @satu0king We can return 403 if the user is not authorized to view this project, right?

PR #1483, opened 2020-06-04
Make urls more friendly using friendly ids

tachyons:
This is to make urls more friendly to read

Previous
`users/165/projects/65675`
New
`users/165/projects/flip-flop`

This feature is backward compatible and all the previous urls work as before


satu0king:
Looks great! But what happens if there are duplicate names? We need to test the backward compatibility part before deploying.

tachyons:
It will append unique postfix to fix duplication, fixed all the backward compatibility issues I could find, you can check beta.circuitverse.org

satu0king:
What happens when the circuit name is changed?

tachyons:
We aren't updating slug when the circuit name changes, but we can add that behaviour by just adding one line to the code, previous urls still work for the backward compatbility

tachyons:
https://stackoverflow.com/a/17765270

PR #1510, opened 2020-06-26
Authentication Enhancements

Nitish145:
1. Add the following endpoint
      - `/api/v1/password/forgot` to send reset password link.
      - `/api/v1/public_key.pem` to GET jwt public key

2. Fix rubocop linting issues for authentication and users specs.

3. Add specs for the above two endpoints


Nitish145:
@tachyons please check!

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -29,6 +27,20 @@ def signup
...
+    @user = User.find_by!(email: params[:email])
+    # sends reset password instructions to the user's mail if exists
+    @user.send_reset_password_instructions
+    render json: { message: "password reset instructions sent to #{@user.email}" }
+  end
+
+  # GET /public_key.pem
+  def public_key
+    public_key = File.open(Rails.root.join("config", "public.pem"), "r:UTF-8")
+    send_file public_key
```
tachyons, line 41: What about making this a json so that this will be easier to consume at client level

Nitish145: Yes!! Would be easier..

Nitish145:
@tachyons please check!!

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -29,6 +27,44 @@ def signup
     end
   end
 
+  # POST api/v1/oauth/login
+  def oauth_login
+    @user = User.find_by!(email: params["email"])
+    token = JsonWebToken.encode(
+      user_id: @user.id, username: @user.name, email: @user.email
+    )
+    render json: { token: token }, status: :accepted
```
tachyons, line 36: Wait, you can pass arbitrary email addresses and this endpoint will return a valid token for that user !!. Big security flow

Nitish145: @tachyons is that so? what we are doing here is just registering or finding the user with that email/info in the DB.. & imo devise handles email validation on its own, so it won't allow any email to generate a valid token..
Correct me if i am wrong..

Nitish145: @tachyons

tachyons: Where are you validating that ?

Nitish145: @tachyons pushed the appropriate changes discussed, can you please once check.. if given a green signal, will add tests for the same..

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -29,27 +31,33 @@ def signup
...
-      @user = User.new(oauth_signup_params)
-      @user.password = Devise.friendly_token[0, 20]
-      @user.save!
+      @user = User.create!(
+        name: @response["name"],
+        email: @response["email"],
+        password: Devise.friendly_token[0, 20],
+        provider: params[:provider],
+        uid: @response["id"] || @response["sub"]
+      )
```
tachyons, line 53: There can be scenarios like empty email from the response, for instance you can create a Facebook account with email linked to it. So handle failing to create user record with proper error message

Nitish145: @tachyons can you please check once!!

app/models/user.rb
```diff
@@ -67,6 +67,16 @@ def self.from_omniauth(access_token)
     user
   end
 
+  def self.from_oauth(oauth_user, provider)
+    User.create!(
+      name: oauth_user["name"],
+      email: oauth_user["email"],
```
tachyons, line 73: what if email is blank ?

Nitish145: @tachyons `create!` will raise validation error which would be handled in the rescue block.. with appropriate error message..

```
rescue ActiveRecord::RecordInvalid
   api_error(status: 422, errors: @user.errors)
end
```

tachyons: @user is not assigned in this case, write a spec for the scenario where email is blank

Nitish145: Oh is it.. will cover that case..

Nitish145: Thanks @tachyons for this insight!! learned something new about how `bang!` only raises an exception.. Thanks :v:

Nitish145: btw made the required changes and added specs.. please check @tachyons

spec/requests/api/v1/authentication_controller/oauth_signup_spec.rb
```diff
@@ -45,6 +45,20 @@
...
+      before do
+        post "/api/v1/oauth/signup", params: {
+          "access_token": "empty_email_token",
+          "provider": "google"
+        }, as: :json
+      end
+
+      it "return status 422 and should have jsonapi errors" do
+        expect(response).to have_http_status(422)
+        expect(response.parsed_body).to have_jsonapi_errors
```
tachyons, line 58: Can you check the exact error message, need to make end user aware that signup failed due to missing email in the linked account

Nitish145: Done! @tachyons

Nitish145:
@tachyons anymore changes here ?

PR #1520, opened 2020-07-06
Color themes

abstrekt:
This PR tries to add themes to the simulator.


abstrekt:
> 
> 
>     1. The Insert Sub-Circuit feature isn't working properly. The selected option isn't being displayed.
yeah, that bug came in while merging the branches, UI bug, will fix that with new PR that will consists most fixes.

>     2. Whilst drawing the wire, the colour used is black which isn't visible
I guess you're talking about dark theme, I'll fix it
>     3. image The dropdowns are not visible in the G&W and Lite-Born
 You mean the drop icon right? I'll see what can be done.
> 
>     image You can see how the currently Night sky is selected and originally the theme was Lite-Born. This functionality isn't visible when we choose Default Theme and High Contrast theme.
Will give it some contrast
> 5. In colour blind theme
> image We should increase the contrast actually, rather than light greyish background colour modal, use a darker colour. Also the placeholders aren't visible.
Should the transparent blur dialogs effect be disabled in color blind theme?
> 6. This brown colour looks a little low in contrast, maybe change it? Also, The ratio is also low hovering any circuit element section.
 >
Which brown is it exactly?
> 7. The colour contrast is very low, use a different colour
>
Stroke color?
> image
> 8. Rename the menu name from Colour Theme to Themes.
>
Ok
> 9. Up/Down Keyboard functioning is not working. Also remove the text - Up/down key to preview.
>
I disabled it, it wasn't working properly.
> 10. Paddings are inconsistent.
> image
It just haven't loaded properly, try refreshing
> High contrast theme looks great! But I am not sure if the circuit colours are fine.
>
Suggestion on the element color theme would be helpful.

nitin10s:
Yes, I meant stroke colour in 7.

6. I meant the brown In the high contrast theme.

public/src/node.js
```diff
@@ -431,27 +432,34 @@ export default class Node {
...
+        let wire = colors['node_norm'];
+        const wireConnect = colors['wire_con']
+        const wirePow = colors['wire_pow']
+        const wireLose = colors['wire_lose']
+        const node = colors['node'];
+
+        if (this.bitWidth == 1) wire = [wireConnect, wirePow][this.value];
+        if (this.value == undefined) wire = wireLose;
         if (this.type == 2) this.checkHover();
-        if (this.type == 2) { drawCircle(ctx, this.absX(), this.absY(), 3, color); } else { drawCircle(ctx, this.absX(), this.absY(), 3, 'green'); }
```
satu0king, line 451: Something looks wrong here. Please use better names here. Names of color variables cannot be `wire` and `wireConnect`

public/src/modules/colors.js
```diff
@@ -0,0 +1,30 @@
+// const colors = {
+//     "hover_select": hover_select,
+// }
+
+const getColors = () => {
```
satu0king, line 5: This function is called in many places. This function extracts the property of each style from the document model right? This is not very efficient and can cause performance issues. Store this in an object and reference this object everywhere instead? Maybe some thing like have a globalObject like `selectedColorTheme` and do `selectedColorTheme = getColors()` when theme changes. Then everywhere else we can refer colors by `selectedColorTheme['color-code'].

satu0king:
I reviewed just the code changes. The JS changes looks good overall except the couple of things above 👍

public/src/node.js
abstrekt: can I prefix them with something like 'color-wire', color-wireConnect'?

public/src/modules/colors.js
abstrekt: Yes, this will be better, I will make the changes.

public/src/node.js
abstrekt: can I prefix them with the word 'color'?

nitin10s:
image
Colour-Blind Theme: The LED is not visible [Outlining the element would be benefecial]
The colour for wire when selected can also be made better. Grey on white doesn't have the best contrast ratio.

abstrekt:
> 
> 
> image
> Colour-Blind Theme: The LED is not visible [Outlining the element would be beneficial]. Same goes for Lite-Born Spring and G&W.
> The colour for wire when selected can also be made better. Grey on white doesn't have the best contrast ratio.

but the LED has been like this since the original color scheme

abstrekt:
I have added a darker shade of grey to the wire-draw on the color blind theme , if you want we can make it black.

nitin10s:
The darker shade is fine but it doesn't signify that the wire is selected, maybe use a bright dark blue for the same? I know the LED colour is the same but in the original colour scheme, the LED looks more distinct (not sure but the LED outline looks darker in the original scheme compared to the newer ones)

nitin10s:
@abstrekt The design for each theme card changes as we switch between themes, I understand it is obvious but is it possible we keep them consistent for all the themes. The thing is they get a little distracting and making them constant for all the themes would be quite better. Whatsay?

abstrekt:
> 
> 
> @abstrekt The design for each theme card changes as we switch between themes, I understand it is obvious but is it possible we keep them consistent for all the themes. The thing is they get a little distracting and making them constant for all the themes would be quite better. Whatsay?

you mean the bottom title bar? where the title & radio button remains? It surely can be done but we'd need some color schemes that would suit all the themes

nitin10s:
Yes, that. Yeah, that's needed, can you come up with something that suits all themes? I was thinking about a shade of dark blue.

public/src/modules/Counter.js
```diff
@@ -101,10 +103,14 @@ export default class Counter extends CircuitElement {
         var ctx = simulationArea.context;
         var xx = this.x;
         var yy = this.y;
-
+        //        
+        ctx.fillStyle = colors['fill'];
+        ctx.strokeStyle = colors['stroke'];
+        ctx.fill();
+        ctx.stroke();
```
satu0king, line 110: The above lines seem unnecessary, similarly in other places

public/src/modules/Stepper.js
```diff
@@ -50,10 +52,14 @@ export default class Stepper extends CircuitElement {
      */
     customDraw() {
         var ctx = simulationArea.context;
-
+        //        
+        ctx.fillStyle = colors['fill'];
+        ctx.strokeStyle = colors['stroke'];
+        ctx.fill();
+        ctx.stroke();
```
satu0king, line 59: Here again

satu0king:
Merged

PR #1525, opened 2020-07-10
Implement simple forum using simple_discussion gem

tachyons:
Implement simple discussion gem

Screenshot 2020-07-10 at 8 15 43 PM


nitin10s:
What will be the entry point for this discussion forum?

aayush-05:
> What should be the entry point for this discussion forum? Navbar Link?

It's in the footer currently.

nitin10s:
No, I meant footer isn't really the right entry point, so where should we add the link. considering navbar already has way too many options.

tachyons:
What exactly is the issue with having forum link in the footer ?

nitin10s:
The forum should be considered a primary feature and thus should have a prominent entry point. Footer nav is for when your users reach the end, and instantly accessible navigation tool waiting to help them move freely around the website. In short, more of secondary navigation.

What we can do is put the forum link in the simulator navbar so that's it's highly visible to people using the simulator.

PR #1554, opened 2020-07-13
Fix Navbar bugs

aayush-05:
**Changes**
1. Fixed width issue with select tag
image
2. Made the search icon right aligned on navbar collapse
image
3. Made Learn and Documentation open up in new tab


aayush-05:
Will make a new one

PR #1581, opened 2020-08-03
Projects Enhancements

Nitish145:
- User Projects / Featured Projects (public only)  are now available without authentication as they should be..
- User Projects /Featured Projects (private) would still require authentication.
- Add user_favourites `/api/v1/users/:id/favourites` endpoint.
- Add / Update specs for the above mentioned changes.


Nitish145:
@tachyons please check!!

app/controllers/api/v1/projects_controller.rb
```diff
@@ -113,10 +115,25 @@ def load_index_projects
     end
 
     def load_user_projects
-      @projects = if @current_user.id == params[:id].to_i
+      # if user is not authenticated or authenticated as some other user
+      # return only user's public projects else all
+      @projects = if @current_user.nil? || @current_user.id != params[:id].to_i
```
tachyons, line 120: `@current_user.id != params[:id].to_i` why are you excluding this? here both triggers same db query

Nitish145: What if some user fetches some other user's projects? Shouldn't we only return public projects then? @tachyons

tachyons: Got it, thanks

tachyons:
Resolve conflicts

Nitish145:
> Resolve conflicts

Done @tachyons

tachyons:
Refer https://github.com/CircuitVerse/CircuitVerse/commit/faea53637ed922aaab8fce39675313924b71b8b8 @current_user won't work anymore

Nitish145:
> Refer https://github.com/CircuitVerse/CircuitVerse/commit/faea53637ed922aaab8fce39675313924b71b8b8 @current_user won't work anymore 

Yeah! I did will be pushing the required changes..

PR #1588, opened 2020-08-08
Code cleanup for Footer, Navbar and some static pages

aayush-05:
Final refactored code cleanup for:
--Navbar (Also added fixes in #1554)
--Footer
--About Us
--Contribute
--Teachers


aayush-05:
ID selector codeclimate issues to be ignored, rest all is fixed

PR #1601, opened 2020-08-14
Update webpacker configuration

tachyons:
1. Remove duplicate jquery import
2. Move jquery ui to webpacker
3. Add canvas2svg to the bundle
4. Add webpack analyser plugin
5. Move webpack dir to simulator


satu0king:
@tachyons, can yo

PR #1602, opened 2020-08-16
Authentication System Revamp - Login/Register setup moved to Popups

aa0501:
image


aayush-05:
Fixed all the major codeclimate issues. The ones involving id selectors (preferred to be used here), use of instance variables in helpers and tests can be ignored

tachyons:
Build is failing

aayush-05:
Yeah, I am not able to figure out why though. Capybara is unable to find the "Log In" link, even when its right there in the Navbar.
image
Can you check once on why this is happening?

app/views/users/sessions/_popup_login.html.erb
```diff
@@ -0,0 +1,59 @@
+<div id="loginModal" class="modal fade" role="dialog">
+  <div class="modal-dialog">
+    <div class="modal-content">
```
nitin10s, line 3: .modal-content require a border: 0 to remove that slight extra white space at the bottom

nitin10s (changes requested):
Extra margin needed at the bottom of the social logins, also the modal looks quite close to the top and could be shifted down

This is not working in sync with simulator.. why so?

The password reset screen looks quite odd, maybe we can remove the illustration and keep the content in centre? Whatsay?

config/routes.rb
```diff
@@ -48,10 +48,9 @@
   resources :featured_circuits, only: %i[index create]
   delete "/featured_circuits", to: "featured_circuits#destroy"
 
-  devise_for :users, controllers: {
-    registrations: "users/registrations", omniauth_callbacks: "users/omniauth_callbacks"
-  }
-
+  get "users/sign_in" => redirect('/')
+  get "users/sign_up" => redirect('/')
```
satu0king, line 52: Does this line need to be repeated twice?

satu0king:
Looks good but the following scenario didn't work very well
1) I log out
2) Go to simulator page
3) Make a circuit and save
4) It just redirects to home page (Not prompting login)

This is step down from the current setup which redirects to a login page.

Me-Priyank:
@Nihal4777  Closing as superseded.

This was fully solved during GSoC 2023 through a better architectural approach:

- #3862 - JWT implementation: RSA-256 signed tokens stored in a cvt cookie (same_site: strict, secure in production) with a 
  custom Warden strategy for API authentication & New API endpoints (/api/v1/me, /api/v1/projects, /api/v1/projects/:id/check_edit_access) backed by JWT auth.
- CircuitVerse/cv-frontend-vue#170 - Simulator wired up to use JWT: simulatorHandler.vue reads the cvt cookie authenticates via API, and saves circuit data to localStorage before any redirect so no work is lost.

A Bootstrap modal wrapping a Devise HTML form (what this PR proposed) is no longer the right approach, the simulator is a decoupled Vue SPA that talks to Rails exclusively via JWT-authenticated API calls, never via HTML forms.

PR #1603, opened 2020-08-17
Add GDPR Confirmation Modal

aayush-05:
Partial PR for making the platform GDPR compliance. Takes User's confirmation after he sign in/sign up (the platform cannot be used without agreeing to the Privacy Policy)
image


nitin10s:
The design part looks fine to me

PR #1622, opened 2020-08-24
fixed bug of referenced .bundle and node_modules folder

darkshredder:
Fixes # fixed bug of referenced .bundle and node_modules folder

changed docker-compose file
earlier we needed to run yarn install and bundle install --without production in our folder before using docker-compose up as these folders were refrenced and were not in repo
but now I have removed referencing of these folders and it will get data from docker container itself
thus our yarn install --checkfiles error is removed completely

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.1'
+version: "3.1"
 services:
   db:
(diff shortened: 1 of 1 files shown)


darkshredder:
@tachyons

PR #1524, opened 2020-07-10
Finalise home page

aayush-05:
**Screenshots**
1.
image
2.
image
3.
image
4.
image
5.
image
6.
image
7.
image
8.
image

PR #1625, opened 2020-08-26
Add GDPR check confirmation 

aayush-05:
Copy of #1603 rebased with the revamped authentication system (from the popup one)


satu0king:
@tachyons I think its best if you review this PR

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,19 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def resource_name
```
satu0king, line 4: ApplicationHelper seems a general class. But this is something to do with user right? I think its better to make a separate class.

app/views/users/registrations/update.js.erb
```diff
@@ -0,0 +1 @@
+location.reload();
```
satu0king, line 1: What is this for?

satu0king:
The UI looks good to me, there is one minor issue, the user cannot log out if he/she doesn't accept the privacy policy (Because the UI doesn't allow closing the privacy policy box). However, this shouldn't be a problem as we can expect all users to accept it)

app/views/users/registrations/update.js.erb
aayush-05: Once the form (GDPR checkbox) gets submitted via AJAX request; the page needs to reload inorder to let user access the platform.

db/migrate/20200821091915_add_gdpr_confirmation_check_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddGdprConfirmationCheckToUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :privacy_confirmation, :boolean, default: false
```
tachyons, line 3: Since this is a boolean field, make attribute name more explicit, like 

`accepted_privacy_policy` so that method invocations will be cleaner

app/helpers/application_helper.rb
tachyons: Methods added here will be available for entire view classes, so generic methods like resource_name shouldn't be overrides here

aayush-05:
@satu0king @tachyons made the changes.
Also, to be merged after #1524

aayush-05:
Also, this doesn't work on Simulator screen as of now. Just requires addition of 7-9 lines for that (haven't implemented to avoid merge conflicts)
@abstrekt maybe you can implement that on the branch you are working; I can help

satu0king (approved):
LGTM, @tachyons please review once

app/helpers/layouts_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+  def resource_name
+    :user
+  end
+
+  def resource
+    @resource ||= User.new
+  end
+
+  def devise_mapping
+    @devise_mapping ||= Devise.mappings[:user]
```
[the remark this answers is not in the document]
tachyons: This seems outdated code, what was the motivation for adding this ?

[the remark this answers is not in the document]
aayush-05: This is to enable access to devise's user model reference in layout views.
image

satu0king:
> Also, this doesn't work on Simulator screen as of now. Just requires addition of 7-9 lines for that (haven't implemented to avoid merge conflicts)
> @abstrekt maybe you can implement that on the branch you are working; I can help

@aayush-05 can we do this now so we can merge this soon!

aayush-05:
Gimme this week, will finalise this

aayush-05:
Moved to #2058

PR #1638, opened 2020-09-08
Improve Search Page | Fix Image loading in Mailer

aayush-05:
image


nitin10s:
Looks super cool now, Should we also increase some margins within the card?

aayush-05:
Not sure, since a single card is already pretty long. You suggest

nitin10s (approved):

PR #1653, opened 2020-09-15
Update to Login page

Dhanushreddy09:
PR 2
Fixes #

Converted "sign up" into a button in login page and removed unnecessary "sign up" link in forgot password page.


aayush-05:
The Signup link has been displayed as a hyperlink to signify secondary importance. Although not sure if we should keep the Signup link on Forgot Password page or not @nitin10s ?

satu0king:
> The Signup link has been displayed as a hyperlink to signify secondary importance. Although not sure if we should keep the Signup link on Forgot Password page or not @nitin10s ?

Yes, please remove the sign up link in forgot password page. But keep the sign up as a hyperlink in the login page.

nitin10s:
Yeah, signup link doesn't make sense on the Forgot Password page [ I don't think we had it in the prototype as well, I guess this got carried from the previous design. ]

Dhanushreddy09:
Now, I made "sign up" as a hyperlink back @nitin10s @satu0king

app/views/users/sessions/new.html.erb
```diff
@@ -46,7 +46,7 @@
           <div class="input-group">
             <%= f.submit "Log in" ,class:"btn primary-button users-login-primary-button" %>
             <div class="users-text-container">
-              <span class="users-text">New User?</span>
```
aayush-05, line 49: "New User" seems fine; we don't have to explicitly mention CircuitVerse. Goes without saying

Dhanushreddy09: Once,Please also check  the teacher's doc file(another change in this commit) .It would be better if I can change both of them in one go.Thanks!);

Dhanushreddy09:
I removed "new user" link in forgot password page @aayush-05

Dhanushreddy09:
Oh sorry ,I haven't noticed my code is outdated. LOL

aayush-05 (approved):
LGTM. The failed tests is not due to this PR mp

PR #1665, opened 2020-09-19
Configure Devise confirmable

satu0king:
Adds confirmable to devise

**Changes in this Pr:**
1. Sidekiq Concurrency to 3
2. Configured Sidekiq for devise mails
3. Added confirmable to user (devise model)
4. Oauth users get confirmed automatically
5. Added resend confirmation option
6. Change default subscribed to false
7. All subscriptions set to false
8. When user acc is confirmed, it is automatically subscribed as well. 

**Migration details:**
1. all user get email notification asking to confirm email
2. giving buffer of 20 days to use account with email confirmation (temporary)


satu0king:
Not sure why tests are failing, 5 tests are failing on master. One new test `Api::V1::AuthenticationController#forgot_password forgot password when user who forgot his/her password exists returns status 200 and should send reset password instructions - spec.requests.api.v1.authentication_controller.forgot_password_spec` is failing I think because devise is now using sidekiq for mails.

satu0king:
Update: One test failing due to async nature of mails, @tachyons please advise. 

image

tachyons (approved):

shubham21155102:
@tachyons 
Can I work on this?
Because I can see this problem is still pending

PR #1666, opened 2020-09-19
Configure Recaptcha

satu0king:
Add's Google Recaptcha for signup and login pages. Configured Recaptcha V2

1. Added flipper configuration
1. Added default flipper variables

Documentation: https://github.com/heartcombo/devise/wiki/How-To:-Use-Recaptcha-with-Devise

image

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 require "flipper/adapters/redis"
+
+# default flipper configuration
+default_flipper_features = {
+  'recaptcha': false,
+  'forum': false,
+  'project_comments': true
+}
+
 Flipper.configure do |config|
   config.default do
     if Rails.env.test?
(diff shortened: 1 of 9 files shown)


aayush-05:
Should we have it on the Login page too? Feels like a strong barrier

satu0king:
> Should we have it on the Login page too? Feels like a strong barrier

I thought about it too. But I felt we should still have it. Note that those who use oauth will not need to do the captcha. @tachyons any inputs on this?

config/initializers/flipper.rb
```diff
@@ -15,7 +24,19 @@
   end
 end
 
-Flipper.enable(:project_comments)
+# set default flipper configuration
+if !Rails.env.test? then
+  client = Redis.new
+  enabled_features = client.smembers("flipper_features")
```
tachyons, line 30: `Flipper.features`

satu0king: Ah I was looking for `Flipper.features`, for some reason I couldn't find it in the documentation easily.

config/initializers/flipper.rb
```diff
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 
 require "flipper/adapters/redis"
+
+# default flipper configuration
+default_flipper_features = {
+  'recaptcha': false,
+  'forum': false,
+  'post_comments': false,
+  'project_comments': true
```
tachyons, line 10: what is the difference between these two ?

tachyons:
@aayush-05 We added invisible captcha to improve the user experience, We are adding re captcha with feature flag so that we can enable/disable this feature easily. Also makes the feature optional for custom installations. Captcha is sometimes a necessary evil

config/initializers/flipper.rb
satu0king: tbh, I don't know, I just put it because https://circuitverse.org/flipper has both of them. Only `project_comments` is used.

PR #1679, opened 2020-09-21
Show element description instead of classname

wallento:
To make it more clear, add a description to each element (dict instead of list). Display this description instead of the element's classname.

image

### DIFF
--- a/public/js/logix.js
+++ b/public/js/logix.js
@@ -56,10 +56,10 @@ function setupElementLists() {
     window.renderOrder = [...(moduleList.slice().reverse()), "wires", "allNodes"]; // Order of render
 
 
-    function createIcon(element) {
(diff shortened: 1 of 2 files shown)


satu0king:
Thank you! We need this change. We are currently transitioning to the new UI deployed on beta.circuitverse.org. 
Can you redirect your PR to new_UI branch? the source files are in public/src. Thanks!

satu0king (changes requested):
Thank you! We need this change. We are currently transitioning to the new UI deployed on beta.circuitverse.org.
Can you redirect your PR to new_UI branch? the source files are in public/src. Thanks!

wallento:
Sorry, missed that one. Is that still relevant?

PR #1697, opened 2020-09-23
[FIXED] After search results of user the category of search should still remain 'Users' and not projects

### LINKED ISSUE #1667: After search results of user the category of search should still remain 'Users' and not projects
**Describe the bug**
when a user is searched it shows result of  all registered users but category of search automatically changes to projects  

**To Reproduce**
1. Go to 'www.circuitverse.org'
2. Click on 'search tab and search for user'
3. after search of user the  category shows projects as category and not users

**Expected behavior**
  If a user is searched, after search results category of search should remain users only and not projects
 and if projects are searched then category of search should still remain projects

**Screenshots**

as shown in screenshot we have results of user search  but category of search gets changed to projects automatically 

Screenshot from 2020-09-19 17-25-28

No

DarthCucumber:
Fixes #1667 

Added a script which retains the search filter applied to the search from **_select tag_**.

before:
if  **Users** filter is selected from **options** (_select tag_) and a name is searched, the **option** (_select tag_)  in the results page reverts to default value rather than staying **Users**.
bug

after:
after fixing it looks like so:
fix

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -20,7 +20,6 @@
       <div class="col-xs-12 col-sm-12 col-md-4 col-lg-4 footer-column">
         <div class="row text-left">
           <div class="col-5 footer-links">
-            ​​<h5><span style="background-color:#42b983; padding:0px; padding-left: 2px; padding-right:2px;">Circuits</span></h5>
             <a href="/simulator"><h6>Simulator</h6></a>
             <a href="/learn" target="_blank"><h6>Learn</h6></a>
             <a href="/examples"><h6>Examples</h6></a>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_footer.html.erb
```diff
@@ -31,7 +30,6 @@
             <% end %>
           </div>
           <div class="col-6 footer-links">
-            <h6>Info</h6>
```
tachyons, line 34: Unrelated

app/views/layouts/_header.html.erb
```diff
@@ -90,4 +90,20 @@
   if (window.location.href.includes('search')) {
       searchBar.style.display = "block";
   }
+
+    /* for retaining select tag search query after a search*/
+  if(window.location.pathname==="/search"){
```
tachyons, line 95: This seems more like a workaround than a fix, ideally properly configured form should be handling this scenario itself

DarthCucumber: Okay, I'll make the changes accordingly and make another pull request. Thanks for the feedback.

app/views/layouts/_footer.html.erb
DarthCucumber: sorry about this. 
This is the previous commit. somehow it got attached. I'll fix it.

DarthCucumber: @tachyons I made changes to the form and this time it's not a workaround. When I'm making a pull request the above commits I made previously are also adding up to the current one. Can you please help me.

Issue #1651, opened 2020-09-15
Combinational Analysis tool fails due to an independent output


nTrfAce:
**Describe the bug**
If the Combinational Analysis tool is asked to create a circuit where one or more of the outputs are not dependent on an input, it outputs nothing or a partial circuit.

**To Reproduce**
I encountered this behavior trying to generate the circuit for a 7-segment decoder, but limited to the states 0-3.  The issue is that the b segment in this case is "degenerate" per Issue #659 - the output is not dependent on the input.  Yes, we could expect the user to notice that, but it is also possible that someone just learning Digital Logic would be confused as to why a 7-segment decoder would work if configured for 0-7 or 0-F, but fail when configured for 0-3.

Steps to reproduce the behavior: (simple case)
1. Go to Tools > Combinational Analysis
2. set the input to "i", and output to "o", and click Next
3. configure the output as shown
```
 i | o
---+---
 0 | 1 
 1 | 1 
```

Steps to reproduce the behavior: (actual case)
1. Go to Tools > Combinational Analysis
2. set the inputs to "i, j", and outputs to "a,b,c,d,e,f,g", put a check for the decimal column, click Next
3. configure the outputs as shown
```
dec | i | j | a | b | c | d | e | f | g
----+---+---+---+---+---+---+---+---+---
  0 | 0 | 0 | 1 | 1 | 1 | 1 | 1 | 1 | 0
  1 | 0 | 1 | 0 | 1 | 1 | 0 | 0 | 0 | 0
  2 | 1 | 0 | 1 | 1 | 0 | 1 | 1 | 0 | 1
  3 | 1 | 1 | 1 | 1 | 1 | 1 | 0 | 0 | 1
```
4. click Next

**Expected behavior**
I expect that the Combinational Analysis tool will generate a valid circuit for the dependent outputs even in the presence of independent outputs, and warn the user of the presence of any independent outputs.

I am not working on the issue.

gr455:
Hello!, I would like to take up this issue

PR #1703, opened 2020-09-25
Fix/combinational analysis

satu0king:
Fixes #1678, #1651

The error was that quinMcCluskey was unable to handle the case of all ones, all zeroes and/or all don't cares. 
This PR fixes the bug by adding a constant val to one of the inputs. quinMcCluskey is used only when required. 
@gr455 @Shivansh2407, please test this PR. 


Shivansh2407:
Yeah sure, Will test this out and get back to you on this @satu0king 👍

gr455 (approved):
LGTM!

PR #1584, opened 2020-08-03
UI fixes

abstrekt:
to the new UI

PR #1705, opened 2020-09-26
Ui fix

satu0king:
Accidentally closed #1584 by mistake. Raising new PR.


public/src/ux.js
```diff
@@ -369,6 +369,7 @@ $('#bitconverterprompt').append(`
  */
 $('#bitconverter').click(() => {
     $('#bitconverterprompt').dialog({
+        resizable:false,
```
satu0king, line 372: Why resizable false, it is a good thing right?

abstrekt: I disabled resizable for all dialogs, resizing dialogs seemed to break to inner html making it look bit odd,... anyways disablling it doesn't seem to have any negative results on the usability

PR #1718, opened 2020-09-27
Fix Priority Encoder mislabeled EN to V "Valid" output.

### LINKED ISSUE #1702: Priorty Encoder Enable input is really a Valid output
**Describe the bug**
The documentation for the Priority Encoder https://docs.circuitverse.org/#/decodersandplexers?id=priority-encoder says "AN enable input is also provided to activate/deactivate the priority encoder."  The connection at the bottom of the symbol is labeled EN.

However, this is not an Enable input, rather it is a output which indicates if the output is Valid.  The Valid output goes high when there is at least one input which is high.  If the Valid output were not present it would be ambiguous if there was no input to the Priority Encoder, or if the lowest priority input was asserted.  Further the Valid output is very useful, as this circuit element can take an arbitrary number of interrupts in, and consolidate them to a single interrupt out.

https://en.wikipedia.org/wiki/Priority_encoder describes this as "The output V indicates if the input is valid."

**To Reproduce**
If you wire an input circuit element to EN and set it to 0, and an input circuit element to any of the Priority Encoder's inputs (0, 1, ...) and drive it to 1, you get "Contention Error: 0 and 1" and "Contention Error: 1 and 0".

**Expected behavior**
The circuit element is working properly, I think this is a documentation bug, that EN should be re-labeled to V, and the documentation updated, adding a V column to the table, a row to the table showing no input asserted, and the help page text "AN enable input is also provided to activate/deactivate the priority encoder." changed to "A 'Valid' output labeled V is provided to indicate that at least one input is 1.  Valid 0 means no valid input is present on any of the inputs, IE all inputs are 0.  Valid 1 means at least one input is 1.  Read the binary value of the output(s) to determine the input number of the highest input which is active.  An example use of this circuit element is to indicate if any interrupt is present to a system, and to provide the value of the highest priority interrupt which is asserted."

**Screenshots**
Here is a project which makes use of the mislabeled but functional "Valid" output signal:
image
https://circuitverse.org/users/37662/projects/priority-encoder-to-hex-7-segment

Here is an image showing a minimal circuit to reproduce the contention error if it were an Enable input:
priority encoder contention issue

No

nTrfAce:
Fixes #1702 

In the Priority Encoder circuit element changed EN to V, and in the module's code changed the word enable to the word valid.  The module already correctly had a "Valid" output, but it was mislabeled.

See also https://github.com/CircuitVerse/CircuitVerseDocs/pull/276

### DIFF
--- a/public/js/modules.js
+++ b/public/js/modules.js
@@ -3363,7 +3363,7 @@ function PriorityEncoder(x, y, scope = globalScope, dir = "RIGHT", bitWidth = 1)
         this.output1.push(a);
     }
 
-    this.enable = new Node(10, 20 + this.inp1[this.inputSize - 1].y, 1, this, 1);
+    this.valid = new Node(10, 20 + this.inp1[this.inputSize - 1].y, 1, this, 1);
 
 
 }
@@ -3377,7 +3377,7 @@ PriorityEncoder.prototype.customSave = function () {
         nodes: {
             inp1: this.inp1.map(findNode),
             output1: this.output1.map(findNode),
-            enable: findNode(this.enable)
+            valid: findNode(this.valid)
         },
         constructorParamaters: [this.direction, this.bitWidth],
     }
@@ -3407,11 +3407,11 @@ PriorityEncoder.prototype.resolve = function () {
     temp = out;
 
     if (out.length != undefined) {
-        this.enable.value = 1;
+        this.valid.value = 1;
     } else {
-        this.enable.value = 0;
+        this.valid.value = 0;
     }
-    simulationArea.simulationQueue.add(this.enable);
+    simulationArea.simulationQueue.add(this.valid);
 
     if (temp.length == undefined) {
         temp = "0";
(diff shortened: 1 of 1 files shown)


public/js/modules.js
```diff
@@ -3457,7 +3457,7 @@ PriorityEncoder.prototype.customDraw = function () {
     for (var i = 0; i < this.bitWidth; i++) {
         fillText(ctx, String(i), xx + this.output1[0].x - 10, yy + this.output1[i].y + 2, 10);
     }
-    fillText(ctx, "EN", xx + this.enable.x, yy + this.enable.y - 5, 10);
```
satu0king, line 3460: Ahh, we cannot rename `this.enable` actually as old circuits will stop working as they expect it to be `this.enable`. However, you can change the fillText() alone.

satu0king (changes requested):

satu0king:
Closing this issue in favor or fixing it in new ui

PR #1737, opened 2020-09-30
Fix/module setup

satu0king:
Removed dependency on modules.js on everything. 
Cyclic dependency is still not removed however. More work needs to be done.


satu0king:
@shubhankarsharma00 please review this.

shubhankarsharma00 (approved):
LGTM Let's test this in production and see if the bug is removed.

PR #1741, opened 2020-09-30
Fixes ui

### LINKED ISSUE #1661: drag out of bound
Dragging of elements should not go out of bound in PR #1585, may take panels to a point of no return.

image

*Creating this issue to keep track of the bug*

abstrekt:
Fixes #1661 , 

#1705 can be closed with this PR.


abstrekt:
@satu0king PR has the changes of #1705

satu0king:
chrome-capture

The nav bar moves on hover -> very ugly. 

This PR:
image

Current:
image

I personally like the current version without the white boxes. Seems more modern. @Shivansh2407 @nitin10s @sal2701 any inputs here?

nitin10s:
The white boxes are used because the white underline was not appropriately signifying am input field at various other sections. So, to give a clear indication of input fields I suggested this change. I strongly feel we should go with boxes.

Shivansh2407:
We can also do like we are doing in the Combinational Analysis Feature, that is Outlining the text box onFocus of the Text Field. When it is out of Focus, we can let it remain as it is now. Let me know your views.

nitin10s:
No, the point of making them boxes is to help users understand that there is an input field to click in the first place.

satu0king:
> No, the point of making them boxes is to help users understand that there is an input field to click in the first place.

I am fine with the box but it is looking a bit glaring on the eyes. How about adding rounded corners, reducing the border thickness and opting for a grey color instead of pure white?

nitin10s:
Yeah, adding rounder borders and reducing border thickness can be done. Don't think grey would be a good colour because it usually signifies disabled fields.

satu0king:
image

image

@nitin10s

satu0king:
chrome-capture

Instead of bold, can we do underline? It seems to work neatly. @nitin10s

nitin10s:
Looks surely better than the previous one, what about hovering changes the colour from white to green?

nitin10s:
> image
> 
> image
> 
> @nitin10s @abstrekt

Border thickness looks very less 😅

satu0king:
> Border thickness looks very less 😅. Maybe let's not change it?

I actually increased the border thickness from 0.5px to 1px. This is how it looks in 2 px.
image

@nitin10s test it locally once, it looks fine for me

sal2701:
In the navbar, can we have the arrow not vanish?
We can have a simple arrow pointing down by default and pointing up when drop-down open. And underline when hovering?

satu0king:
> In the navbar, can we have the arrow not vanish?
> We can have a simple arrow pointing down by default and pointing up when drop-down open.
> 
> And while hovering underline looks good, or we could have a arrow point left on hover. I'll try doing it.
> 
> Thoughts?

I was actually thinking the same but thought this was intended design.

nitin10s:
No, it was not intended, we, in fact, found this issue but it eventually got overlooked.

PR #1758, opened 2020-10-02
Added a down caret to getting started link

### LINKED ISSUE #1655: Indicate that Getting Started [in Navbar] is a dropdown
**Describe the bug**
The Getting Started nav-link in the navbar is a dropdown but the same isn't indicated.
image

**Expected behavior**
A down arrow to indicate that it is, in fact, a dropdown 

No

HADES-01:
Fixes #1655 

Added a downward caret to getting started link in the header to indicate that it is a dropdown.

### DIFF
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -62,7 +62,7 @@
   }
 }
 
-.dropdown-toggle:nth-child(2) {
+.dropdown-toggle.user-dropdown {
   cursor: pointer;
 
   &::after {
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -62,7 +62,7 @@
   }
 }
 
-.dropdown-toggle {
+.dropdown-toggle:nth-child(2) {
   cursor: pointer;
 
   &::after {
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -62,7 +62,7 @@
   }
 }
 
-.dropdown-toggle.user-dropdown {
+.dropdown-toggle.navbar-user-dropdown {
   cursor: pointer;
 
   &::after {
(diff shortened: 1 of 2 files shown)


satu0king:
Please add screenshots

aayush-05:
@HADES-01 It must have been hidden somewhere in the classes defined there. Can you check and rather remove that additional css property?

HADES-01:
@aayush-05 I think I've found the faulty class which is `navbar-dropdown-toggle-hidden` in the parent element. Removing it brings the caret back but its position is all messed up like this 

image

I've found the cause of this fault as well, It is happening because the  `dropdown-toggle::after` has been override in the application's CSS because otherwise the caret adjacent to the name of logged-in user gets messed up. Should I make both changes ?

aayush-05:
@HADES-01 limiting overriding css to only the username dropdown caret should do

HADES-01:
> @HADES-01 limiting overriding css to only the username dropdown caret should do

@aayush-05 I already did that and it works just fine.

aayush-05 (approved):
LGTM @satu0king . Please update the screenshots too

HADES-01:
@aayush-05 First solution was not working now but I have fixed it.

**Screenshot**
image

app/assets/stylesheets/navbar.scss
```diff
@@ -62,7 +62,7 @@
   }
 }
 
-.dropdown-toggle {
+.dropdown-toggle.user-dropdown {
```
aayush-05, line 65: Use sass syntax. The "user-dropdown" class should be a subobject. Also prefix this class with "navbar"

HADES-01: @aayush-05 I prefixed the class name with navbar but I don't think it nesting applies here because they are same level classes.

aayush-05: Got it! My bad

aayush-05 (approved):
LGTM @tachyons

PR #1775, opened 2020-10-03
removing reduntant files + some fixes

abstrekt:
* Removed files that were no longer in use.
* Small fix to custom radios.
* reduced 1 color variable.
* Changes circuit del icon from ```x``` to trash.

image


abstrekt:
@nitin10s @satu0king please review

satu0king:
@abstrekt why are these files no longer in use? is it because we have added them inline in the html?

abstrekt:
> 
> 
> @abstrekt why are these files no longer in use? is it because we have added them inline in the html?

yes the svgs were added inline as to support color themes,

satu0king:
@abstrekt we should try and find a way to not include this inline and support color themes.

abstrekt:
> @abstrekt we should try and find a way to not include this inline and support color themes. Lets keep the files for now, incase we find a way to do so in the future.

okay sure, I'll rebase, there's one plugin that is no longer being used, that surely needs remove.

satu0king:
@abstrekt can u fix this PR?

tachyons:
@abstrekt Is this PR still relevant?

PR #1778, opened 2020-10-03
checkbox fix#1673

### LINKED ISSUE #1763: Complete div behaves as checkbox rather than just the checkbox on Login Page
**Describe the bug**
Clicking the whole div selects/deselects the checkbox.

**To Reproduce**
1. Go to [CircuitVerse Login Page](https://circuitverse.org/users/sign_in)
2. Click on and around the 'Remember Me' Checkbox
3. It selects/deselects the Remember Me Checkbox

**Expected behavior**
Only clicking the checkbox should select/deselect it.

**Screenshots**
image

Nope.

pavanjoshi914:
Fixes #1763 

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -257,7 +257,7 @@
   font-size: 22px;
   left: 7rem;
   position: relative;
-  
+
   input {
     height: 0;
     opacity: 0;
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -251,13 +251,13 @@
 }
 
 .primary-checkpoint-container {
+  bottom: 1rem;
   cursor: pointer;
   display: inline-block;
   font-size: 22px;
-  position: relative;
   left: 7rem;
-  bottom: 1rem;
-
+  position: relative;
+  
   input {
     height: 0;
     opacity: 0;
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -252,9 +252,11 @@
 
 .primary-checkpoint-container {
   cursor: pointer;
-  display: block;
+  display: inline-block;
   font-size: 22px;
   position: relative;
+  left: 7rem;
+  bottom: 1rem;
 
   input {
     height: 0;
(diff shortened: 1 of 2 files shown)


satu0king:
Looks good to me!

satu0king (approved):

satu0king:
@aayush-05 check this one

app/assets/stylesheets/components.scss
```diff
@@ -251,9 +251,11 @@
 }
 
 .primary-checkpoint-container {
+  bottom: 1rem;
   cursor: pointer;
-  display: block;
+  display: inline-block;
   font-size: 22px;
+  left: 7rem;
```
aayush-05, line 258: Why was this required? If you require it to align the checkbox better; please add a separate class to this particular checkbox and move the style there. It can be saved in app/assets/stylesheets/users.scss
Changing this on main checkbox class would break other ones on platform.

app/views/users/sessions/new.html.erb
```diff
@@ -36,7 +36,8 @@
         </div>
         <% if devise_mapping.rememberable? -%>
           <div class="form-group newproject-checkbox">
-            <label class="primary-checkpoint-container users-primary-checkpoint"><h6>Remember Me</h6>
+           <h6>Remember Me</h6>
+            <label class="primary-checkpoint-container users-primary-checkpoint">
```
aayush-05, line 40: This change isn't required. We want the checkbox to be accessible via the Label text.
Rest all LGTM

app/assets/stylesheets/components.scss
pavanjoshi914: because i moved h1 above label somehow checkbox lost its alignment so to make it align done this

app/views/users/sessions/new.html.erb
pavanjoshi914: It was mentioned in the issue that checkbox should be accessed only through checkbox so according to mentions in issue 
I excluded h1
if  we allow then only change we need is display:inline-block

aayush-05: Will confirm this and then we can move ahead

satu0king: > Will confirm this and then we can move ahead

Any update?

aayush-05: Having it accessible from the label text is a good option. Let us go with that @pavanjoshi914 . The fix you mentioned would do

pavanjoshi914: Ohk implementing following changes

pavanjoshi914:
after discussion checkbox will we accessible through the label... change made in #2029

PR #1779, opened 2020-10-03
Docker: Start webpack dev server with web container

gr455:
Starts webpack-dev-server when the web container boots instead of having to start it manually from within the container


satu0king:
Will merge for now but I think the webpack dev server should ideally be done in a separate container.

Issue #1708, opened 2020-09-26
Zoom in and Zoom out buttons


satu0king:
The Zoom slider in the new UI at [beta.circuitverse.org](beta.circuitverse.org) works but the + and - symbols are not buttons. It would be nice if they act as buttons for zooming in and zooming out respectively. 

Branch: new_UI
JS: public/src
HTML: app/views/simulator/edit.html.erb

The class name for the zoom slider is `zoom-slider`

**Are you working on this? (Yes/No)**
No

![image](https://user-images.githubusercontent.com/2092958/94338795-a86c1900-0012-11eb-895e-8eab57103b14.png)

arjungarg07:
Hey @satu0king Can i work upon this feature?

satu0king:
@arjungarg07 go ahead

rohan2734:
@satu0king  if @arjungarg07 couldnt do it, then assign me otherwise i will look on other issue.

satu0king:
@arjungarg07 working on this?

arjungarg07:
Yeah! , @rohan2734  can work on it, I am solving other issues.Thanks!

rohan2734:
@satu0king @arjungarg07 ok i am working on this.

rohan2734:
```
<script type="text/javascript">
  function zoomIn() {
      console.log("clicked zoom in");
      var canvasLayout = document.getElementById("backgroundArea");
      var currLeft = canvasLayout.style.left;
      var currTop = canvasLayout.style.top;
      canvasLayout.style.left = (currLeft - 10)+"px";
      canvasLayout.style.top = (currTop - 10)+"px";
  }
  function zoomOut() {
      console.log("clicked zoom out");
      var canvasLayout = document.getElementById("backgroundArea");
      var currLeft = canvasLayout.style.left;
      var currTop = canvasLayout.style.top;
      canvasLayout.style.left = (canvasLayout -20)+"px";
      canvasLayout.style.top = (canvasLayout -20)+"px";
  }
</script>
```

```
<canvas
      id="backgroundArea"
      style="
        position: absolute;
        left: 0;
        top: 0;
        z-index: 0;
        width: 100%;
        height: 100%;
      "
    ></canvas>
```

```
<div class="zoom-slider">
            <button type="button" onclick="zoomOut()">
                 <i class="fa fa-minus" aria-hidden="true" ></i>
            </button>
[...]

PR #1793, opened 2020-10-06
added zoom funcitonality, but not working,i need help

rohan2734:
Fixes #
for issue #1708 
- I have added the zoom functionality ,that is added a temporary zoom "-" and zoom "+" buttons, 
- styling is not done properly now, they are in black colour now, 
- I will make it better when the implementation is done
- please check the current changes, i need help, 
- the top and left of the canvas,where id="backgroundArea" before implementation are ```top:0``` and ```left:0```
-but after this functionality they are initliased with ```top:-10`px`` and  ```left:-10px```, 
- expected behaviour is  to only change when the buttons are clicked

zoom_button_top_and_left_initialized_as_ten

### DIFF
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -7,9 +7,33 @@
   crossorigin="anonymous"
 ></script>
 -->
-<!-- 
-<link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css"> 
+<!--
+<link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
 -->
+<script type="text/javascript">
+  function zoomIn() {
+      console.log("clicked zoom in");
+      var canvasLayout = document.getElementById("backgroundArea");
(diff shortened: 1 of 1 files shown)


abstrekt (commented):
Not a review for the code, but this doesn't even work?

rohan2734:
> Not a review for the code, but this doesn't even work?

ya i am very close to it, actually when we click on th buttons, the console is logging correctly, but the value is not being updated, and a error of get request is poppin g some times and sometimes not, also the top and let value is initially zero before i changed the code, but it is now 10, so it is showing some wierd behaviour. 

can you help?

satu0king:
@rohan2734, you cannot add any code directly in the `edit.html.erb`, check `public/src/data.js` and `public/src/ux.js`

rohan2734:
> @rohan2734, you cannot add any code directly in the `edit.html.erb`, check `public/src/data.js` and `public/src/ux.js`

yes @satu0king  sir, i will move it to the respective files when, the functionality is done, but i need some help ,here for the implementation, it is 90% done, but 10%  i am not able to do.

PR #1818, opened 2020-10-18
Fix project preview url for friendly ids

tachyons:
Fixes https://github.com/CircuitVerse/CircuitVerse-Presentation-Embed/issues/17


satu0king (approved):
Looks good to me. But do APIs work with slugs generally or project ids?

tachyons:
Depends on the app, since slug is a first class citizen in our app, api must support slug as well

PR #1847, opened 2020-10-27
Improve search speed

tachyons:
Improve search field using stored attribute and database index, requires postgres 12 or higher


tachyons:
Blocked by https://github.com/rails/rails/pull/39368

PR #1914, opened 2020-11-12
boolean function --> generate circuit

### LINKED ISSUE #1706: Boolean function -> Circuit Generation
Currently, we have Combinational Analysis to generate circuits.
We also have ways to generate boolean logic table given a boolean function. See [here](https://learn.circuitverse.org/docs/bool2.html)

We should easily be able to integrate the 2. Given a boolean function like `X = AB' + C`, we should get a minimized circuit for the same. 

Please send PRs to `new_UI` branch. 

No

pavanjoshi914:
closes #1706 
 Integrating functionality to generate circuit's through boolean function

    1>allowing user to generate circuit using combinational analysis or using directly boolean function
    2> solving boolean function and generating a automated truth table
    3> constructing a proper data pipeline to generate circuit through the given boolean function 

generating-circuits-through-bool


public/src/combinationalAnalysis.js
```diff
@@ -296,3 +320,146 @@ function drawCombinationalAnalysis(combinationalData, inputList, outputListNames
...
+                    var style = '<style> table {font: 20px Calibri;} table, th, td {border: solid 1px #DDD;border-collapse: collapse;} padding: 2px 3px;text-align: center;} </style>';
+                    var win = window.open('', '', 'height=700,width=700');
+                    win.document.write('<html><head>');
+                    win.document.write('<title>Boolean Logic Table</title>');
+                    win.document.write(style);
+                    win.document.write('</head>');
+                    win.document.write('<body>');
+                    win.document.write(`<center>${sTable}</center>`);
+                    win.document.write('</body></html>');
+                    win.document.close();
```
tachyons, line 459: Since document write can be expensive, can you build the whole string and write to dom ? 

htmlBody += <html><head>
htmlBody += 
htmlBody += 
htmlBody += 
htmlBody += 
document.write(htmlBody)

pavanjoshi914: made changes accordingly pls have a look on it !!

satu0king: Alternatively use multiline string @pavanjoshi914 , it will look better

satu0king (changes requested):
Worked very well in my tests. Please document the code where necesessary, and write a paragraph as to how you are achieving this feature (and put it in the code as well).

public/src/combinationalAnalysis.js
```diff
@@ -296,3 +321,147 @@ function drawCombinationalAnalysis(combinationalData, inputList, outputListNames
...
+                return 1;
+            return '';
+        } catch (e) {
+            return '';
+        }
+    }
+ 
+    $('#combinationalAnalysis').empty();
+    $('#combinationalAnalysis').append(s);
+    $('#combinationalAnalysis').dialog({
```
satu0king, line 414: Are you duplicating the code here? Is it possible to reference the old code itself and just update the boolean table there? or am I missing something here.

pavanjoshi914: during the process what I have understood is this...(correct me if I misunderstood something!!)

we have to take input in export function ```createcombinationalanalysispromt``` and call functions inside it with arguments as input and scope because through scope a object is generated which is needed to be passed to function which contains timestamps and other autogenerated info according to input variables which will help drawcomboanalysis function to generate circuit

so to treat combinational analysis and boolean function as alternate option we have to use different functions and different dialogues are used for both 

above mentioned code  is inside the function ```createbooleanprompt```  in older code  to use same code for append string and create dialogue box 
we have to call older function from new function  and pass the string and use if else before appending not just here many things will conflict in datapipeline also when we generate circuit 
also combinational analysis have different data  
 so just settled up using different for both
string we are appending is toally different 
the only part duplicating is creating combinational data to generate minimized circuit and truth table print part

guide me if any possibilites ? thanks!!

public/src/combinationalAnalysis.js
pavanjoshi914: okay making changes in latest commits !!

pavanjoshi914:
> Worked very well in my tests. Please document the code where necesessary, and write a paragraph as to how you are achieving this feature (and put it in the code as well).

yup!!

public/src/combinationalAnalysis.js
satu0king: Yeah, so I think you can simply modify the older function as follows: 

`function createBooleanPrompt(inputListNames,outputListNames,scope=globalScope, values = [])`

If values is empty, then set everything to 0, otherwise set it to whatever values u pass. Something along these lines. Dies it make sense?

pavanjoshi914: Not sure about this.. but I have a idea 
What we can do is just solve a Boolean expression and return the output in form of array we catch that output array in first function and there only call createbooleanpromt by passing output array and other parameters

As combinational analysis uses 0,x,1 and in Boolean we have to print ouput array we can create a Boolean type  var and pass it to createbooleanpromt and print corresponding truth table using if else 
Remaining part can also be handled using conditional 
I haven't tried it but I will try 
Is this solution ohk with you? @satu0king

satu0king: I didn't follow you but I think I will fix this PR myself, just fix the merge conflict.

pavanjoshi914: yup its hard to pass info in words... i wanted to say something presented in screenshot below

satu0king: I said exactly this `function createBooleanPrompt(inputListNames,outputListNames,scope=globalScope, values = [])` 
I called it `values`, you called it `output`

pavanjoshi914: oops sorry for confusion !!! anyways  so shall i proceed in this way?

satu0king: @pavanjoshi914 can u fix this soon?

pavanjoshi914: yes fixing it now thanks!!

pavanjoshi914:
referencing commits to #1942

pavanjoshi914:
please have a look on it @satu0king

pavanjoshi914:
as new_UI is deployed creating new pr to master #1971

PR #1930, opened 2020-11-18
Image Annotation Feature

satu0king:
The image annotation feature was surprisingly easy to implement!. 

image

image

@pavanjoshi914 I am really loving Imgur, check this out.


pavanjoshi914:
its working great!!! :partying_face: 
we can also think about  handy resizing using corners of image in future commits!!
also it is very usefull for teachers so we can think adding icon in button panel for faster access

PR #1937, opened 2020-11-20
Project Link Fix

satu0king:
Fixes #

@tachyons  please review this asap

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -27,7 +27,7 @@ class AssignmentsController < ApplicationController
     @project.assignment_id = @assignment.id
     @project.project_access_type = "Private"
     @project.save
-    redirect_to user_project_path(current_user, @project)
+    redirect_to user_project_path(@project.author_id, @project)
   end
 
   # GET /assignments/new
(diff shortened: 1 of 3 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -27,7 +27,7 @@ def start
     @project.assignment_id = @assignment.id
     @project.project_access_type = "Private"
     @project.save
-    redirect_to user_project_path(current_user, @project)
+    redirect_to user_project_path(@project.author_id, @project)
```
tachyons, line 30: what is the diffrenece this makes as current_user and project.author are same here ?

satu0king: Just making it consistent across all controllers.

tachyons: Do we know the exact issue ?

satu0king: current_user = @project.author_id if its the logged in user is @project user. But otherwise its not the case. The main issue is `user_project_path(current_user, @project)` can be wrong depending on who the `current_user` is. But ` user_project_path(@project.author_id, @project)` is always correct irrespective of that. 

We should not be having the first parameter in the first place. We should have a helper function which does this.

PR #1965, opened 2020-11-24
wire node color

abstrekt:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN mkdir /circuitverse
 WORKDIR /circuitverse
 
 # install dependencies
-RUN apt-get update -qq && apt-get install -y imagemagick && apt-get clean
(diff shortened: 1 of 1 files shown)


public/src/node.js
```diff
@@ -448,9 +448,9 @@ export default class Node {
             }
         }
         var colorNode = colors['stroke'];
-        const colorNodeConnect = colors['color_wire_con']
-        const colorNodePow = colors['color_wire_pow']
-        const colorNodeLose = colors['color_wire_lose']
```
satu0king, line 453: You just changed the color of all the nodes to one color! I think it looks ugly. We need to give different colors depending on the value. Can you check old ui colors?

satu0king (changes requested):

public/src/node.js
abstrekt: Ok I'll revert back to this

PR #1883, opened 2020-11-06
Added option to directly edit the project title from header

dipanshparmar:
Fixes #1808 

#### Describe the changes you have made in this PR -
1. Added option to directly edit the project title from the header itself.

### Screenshots of the changes (If any) -
**Initial State**
image

**On Change**
(Changes are synced automatically between both of the project title inputs)
image

PR #1968, opened 2020-11-24
Added option to directly edit the project title from header

### LINKED ISSUE #1808: Add feature to directly edit the project title
Right now, a prompt opens up asking for the title, we can improve this experience if we could directly edit the project title by clicking on it.

Edit the project title by directly clicking on the title, it can be implemented by using an input tag in place of the span tag 

**Screenshots**
image

dipanshparmar:
Fixes #1808 

1. Added title input bar in the header so that the title can be changed directly

**Initial State**
image

**On change**
image


dipanshparmar:
@satu0king Please have a look at this one. Upgraded from #1883

public/src/ux.js
```diff
@@ -21,6 +21,23 @@ import { generateImage } from './data/save';
...
+
+    setProjectName($('#projectName').val()); // setting initial value
+    $('#projectName').on('input', () => { // setting project name on change
+        setProjectName($('#projectName').val());
+    })
+
+    watchOneInputChangeAndUpdateAnother('[name="setProjectName"]', '.projectName'); // setting the value of header title on sidebar title change
+    watchOneInputChangeAndUpdateAnother('.projectName', '[name="setProjectName"]'); // setting the value of sidebar title on header title change
+})
+
```
satu0king, line 40: Please put this in a `setup` function. `$(document).ready` should not be used for everything

satu0king (changes requested):

dipanshparmar:
@satu0king I moved the **header title input** code into the setupUI function of the same file. Please review.

public/src/listeners.js
```diff
@@ -43,23 +43,6 @@ export default function startListeners() {
...
     });
 
-    $('#projectName').on('click',() => {
-        simulationArea.lastSelected = globalScope.root;
-        setTimeout(() => {
-            document.getElementById("projname").select();
-        }, 100);
-    });
-    /* Makes tabs reordering possible by making them sortable */
-    $("#tabsBar").sortable({
```
satu0king, line 53: @dipanshparmar why is this code removed?

public/src/ux.js
```diff
@@ -144,6 +144,21 @@ export function setupUI() {
     // var dummyCounter=0;
 
 
+    // Function updates the values of project inputs accordingly
+    const watchOneInputChangeAndUpdateAnother = (objectToListen, objectToChange) => $(`${objectToListen}`).on('input', (e) => { $(`${objectToChange}`).val(e.target.value); });
```
satu0king, line 148: This is good, can you encapsulate all of this code into a function say `setupProjectNameListeners` and just call it here. This will be more readable and maintainable.

satu0king (changes requested):

public/src/listeners.js
dipanshparmar: @satu0king this code was used to handle the click on the header project title.
image
Like when someone clicks on the project title it activates the project input section in the right to enter the project name.
But as this is no longer required because there is title input in the header itself now that's why this code is removed.
***
And I guess I accidentally removed the block of $('#tabsBar").sortable(). I'll restore this one.

public/src/ux.js
dipanshparmar: Sure @satu0king

PR #1980, opened 2020-11-27
Added buttons to control wave height in Timing Diagram

dipanshparmar:
1. Added option to control the wave form height in "Timing Diagram"

image

**Max height**
image

**Min height**
image


dipanshparmar:
@satu0king please have a look.

satu0king:
Good work! I think we need to create a timingDiagram size manager which manages all these variables.

dipanshparmar:
yeah @satu0king! seems to be a good idea.

PR #1997, opened 2020-12-02
group invitation system/adding members to groups via link

### LINKED ISSUE #1922: Group Invite by Link
It is difficult to invite users in large groups in some cases as teachers don't have list of email ids. 

We can have group invite link similar to the one used in whatsapp. 

Url can be of the form `groups/id/invite/<secret token>`

I think there is a rails API to create tokens easily, something along the lines `has_secure_token :auth_token` @tachyons any inputs here? I think this resource might help: https://blog.bigbinary.com/2016/03/23/has-secure-token-to-generate-unique-random-token-in-rails-5.html

- [ ] Add Token to groups Table
- [ ] Existing groups are assigned a token in the previous migration
- [ ] New groups get a token when created
- [ ] Button in group page to copy invite link
- [ ] Button in group page to reset invite link (should we have this option?) 
- [ ] Add method to groups_controller to add current user to the group if the secret token matches

We can also consider deprecating the way we do currently, i.e. using `pending invitations` for email invite. Instead we can send a mail invite with the link. This will also give a way for users to accept group invite rather than forcibly be put into a group (privacy reasons). Another advantage is if user for some reason signs up on personal account instead of college account, it still works. 

No

pavanjoshi914:
close #1922 

       1 Migration file to introduce new column group_token in groups table
       2 New groups get a token when created
       3 Real time group token generation for existing groups
       4 Functionality to add members to group via link
       5 Set up expiration criteria of two weeks for group token
       6 handling events and generating proper responses for url expired, invalid url, member already present in group.
       7 ajax functionality to regenerate token if it is expired  and update dom in real time
       8 UI

 
group-invite-system


config/routes.rb
```diff
@@ -40,6 +40,8 @@
   resources :privacy, only: :index
 
   scope "/groups" do
+    get ":id/invite/:group_token", to: "groups#group_invite", as: "group_invite"
+    post ":id", to: "groups#group_token_check",as: "group_token_check"
```
tachyons, line 44:
```ruby
member do
  get "invite/:token"
  put :generate_token
end

```

app/views/groups/_add_member_modal.html.erb
```diff
@@ -22,13 +22,39 @@
...
+        <div class="collapse" id="collapseExample">
+          <textarea id="embedInviteLink" class="group-invite-link-text" rows="4" cols="35"><%= group_url(@group) %>/invite/<%= @group.group_token %></textarea>
+        </div>
+      </div>
     </div>
   </div>
 </div>
+
+<script>
+   function linkGenerate() {
```
tachyons, line 47: use  rails remote form http://www.korenlc.com/remote-true-in-rails-forms/

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
     end
   end
 
+  def group_token_check
+    @group = Group.where("id=?", params[:id])
```
tachyons, line 22:
```suggestion
    @group = Group.find(params[:id])
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
     end
   end
 
+  def group_token_check
```
tachyons, line 21:
```suggestion
  def generate_token
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
...
+    @group = Group.where("id=?", params[:id])
+    @group.group_token_generate
+    render json: { "updated" => @group }
```
tachyons, line 24:
```suggestion
    render json: @group
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
...
+  end
+
+  def group_invite
+    @group = Group.find(params[:id])
+    @current_user = current_user.id
+    if Group.where("group_token = ? AND updated_at >= ?", params[:group_token], 12.days.ago).exists?
```
tachyons, line 30: Add a rails scope with name with_valid_token

```suggestion
    if Group.with_valid_token.where(token: params[:token]).exists?
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
...
+      if GroupMember.where("group_id = ? AND user_id = ?", @group, @current_user).exists?
```
tachyons, line 31:
```suggestion
      if GroupMember.where(group: group, user:  @current_user).exists?
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
...
+        redirect_to group_path(@group), notice: "Member is already present in the group."
+      else
+        GroupMember.where(group_id: @group, user_id: @current_user).first_or_create
+        redirect_to group_path(@group), notice: "Group member was successfully added."
+      end
+    else
+      if Group.where("group_token = ?", params[:group_token]).exists?
```
tachyons, line 38:
```suggestion
      if Group.where(group_token: params[:group_token]).exists?
```

tachyons (changes requested):

satu0king:
@pavanjoshi914 Looks good, please write some spec as well.

config/routes.rb
pavanjoshi914: using routes as member

app/views/groups/_add_member_modal.html.erb
pavanjoshi914: using remote:true for ajax request

app/controllers/groups_controller.rb
pavanjoshi914: using new syntax

app/controllers/groups_controller.rb
pavanjoshi914: renamed

app/controllers/groups_controller.rb
pavanjoshi914: made changes

app/controllers/groups_controller.rb
pavanjoshi914: add scope with_valid_token

app/controllers/groups_controller.rb
pavanjoshi914: using new syntax

app/controllers/groups_controller.rb
pavanjoshi914: using new syntax

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
     end
   end
 
+  def generate_token
+    @group = Group.where(id: params[:id])
+    @group.group_token_generate
```
tachyons, line 23: Above method will generate token for multiple groups
```suggestion
    @group = Group.find(params[:id])
    @group.regenerate_group_token
```

app/controllers/groups_controller.rb
```diff
@@ -18,6 +18,31 @@ def show
...
+    render json: @group
+  end
+
+  def group_invite
+    @group = Group.find(params[:id])
+    @current_user = current_user.id
```
tachyons, line 29: Why is this needed ?

app/views/groups/_add_member_modal.html.erb
```diff
@@ -22,13 +22,31 @@
...
+        <div class="collapse" id="collapseExample">
+          <textarea id="embedInviteLink" class="group-invite-link-text" rows="4" cols="35"><%= group_url(@group) %>/invite/<%= @group.group_token %></textarea>
+        </div>
+      </div>
     </div>
   </div>
 </div>
+
+<script>
+document.body.addEventListener('ajax:success', function(event) {
```
tachyons, line 47: use js.erb approach https://flatironschool.com/blog/jquery-and-javascript-meet-erb-embedded-ruby

app/controllers/groups_controller.rb
pavanjoshi914: thats from devise gives id of current user to check if it is already present or not it not present add that user id in group member table

tachyons: I mean why can't use current_user directly, why do you need another instance variable for this ?

tachyons:
Good work, code is almost ready for merge, make codeclimate happy as well

tachyons:
configure rubocop in the IDE or texteditor you use so that you can formatting errors before pushing to githhub

app/controllers/groups_controller.rb
pavanjoshi914: truee!!! I thought declaring data at first will make it more readable, I will use directly then!!

app/views/groups/_add_member_modal.html.erb
pavanjoshi914: using js.erb  file to catch response!!

app/controllers/groups_controller.rb
pavanjoshi914: what we are trying to achieve through this change?... this will change token everytime we click on button...

tachyons: In your current approach, it will reset token for multiple groups and is expensive, this change will limit it to just one group

pavanjoshi914: I tested it.. its running modal method is running only on the current group
here is the sql query

pavanjoshi914:
```
  @group = Group.find(params[:id])
    if (@group.updated_at <=12.days.ago or @group.group_token == nil)
      @group.regenerate_group_token
    end
```
instead of modal method i will use this...

pavanjoshi914:
@tachyons I realised one draw back in using updated_at, it will get its timestamp changed when user  edit the name of group..
I was thinking for a new column ```group_token_created_at``` to hold the creation timestamp for token...

tachyons:
> @tachyons I realised one draw back in using updated_at, it will get its timestamp changed when user edit the name of group..
> I was thinking for a new column token_created_at to hold the creation timestamp for token...

That will work, but `token_expires_at` will make it more flexible

pavanjoshi914:
> > @tachyons I realised one draw back in using updated_at, it will get its timestamp changed when user edit the name of group..
> > I was thinking for a new column token_created_at to hold the creation timestamp for token...
> 
> That will work, but `token_expires_at` will make it more flexible
then,  shall proceed for token_expires _at column ?

pavanjoshi914:
@tachyons  I think it ready to merge now please a have review on it.....

app/controllers/groups_controller.rb
```diff
@@ -20,7 +20,10 @@ def show
 
   def generate_token
     @group = Group.find(params[:id])
-    return unless @group.updated_at <= 12.days.ago || @group.group_token.nil?
+    if @group.token_expires_at.nil? || (@group.token_expires_at < Time.zone.now)
+      @group.update(token_expires_at: Time.zone.now + 12.days)
+    end
+    return unless @group.group_token.nil? || (@group.token_expires_at < Time.zone.now)
```
tachyons:
```ruby
 if @group.has_valid_token?
 //
else
  //
end
```

```ruby
class Group
  TOKEN_DURATION = 12.days
  def has_valid_token?
  end
end
```

Current logic is hard to read, follow above approach

pavanjoshi914:
```
  if @group.token_expires_at.present?  && @group.token_expires_at < Time.zone.now
        @group.regenerate_group_token
    elsif @group.has_valid_token?  &&  @group.group_token.nil?
      @group.regenerate_group_token
    end
```
```
 def has_valid_token?
    if token_expires_at.nil? || token_expires_at < Time.zone.now
      update(token_expires_at: Time.zone.now + TOKEN_DURATION)
    end
 end
```
any improvements in this @tachyons ?

tachyons: `has_valid_token?` must be a [pure function](https://en.wikipedia.org/wiki/Pure_function) and hence should not update database values

```ruby
 def has_valid_token?
     token_expires_at.present? && token_expires_at > Time.zone.now
 end

if !@group.has_valid_token?
     @group.regenerate_group_token
end

```

pavanjoshi914: made changes accordingly , and added some specs

pavanjoshi914:
It got merged :partying_face:  thanks @tachyons for assistance throughout the pr learnt a lot new things through the reviews....

tachyons:
Now we need to add this feature to api so that mobile app can make use of this feature

pavanjoshi914:
> Now we need to add this feature to api so that mobile app can make use of this feature

I would surely like to work on this...

PR #2009, opened 2020-12-06
Docker: Dockerize Webpacker

gr455:
Follow up to #1779 

## This pull
- Creates a seperate container for Webpacker on default network

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ services:
     environment:
       REDIS_URL: "redis://redis:6379/0"
       CIRCUITVERSE_USE_SOLR: "false"
+      WEBPACKER_DEV_SERVER_HOST: "webpacker"
   redis:
     image: redis
   webpacker:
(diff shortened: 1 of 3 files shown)


gr455:
@tachyons what do you think about this

tachyons:
https://github.com/rails/webpacker/blob/master/docs/docker.md this is the officially recommended setup

gr455:
> https://github.com/rails/webpacker/blob/master/docs/docker.md this is the officially recommended setup

Thanks for sharing that, this will hopefully save a lot of trouble during installation

docker-compose.yml
```diff
@@ -43,11 +44,12 @@ services:
       - NODE_ENV=development
       - RAILS_ENV=development
       - WEBPACKER_DEV_SERVER_HOST=0.0.0.0
-    command: ./bin/webpack-dev-server
+    entrypoint: /circuitverse/bin/docker_webpacker_run
     volumes:
       - .:/circuitverse
-    ports:
-      - '127.0.0.1:3035:3035'
```
gr455, line 50: Do we need to expose the port?

tachyons: This for assets, so better keep this

gr455: @tachyons is this fine?

satu0king:
Great work @gr455, its working fine on my laptop. I think the local dev setup can be further optimized, will create an issue with some ideas. Merging for now.

PR #2019, opened 2020-12-09
Showing title "Sign up with" on the Sign up page's OAuth options

sarthakm21:
Initially, the sign up page as well as the login page showed "Log In With" on the OAuth options. Changed the Sign up page to display "Sign Up With"

Initially: 
login_initial
signup_initial

Now: 
login_final
signup_final

### DIFF
--- a/app/views/users/registrations/new.html.erb
+++ b/app/views/users/registrations/new.html.erb
@@ -59,7 +59,7 @@
               </div>
             </div>
           </div>
-          <%= render "users/shared/links" %>
+          <%= render partial: 'users/shared/links', locals: {displayText: "Sign Up With"} %>
         </div>
         <% end %>
       </div>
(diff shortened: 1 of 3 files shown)


pavanjoshi914:
Oauth allows user to avail services through third party authentication.. it's not actually a sign up process..
Why signup for oauth ?

sarthakm21:
@pavanjoshi914 I've changed the text which originally said login instead of signup on the signup page.
I think it makes more sense for the text to be "Sign up with" since the user is technically signing up and not logging in.

tachyons:
But it can be confusing as buttons on both page does the same

satu0king:
Shouldn't make any difference. If I remember correctly, both signup and login do the same thing and either will work. Both will ask permission if needed and authenticate. Merging this for now.

Issue #2005, opened 2020-12-03
Create "Humans of CircuitVerse" markdown page


ShreyaPrasad1209:
**Is your feature request related to a problem? Please describe.**
 This feature will show significant contributions of individual members in CircuitVerse so as to appreciate the efforts made by them and motivate others to contribute to open source.

**Describe the solution you'd like**
They can be displayed in following manner:

-    Contribution - Person

This allows people to be grouped into a contribution as generally multiple people work to resolve an issue. It should be written in markdown so that it can be readable on GitHub as well.

**Describe alternatives you've considered**
Currently I'm looking into similiar existing pages, will post here once I get relevant references. You are free to come up with your own ideas.

**Additional context**
None

**Are you working on this? (Yes/No)**
Open for everyone

harshitaaaaaa:
Can you assign this issue to me?

ShreyaPrasad1209:
Sure! @satu0king can you assign this issue to her? Seems like I don't have the required permissions.

Shivansh2407:
@harshitaaaaaa you can start working on the issue, You just need to Comment "Working on it" or something Similar to make it clear to the Admins/Maintainers/Other Contributors that you've taken up this issue. Looking forward to your PR 🚀

harshitaaaaaa:
Yeah, I'm working on it!

harshitaaaaaa:
@ShreyaPrasad1209 I have started making the markdown page. Can you have a look at it whether it looks the same as you expected or do I have to make some changes in it. After that I'll proceed further. I am hereby attaching screenshots, you can have a look at them and please tell me if it needs any changes or improvements.
![image](https://user-images.githubusercontent.com/66556662/101668509-a1cc3a00-3a76-11eb-9c73-75219c91ca1d.png)
![image](https://user-images.githubusercontent.com/66556662/101668547-ad1f6580-3a76-11eb-93f7-b2f6729eaa43.png)

ShreyaPrasad1209:
@harshitaaaaaa this looks good. Some suggestions -

- Give user the option to enter their details manually. 
- Right below Humans of CircuitVerse headline, reserve some space for a introduction paragraph. I will give you the content in some time. Add dummy paragraph for now.

harshitaaaaaa:
@ShreyaPrasad1209 Yeah thankyouuuu. I'll do as directed.

harshitaaaaaa:
[...]

PR #2041, opened 2020-12-15
Contributors page

harshitaaaaaa:
Fixes #
I have added a feature.

I've made an HTML page mentioning the detials of the contributors.

### Comment:
I have added a dummy paragraph right below the title "Humans of CircuitVerse", instead of that some content according to the mentor has to be added.


satu0king:
@harshitaaaaaa please reference #2005 in PR as fixes #2005

satu0king:
Please share screenshots here. Also this has to come on the main app. So either it can go to public folder or better, it has to be added in logix_controller. 

How did you generate the content for each person, I hope you didn't do that manually? Also the exact content will need to be decided. The languages and number of commits are not very interesting. What is more interesting is features implemented and other contributions. Like for you - you designed the humans of circuitverse page. 

@ShreyaPrasad1209 please review this PR

ShreyaPrasad1209:
@harshitaaaaaa few things I would like to clear before you move ahead-
- This page is supposed to follow CRUD (Create, Read, Update, Delete) operations and access will be given only to the contributers so the logic will be - if the user is present as a CircuitVerse contributor only then they can perform CRUD operations otherwise they can't. 
- You also have to check whether the user is logged in or not.
@robin025 you can help @harshitaaaaaa with this

ShreyaPrasad1209 (changes requested):
@harshitaaaaaa few things I would like to clear before you move ahead-
- This page is supposed to follow CRUD (Create, Read, Update, Delete) operations and access will be given only to the contributors so the logic will be - if the user is present as a CircuitVerse contributor only then they can perform CRUD operations otherwise they can't. 
- You also have to check whether the user is logged in or not.
@robin025 you can help @harshitaaaaaa with this

robin025:
Hi @ShreyaPrasad1209 , thanks for your reply. I was referring to #2005 as you mentioned to create a "Humans of CircuitVerse" markdown page.

So here my suggestion request!
To be able to use a markdown page that will show significant contributions of individual members in CircuitVerse so as to appreciate the efforts made by them ,we can use Jekyll theme which is a static website generator that’s designed for building minimal, static blogs and pages to be hosted on GitHub Pages.
With the help of jekyll ,it strips everything down to the bare minimum, eliminating a lot of complexity
1.We can easily transforms our plain text into static websites pages and blogs.
2.It requires no databases,cms or any updates
3.It supports static HTML, CSS javascript and even Markdown, so with the help of markdown we can easily use all types of labels like you mentioned in #2005
4.What’s best is it goes well with Github-Pages
5.It loads faster

Also to be able to create a CRUD functionality ,i don't think so this can happen in static website but we can create layouts for contributors in markdown, after that contributors just have to edit those markdown files and commit changes.

looking forward to your reply
thanks

ShreyaPrasad1209:
@robin025 we can discuss the implementation details here

robin025:
> @robin025 we can discuss the implementation details here


@ShreyaPrasad1209 i had a question in my mind ,like is this feature is only for contributors of Circuit Verse or for all the users

robin025:
> @robin025 we can discuss the implementation details here

hi @ShreyaPrasad1209  if this [suggestion request](https://github.com/CircuitVerse/CircuitVerse/issues/2005#issuecomment-750414284) is fine with u I can complete this task with jekyll theme, but then dynamic functionality wont be possible.

PR #2044, opened 2020-12-16
Fix bitwidth propagation in Tunnel

### LINKED ISSUE #2028: Bitwidth conflict in tunnels
**Describe the bug**
If the bitwidth of paired tunnels is not manually set to the same value, then one of the tunnels may get value greater than it's max value (2^n - 1). If this tunnel's node is connected to say an output node, then it causes a bitwidth conflict and raises an uncaught exception.

**Proposed solution**
 Sync the bitwidths of paired tunnels to the same value everytime.

**To Reproduce**
1. Go to 'circuitverse.org/simulator'
2. Make a circuit similar to the one in the screenshot
3. Connect the output to the tunnel.

**Expected behavior**
Either the output tunnel should get the same bitwidth as the input tunnel
or a bitwitdth error should be raised

**Screenshots**
Screenshot from 2020-12-11 21-12-27

Here the bitwidth of the upper tunnel is 2 and that of the lower one is 1, connecting the output here causes an exception.

**EDIT:** The exceptions appear to be not due to bitwidth itself. They are raised along with BitWidth Error if the bitwidths don't match( of output tunnel and output element ) too.
The value of the tunnel doesnt seem to propagate properly
Screenshot from 2020-12-16 15-11-57

Here, the node pointed by the cursor has 'X' value whereas the tunnel is outputing 3 ( no bitwidth problem here )

gr455:
Fixes #2028 

- Fixes bitwidth propagation error
- Syncs bitwidths of paired tunnels

### DIFF
--- a/public/src/modules/Tunnel.js
+++ b/public/src/modules/Tunnel.js
@@ -36,6 +36,7 @@ export default class Tunnel extends CircuitElement {
         this.xSize = 10;
         this.plotValues = [];
         this.inp1 = new Node(0, 0, 0, this);
+        this.checked = false; // has this tunnel been checked by another paired tunnel
         this.setIdentifier(identifier || "T");
         this.setBounds();
         // if tunnels with this's identifier exist, then set the bitwidth to that of those tunnels
@@ -85,6 +86,11 @@ export default class Tunnel extends CircuitElement {
      * resolve output values based on inputData
      */
     resolve() {
+        // Don't check for paired tunnels' value if already checked by another paired tunnel (O(n))
+        if (this.checked) {
+            this.checked = false;
+            return;
+        }
         // Check for bitwidth error since it bypasses node's resolve() function which usually checks bitwidths
         for (const tunnel of this.scope.tunnelList[this.identifier]) {
             if (tunnel.inp1.bitWidth !== this.inp1.bitWidth) {
@@ -96,6 +102,7 @@ export default class Tunnel extends CircuitElement {
                 tunnel.inp1.value = this.inp1.value;
                 simulationArea.simulationQueue.add(tunnel.inp1);
             }
+            if (tunnel !== this) tunnel.checked = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


public/src/modules/Tunnel.js
```diff
@@ -184,6 +172,12 @@ export default class Tunnel extends CircuitElement {
         if (this.scope.tunnelList[this.identifier])
             this.scope.tunnelList[this.identifier].push(this);
         else this.scope.tunnelList[this.identifier] = [this];
+
+        // Change the bitwidth to be same as the other elements with this.identifier
+        if(this.scope.tunnelList[this.identifier] && this.scope.tunnelList[this.identifier].length > 1) {
+            this.newBitWidth(this.scope.tunnelList[this.identifier][0].bitWidth);
```
satu0king, line 178: just set bitwidth here, no need to call newBitWidth

public/src/modules/Tunnel.js
```diff
@@ -75,38 +80,21 @@ export default class Tunnel extends CircuitElement {
...
+        // Check for bitwidth error since it bypasses node's resolve() function which usually checks bitwidths
+        for (const tunnel of this.scope.tunnelList[this.identifier]) {
+            if (tunnel.inp1.bitWidth !== this.inp1.bitWidth) {
+                this.inp1.highlighted = true;
+                tunnel.inp1.highlighted = true;
+                showError(`BitWidth Error: ${this.inp1.bitWidth} and ${tunnel.inp1.bitWidth}`);
+            }
+            if (tunnel.inp1.value !== this.inp1.value) {
+                tunnel.inp1.value = this.inp1.value;
+                simulationArea.simulationQueue.add(tunnel.inp1);
```
satu0king, line 97: This will actually be a N2 algorithm u no? Each tunnel checks if other tunnels are updated. This was how it was before also. Can we eliminate it somehow?

satu0king (changes requested):

public/src/modules/Tunnel.js
gr455: We can make another boolean array that holds whether the tunnel was checked. Then if the element is checked we just do checked[i] = false and return instead of resolving

gr455: @satu0king can you please review

PR #2045, opened 2020-12-16
New Feature diode with required changes 

mayank-kabra2001:
Fixes #1992 

#### Describe the changes you have made in this PR -
Created a new feature of diode . 

### Screenshots of the changes (If any) -
image1
image2

Note: Please check **Allow edits from maintainers.** if you would like us to assist in the PR.

PR #2046, opened 2020-12-16
New Feature buzzer

mayank-kabra2001:
Fixes #

Created a new feature of buzzer - 

Inputs are frequency of the sound , duration upto which sound is played , and a play button . 
eg . input of 426 as frequency and 2 as duration and 1 as play will produce a C chord(426 Hz) for 2sec if play input is 1.

In properties its volume can be adjusted .
Also whether we want that frequency (eg. C chord) to be sharp or flat (i.e. C# or C') can be controlled by a checkbox .  

Screenshot from 2020-12-17 02-09-28

### DIFF
--- a/public/src/moduleSetup.js
+++ b/public/src/moduleSetup.js
@@ -43,6 +43,7 @@ import TwoComplement from './modules/TwoComplement';
 import VariableLed from './modules/VariableLed';
 import XnorGate from './modules/XnorGate';
 import XorGate from './modules/XorGate';
+import Diode from './modules/Diode' ; 
 import Clock from './sequential/Clock';
 import DflipFlop from './sequential/DflipFlop';
 import Dlatch from './sequential/Dlatch';
@@ -63,6 +64,8 @@ import verilogPower from './modules/verilogPower';
 import verilogShiftLeft from './modules/verilogShiftLeft';
 import verilogShiftRight from './modules/verilogShiftRight';
 import verilogRAM from './sequential/verilogRAM';
+import Buzzer from './modules/Buzzer'; 
+
 
 export default function setupModules() {
     var moduleSet = {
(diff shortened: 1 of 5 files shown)


public/src/ux.js
```diff
@@ -233,7 +233,7 @@ export function showProperties(obj) {
     if(layoutModeGet()){
         // if an element is selected, show its properties instead of the layout dialog
         if (simulationArea.lastSelected === undefined || ['Wire', 'CircuitElement', 'Node'].indexOf(simulationArea.lastSelected.objectType) !== -1){
-            $('#moduleProperty').hide();
+            $('#modulePropshowPropertieserty').hide();
```
abstrekt, line 236: is this a mistake?

public/src/moduleSetup.js
```diff
@@ -43,6 +43,7 @@ import TwoComplement from './modules/TwoComplement';
 import VariableLed from './modules/VariableLed';
 import XnorGate from './modules/XnorGate';
 import XorGate from './modules/XorGate';
+import Diode from './modules/Diode' ;
```
abstrekt, line 46: trouble with webpacker due to this import. might have mixed code with #2045 ?

abstrekt (changes requested):
I was testing this feature out, there some to be some unrelated codes maybe relating to PR #2045, I'm not sure if this feature is working as it suppose to in a normal use case but there is a bug where the buzzer won't stop -- to reproduce try playing the buzzer without providing any input in for the duration.

Seems like a great feature --would love to see this through, can you fix the above mention issues so that @satu0king can review it to be working properly?

gr455:
This PR still needs work. A bunch of bugs with this one

tachyons:
Closing due to inactivity, feel free to re-open once the issues are addressed

PR #2058, opened 2021-01-03
Add GDPR check functionality

aa0501:
Add users (new as well as existing) to explicitly agree to Privacy Policy and TnC. (they won't be able to use the platform without it).


app/controllers/users/registrations_controller.rb
```diff
@@ -3,7 +3,9 @@
 class Users::RegistrationsController < Devise::RegistrationsController
   prepend_before_action :check_captcha, only: [:create]
   before_action :configure_sign_up_params, only: [:create]
-  invisible_captcha only: %i[create update], honeypot: :subtitle unless Rails.env.test?
+  before_action :configure_account_update_params, only: [:update]
+  respond_to :html, :js
+  invisible_captcha only: %i[create], honeypot: :subtitle unless Rails.env.test?
```
tachyons, line 8: Related ?

aayush-05: Yeah, accepting GDPR policy is basically an update operation to the user model. Also, the invisible captcha is not a related change but I don't think we are having captcha div in the update page; so disabled it at the controller level too.

Asrani-Aman (commented):
closing this as it now has too much conflicts and no resolution since too long

Issue #2026, opened 2020-12-11
Offline save not working as desired.


shubhankarsharma00:
**Describe the bug**
Successfully saving the project gives a message showing which is a hint to where this probably is going wrong.
when you save the project to your local object it is saved with key "undefined" as compared to earlier when it was saved with projectId as key.

![img](https://files.slack.com/files-pri/TJTJJ6R1A-F01HE4ZK5AL/image.png)

**To Reproduce**
1. Go to 'circuitverse.org/simulator'
2. Make something
3. Try to save offline
4. See error as weird success message and in local storage > ProjectList > data with key as "undefined".

**Expected behavior**

**Screenshots**

**Device Information [optional]:**

**Additional context**

**Are you working on this issue? (Yes/No)**

dipanshparmar:
Working!

satu0king:
@dipanshparmar, this issue is not the same one you fixed. Please check this issue carefully.

dipanshparmar:
Sure @satu0king

dipanshparmar:
@satu0king I'm no longer working on this issue because my industrial training is starting from today up to 31 Jan. So I'm busy between this time span. I'm sorry for the inconvenience. If anyone wants to take this issue then please go ahead and if no one takes this issue till 31st of Jan then I'll make a PR for this after 31 Jan. Thanks

raffayatiq:
Hello. I would like to work on this issue. However, I am not sure what code is handling the Save Offline functionality. This is the HTML for the Save Offline button inside the canvas: 
![image](https://user-images.githubusercontent.com/58357644/103442317-e5315700-4c76-11eb-91b6-85b4e83d17ef.png)
And this is HTML for the Save Offline link when you click on "Projects" top-left on the File menu:
![image](https://user-images.githubusercontent.com/58357644/103442334-1873e600-4c77-11eb-88c5-ff18e726f070.png)
Can anyone guide me on what code is being linked to these two elements and how is it being linked? Is it through the id="saveOffline"? Is the save to the local storage being handled by a Rails Controller or through JavaScript? If so, what file is handling it? I am searching for it in the source code but I am somewhat lost. Any guidance would be appreciated!

dipanshparmar:
Hey @raffayatiq,
The save offline functionality is implemented using javaScript (Local Storage). These are the code sections that are probably handling this functionality:

**File Location**: public/js/**data.js**
[...]

PR #2071, opened 2021-01-16
Fixed saveOffline functionality

raffayatiq:
Fixes #
A fix to issue #2026 

I am setting the global variable projectId to the project id being generated in the setupEnvironment function in src/setup.js. It is being set to undefined in src/utils.js. Before this fix, this value remained undefined. Due to it being undefined, each project was being saved in the undefined key when trying to save projects offline. Now, each project is being correctly saved in its own project id.
After correcting this, I also found a bug where when a user loads a project offline the projectId value remained as the last time the projectId global variable was set. When a user loads a project, the projectId should be set to the newly loaded project's id. I fixed this by setting window.projectId to the newly loaded project's id, in the openOffline function in src/data/project.js.
Also, I fixed a generateId() is not defined error (file src/data/project.js, line 137), which occurred when New Project was clicked. This error happened because of not importing generateId function from src/utils.js in src/data/project.js.
image

### DIFF
--- a/public/src/data/project.js
+++ b/public/src/data/project.js
@@ -4,7 +4,7 @@
 /* eslint-disable no-restricted-globals */
 /* eslint-disable no-alert */
 import { resetScopeList, scopeList, newCircuit } from '../circuit';
-import { showMessage, showError } from '../utils';
+import { showMessage, showError, generateId } from '../utils';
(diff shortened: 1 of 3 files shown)


tachyons:
Merged before reviewing by mistake, can you please reopen the PR ?

Issue #1503, opened 2020-06-18
Bug while signing up in the "Name"


krishnakant08:
Signing up, in the "Name" option.
While signing up where the blank asks to fill the Name.
The blank accepts any random symbol and any amount of unwanted space.
Example: the Name blank accepts "/*--*/+uioyr76$%^^" as name.

**To Reproduce**
1. Go to 'https://circuitverse.org/'
2. Click on 'login/signup'
3. Scroll down to 'signup'
4. In the Name blank when try to enter name as "/*--*/+uioyr76$%^^" it accepts and registers.
5.This shouldn't happen. This is where the error lies.

**Expected behavior**
When tried to enter Name as "/*--*/+uioyr76$%^^" it should display a prompt message to enter a valid name and not let the user to sign up with invalid names.

**Device Information [optional]:**
 - OS: [windows 10]
 - Browser [chrome]
 - Version [Version 83.0.4103.106 (Official Build) (64-bit)]

I am working on this,
[name error.docx](https://github.com/CircuitVerse/CircuitVerse/files/4800361/name.error.docx)

krishnakant08:
I am working on this.

lopeselio:
can I work on this ?

krishnakant08:
> can I work on this ?

sure.

bankebihariagrawal:
Can I work on this sir?

Kushagraw12:
Is it done, or can i take it up?? 
Thanks!

bijoysijo:
I'm working on this.

tachyons:
Accept combination of alphabets and spaces, remove special characters and numbers, 

Refer https://makandracards.com/makandra/35477-matching-unicode-characters-in-a-ruby-1-9+-regexp

koushik-mahamkali:
can i work on this sir? if this is not fixed

PR #2086, opened 2021-01-21
Signup validation#1503

Grumpyyash:
Fixes the issue #1503 of taking even special characters as input for sign-up usernames

I have added a form validation code to users model to not allow certain special characters to be input as username

Screenshot of the changes that this code does-
Resolve


tachyons:
Thanks 🚀

Issue #2065, opened 2021-01-09
Export Circuits as GIFs


sal2701:
**Is your feature request related to a problem? Please describe.**
At the moment, we have the option to download the circuits as images ( PNGs, JPEGs, etc ). Some users prefer to have the option to share their circuits as GIFs. This requires an external tool.

**Describe the solution you'd like**
To help with this we can have an inbuilt GIF exporter making it easy for users to share their circuits with others.

**TODO**
- We already have a menu for Downloading Images in ```Tools > Download Image```. There is also a GIF button which currently downloads PNG.
- We have to add the feature to create a GIF.
- Add the option to start and stop the recording for the GIF
- Add option to Download GIF

**Are you working on this? (Yes/No)**
No

ishandeveloper:
I might be able to work on this. Can you explain a bit more about "Add the option to start and stop the recording for the GIF" ?

What do we want to animate in the gif, should there be a seek bar or something else, etc?

arnab15:
hello,
I have knowledge of javascript ,nodejs,reactjs,expressjs, Mongodb but I haven't contributed to any open-source project so I'm a beginner.I would like to contribute to this project.
how should I start can anyone suggest to me the way should I start contributing.

ZadeAbhishek:
@sal2701 
my idea is if any body wants to save the circuit as GIF the animated part would be the input vs output.
so for external tools to save as GIF we can use "gifshot" library. 
https://yahoo.github.io/gifshot/ 
can i help you with this feature ?

sal2701:
Since the circuit outputs change based on the inputs & the clock cycle, users might want to record it for multiple cycles or even for given seconds. So I feel what we need is a proper way to facilitate this. Yeah a seek bar might make it easier.

@ishandeveloper interested in this issue?

Hey @arnab15 good you want to contribute here. I suggest that you first setup CircuitVerse and use the simulator, it will give you a better idea of the platform & issues as well.

@ZadeAbhishek I will let you know if no one else is working on this issue. Feel free to try out other issues

ZadeAbhishek:
yes sure

ishandeveloper:
Thanks @sal2701 . Thanks for offering me the chance to work on this, however something came up and I might be busy for next few weeks. You can move forward and assign this to @ZadeAbhishek , meanwhile I'll be keeping an eye out for other issues, once I'm free : )

sal2701:
[...]

PR #2098, opened 2021-01-31
Gif feature

ZadeAbhishek:
Fixes # 
Export Circuits as GIFs #2065 
This pull request is for to add new feature of exporting Circuit as GIF

1.Added two new file gifshot.js and gifshot.min.js (This files include code which help to implement this feature)
2.Added new code to export as GIF location (public/src/data/save.js)
3.Added new line main.stylesheet.css
4.Added new line in (app/views/simulator/edit.html.erb)

Screenshot from 2021-01-31 17-00-21
Screenshot from 2021-01-31 17-00-27
final3
Final


satu0king:
@shubhankarsharma00 @abstrekt  can u review this PR? 
@ZadeAbhishek thanks for this PR. I would like to see this done. 
But please try to add the library as a NPM dependency instead of adding it as code.

ZadeAbhishek:
sure @satu0king will modify and also research about mp4 format addition.

abstrekt (changes requested):
I tested the feature locally, seems like it's a bit performance-heavy on my setup.
A few things to be noted:
* Dragging the canvas view while recording breaks the gif file being recorded, users should be restricted until the capturing finishes.
* Nothing is being captured in `full screen view`, better to disable the full screen option while on `gif` else the feature might seem broken to some.
* Can you check the files changes in this PR, too many changes-- what are `~/.vscode-root/` directory files?  Better to remove unnecessary files.
* I could notice a bug where the recording is way longer than 15s (can’t seem to reproduce)
* For the blinking animation, please use display none instead of of visibility hidden --also maybe set it to a proper red color that we usually see in recording icon.
* For now can we have stop recording button? maybe the blinking animation can be the stop button.. also it would be good to display the time as the capturing goes on.


Great work!

ZadeAbhishek:
Sure noted working on this

ZadeAbhishek:
Hey @satu0king  i have implemented stop button feature . also change bit UI 
Screenshot from 2021-02-27 16-42-26

here is the video
https://user-images.githubusercontent.com/66520848/109385660-81947480-791b-11eb-87d3-4fef80de9d63.mp4

also working on video feature implementation

ZadeAbhishek:
hey @satu0king  @abstrekt 
i have done somechanges i you asked
0.seperate GIF and GIF-animated 
1.Added stop button
2.Added video feature 
3.Max limit to record GIF is 2 min i.e 120 sec 
4.used npm to add dependency

Limitation 
Video recoding as black as blackground i have used WebRTC which is most commly used to record canvas and WebRTC does not support alpha channel yet. WebRTC will convert pixel blocks into I420 pixel format which has 24 bit channel and does not have support for Alpha channel. On the other end , stream will be converted back to another pixel but alpha channel information will be lost. 

also i am not able to lock simulator while recording GIF is there any why to lock  ???

shubhankarsharma00:
@ZadeAbhishek this looks like a great feature. One thing I will like to point is that it would be great if we don't have to put a prompt for asking frame rate for GIF and have it in the menu next to the transparent background option because prompts don't work well with the Electron app. There is no defined way to lock the simulator, setting listenToSimulator = false and embed = true will shut all hotkey binds and wire logic but things like drag and select which can be done with the mouse will still work. Anyways I don't see much need to lock anything except maybe dragging and you'll have to write the logic for that which can be a bit trickier.

ZadeAbhishek:
@shubhankarsharma00 working on it .

ZadeAbhishek:
Hey @shubhankarsharma00  i have done the changes please check this

ZadeAbhishek:
hey @satu0king @shubhankarsharma00 @abstrekt  can you check the feature i have done the changes

PR #2146, opened 2021-03-11
7 segment color#49

### LINKED ISSUE #49: Color setting for 7-segment displays
The 7-segment and hex displays are lacking a color setting, which would be great to have. The variable LED is also lacking a color setting.

felixschlegel:
Closes #49 

As stated in #49 the VariableLed, HexDisplay and SevenSegDisplay components are lacking a color setting feature. This PR lets the user change the colour of their component in the properties panel.

# VariableLed
variableLed 
Screenshot 2021-03-15 at 16 41 06

# HexDisplay
Screenshot 2021-03-15 at 16 49 09Screenshot 2021-03-15 at 16 49 30

# SevenSegDisplay
Screenshot 2021-03-15 at 16 53 03Screenshot 2021-03-15 at 16 53 13


satu0king:
Good work! 
So do fix the codeclimate issues. Install codeclimate extension to see the issues in the code. 

You asked about whether to refactor using superclass right? So the thing is there is no good super class for this. We cannot say `ColoredElement` as a superclass. Also its actually bad to use inheritance for this. But it is very good to use composition for this. Do read about composition vs inheritance.

simulator/src/modules/VariableLed.js
```diff
@@ -27,6 +28,19 @@ export default class VariableLed extends CircuitElement {
...
         this.fixedBitWidth = true;
+        this.color = color;
+        this.actualColor = colorToRGBA(this.color);
+    }
+
+    /**
+     * @memberof VariableLed
+     * function to change color of the led
+     */
+    changeColor(value) {
```
satu0king, line 39: Why not using properties here?

felixschlegel:
What do you think about having the ColorProperty class? @satu0king

gr455:
Implemented in #2741. Thaks a lot!

PR #2151, opened 2021-03-13
Feature : gif and video record 

ZadeAbhishek:
Feature #
Added GIF and Video recording feature 

0.separate GIF and GIF-animated
1.Added stop button
2.User can also select frame rate .
3.Added video feature
4.Max limit to record GIF is 2 min i.e 120 sec

Limitation
Video recording has black background i have used WebRTC which is most commly used to record canvas and WebRTC does not support alpha channel yet. WebRTC will convert pixel blocks into I420 pixel format which has 24 bit channel. On the other end , stream will be converted back to another pixel but alpha channel information will be lost.

UI change 
Screenshot from 2021-03-13 20-22-08

video  of example 
https://user-images.githubusercontent.com/66520848/111034151-42494600-843a-11eb-9dc3-a3bb4a608983.mp4

GIF example 
Main (5)

video
https://user-images.githubusercontent.com/66520848/111034197-74f33e80-843a-11eb-8c19-b35732290e05.mp4


ZadeAbhishek:
@satu0king can you please review this PR.

abstrekt (changes requested):
- Seems like a bug was introduced here with the project title 
image
- Make sure proper color variables are set, check `color_theme.scss`, set proper color variables should fix this
image

- panning the canvas while recording breaks the gif (not in the case of video), as previous frames have old position gets captured along with the new position, 
one solution could be adding a backdrop with some opacity, set proper z-index & disable `pointer-event` to none? -- the idea here is to place this backdrop above everything but behind the stop recording button, acting as a barrier.

image
- I'm unable to verify if the framerate is actually set, still the default framerate should appear when opening the modal, and the limitation on the framerate range should be enforced in the modal itself.
- There are other style issues that you can come to later but the above suggestions are necessary for roll out.

ZadeAbhishek:
@abstrekt working on it .

abstrekt:
also, the commit counts are increasing in this PR-- makes it hard to do code review, make sure to squash (git rebase -i) similar & small commits like 
image

& always use proper commit messages, not like this 
image

Tip: use code climate vscode extension to check the code climate issue in the editor itself.
👍🏻

ZadeAbhishek:
okay @abstrekt

ZadeAbhishek:
Hey @abstrekt I have done the changes please once review this one

SAMREEN111-hash:
Hi, I came across the old PR related to GIF/Video recording and found the idea very interesting.
I’m interested in working on this feature using the current codebase, but wanted to check if this is something maintainers would like to pursue now.
Any guidance on whether this is relevant and where to start would be really helpful. Thanks!

Me-Priyank:
Outdated . no MediaRecorder/gif.js anywhere in current main

PR #2177, opened 2021-03-17
Minimalist gitpod configuration

RakeshSeervi:
#### Fixes the following issues
I was not liking the current configuration of gitpod... like it would work fine when it is started the first time...

But would fail when stopped and restarted and would succeed only on multiple trials.

I have updated the configuration to the best of my knowledge... still if someone suspects of some feature not working as expected, lemme know I shall test it out and update the status here :)

Explicitly defined what actions needs to be taken for each exposed port, so that new contributors do not find themselves in a fix as to what needs to be done (although the wiki has instructions).
With the above changes, we wouldn't be prompted to take any action, instead the ports would be made public by default and the webpage being served would open in a new browser tab.

Moved the instruction to start redis server to the `before` task, so that it is the first most task done.
It cannot stay in `init` task as the redis server needs to be started everytime the workspace is started.
Secondly, it cannot stay in `command` task as the database initialisation in the `init` task requires redis to be running beforehand.

Not knowing the purpose of installing nvm, I have removed that instruction. Will add it if asked to do so.

Removed duplicate instructions from the `command` task.


RakeshSeervi:
I would like to update gitpod wiki pages once this pr is merged. I am open to removing the debugging statements present, but I feel they are helpful.

.gitpod.yml
```diff
@@ -2,35 +2,46 @@ image:
...
-      yarn &&
-      bundle exec rails db:create &&
-      bundle exec rails db:schema:load &&
-      bundle exec rails db:migrate &&
       bundle exec rake sunspot:solr:start &&
-      bundle exec rails db:seed &&
-      (./bin/webpack-dev-server . &) &&
-      bundle exec rails server
+      bundle exec rails server &&
+      ./bin/webpack-dev-server
```
satu0king, line 43: How will this work? Unless you start one of the processes in the background, both processes cannot run in parallel. Check if the simulator page works properly. Trying making a change and see if it is reflected.

satu0king (changes requested):

satu0king:
Other changes look good CC: @tachyons

.gitpod.yml
RakeshSeervi: ahh, right!

I made small UI change to test the behavior, and found that the changes did not reflect even upon refreshing!

Locally, the changes didn't require refreshing, they were reflected immediately.

And in gitpod, after going back to how it was configured previously, the changes were reflected upon refreshing... 

so the last commit resolves the issue, but can we run one of the two command in another terminal or this way it is better?

RakeshSeervi: @tachyons @satu0king

tachyons:
Thanks @RakeshSeervi , we moved away from gitpod

PR #2052, opened 2020-12-30
Added Date Created and Date Modified below project description

raffayatiq:
Fixes #
A fix to this issue: https://github.com/CircuitVerse/CircuitVerse/issues/2047
#### Describe the changes you have made in this PR -
I have added Date Created and Date Modified below the project description so that users can see when a project was made and when it was last modified.
### Screenshots of the changes (If any) -
image

Note: Please check **Allow edits from maintainers.** if you would like us to assist in the PR.

Issue #2047, opened 2020-12-17
Showing date/time of creation and last modification of a circuit


r-arvind:
**Is your feature request related to a problem? Please describe.**
When I look at a circuit, it would be really helpful to see when the circuit was created and when it was last modified.

**Describe the solution you'd like**
Display `Date created` and `Date Modified` below the circuit description

**Describe alternatives you've considered**
The best estimate of date-time I could make was by using the timestamps of the comments. But it's not always helpful.

**Are you working on this? (Yes/No)**
No

shubhankarsharma00:
hmm should be an easy feature to add. Created_at and Updated_at is already a part of schema. 
https://github.com/CircuitVerse/CircuitVerse/blob/129182c0f422c9482abe149f730e03292ec1a7ef/db/schema.rb#L281-L282

Cronus1007:
@ShubhankarKG Let me implement this one then.

shubhankarsharma00:
@Cronus1007 you tagged the wrong guy I guess, but yeah go ahead.

silentC1C4D4:
Is this issue still open? @Shivansh2407

PR #2178, opened 2021-03-17
Feature: Date Created and Date Modified in project show page

raffayatiq:
Fixes #
A fix to this issue: #2047 
Original PR: #2052 

I have added Date Created and Date Modified functionality in Project show page, according to relative and absolute time in the way Github does it.
About the code I added to en.yml: time_ago_in_words() (as well as the similar function distance_of_time_in_words) returns the time as "about x hour" or "less than x minutes" etc. I wanted to cut out these words like "about" and "less than" in order to display time exactly the way Github does. Since time_ago_in_words() and similar "time ago" functions use localization keys, I have overridden them in en.yml to display the time exactly the way Github displays it, in an i18n-safe way.

Screenshot from 2021-03-17 19-41-46

Screenshot from 2021-03-17 20-06-22

Screenshot from 2021-03-17 19-58-11
Screenshot from 2021-03-17 20-04-43

I have also added functionality to show the complete time stamp when you hover over the date:
Screenshot from 2021-03-17 19-58-24
It is in UTC +00:00 timezone because Rails stores time in the database in this timezone. I can make it show the time stamp in the user's own timezone but there are multiple ways to implement it. If it needs to be done, here is how I think we can show it in the user's timezone:

1) At the backend, we can store users' timezone in the users table when they sign up. Users can change this timezone any time.
2) At the frontend, we can use JavaScript to detect the timezone of a user and change it. I've searched about this and this seems to be the preferred way, although it might be a little complex.

If it needs to be done, I can get started on this. Although, the timezone is only shown when a user hovers over the Date Created and Date Modified time.

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -80,6 +80,22 @@
             </p>
             <div class="projects-details-description"><%= sanitize@project.description %></div>
           <% end %>
+          <p title="<%= @project.created_at.strftime("%B %d, %Y, %l:%M %p UTC %:z") %>">
+            <strong>Date Created : </strong>
+            <% if @project.created_at >= 14.days.ago %>
+              <%= time_ago_in_words(@project.created_at, include_seconds: true) + " ago" %>
+            <% else %>
+             <%= @project.created_at.strftime("%b %d, %Y") %>
+            <% end %>
+          </p>
+          <p title="<%= @project.updated_at.strftime("%B %d, %Y, %l:%M %p UTC %:z") %>">
+            <strong>Date Modified : </strong>
+            <% if @project.updated_at >= 14.days.ago %>
+              <%= time_ago_in_words(@project.updated_at, include_seconds: true) + " ago" %>
+            <% else %>
+              <%= @project.updated_at.strftime("%b %d, %Y") %>
+            <% end %>
+          </p>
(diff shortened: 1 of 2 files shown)


satu0king (changes requested):
Looks neat!
But I think it is better to use

Created: 
Updated: 

Instead of 
Date Created: and
Last Modified

CC @Shivansh2407

raffayatiq:
> Looks neat!
> But I think it is better to use
> 
> Created:
> Updated:
> 
> Instead of
> Date Created: and
> Last Modified
> 
> CC @Shivansh2407

Yes that does look better. But there is just one thing:
image
What about the extra space before the colon? For example there is an extra space before the label's colon i.e. "Author : " and "Project access type : ". Should I get rid of the space? So that the labels become "Author: user1" and "Project Access Type: Public". Personally I think we should and also capitalize the 2nd label.

nitin10s:
Yes, please get rid of the extra space in other fields, and 2nd label seems fine, we need not capitalize each word.

Shivansh2407 (approved):
Great Work @raffayatiq 🎉 Looks good to me @satu0king 👍

raffayatiq:
@Shivansh2407 Thank you! :slightly_smiling_face:

Issue #2203, opened 2021-04-02
clashing inputs on generating 2 combinational circuits


pranay-simejia:
**Describe the bug**
Simulator overwrites the first input element when generating 2nd combinational circuit with different boolean variables.

**To Reproduce**
1. Go to '[simulator](https://circuitverse.org/simulator)'
2. Click on 'tools>combinational analysis.'
4. Create a combinational circuit with boolean variables A, B, C ... and generate a circuit.
5. Repeat steps 1 to 4 with unused boolean variables. 
**Expected behavior**
A new circuit should be generated away from the existing circuit.

**Screenshots**
![image](https://user-images.githubusercontent.com/70064696/113412638-66170080-93d6-11eb-8010-b4e08d4426c1.png)

**Are you working on this issue? (Yes/No)**
No.

ZadeAbhishek:
Hey @pranay-simejia  can i work on this issue

pranay-simejia:
Sure, go-ahead. @ZadeAbhishek

pulkit-30:
Hey @tachyons @ZadeAbhishek, 
This issue still exists 

<img width="1045" alt="Screenshot 2022-01-03 at 6 55 33 PM" src="https://user-images.githubusercontent.com/76155456/147935702-d653217e-c76f-41f1-be97-cd2acdd215e8.png">

Is anyone working on this?
If not, may you please assign me this Issue?

ZadeAbhishek:
@pulkit-30  have you tried this branch #2205

pulkit-30:
Yes, @ZadeAbhishek, I tried this and the changes are looking good.
but the changes need to be optimized, too many changes for a small bug.
and it would be great if we can implement the requested feature by reviewers @pranay-simejia.
>could we add circuits to the next row if the x coordinate reaches a certain limit? With the current functionality, the circuits get pretty small if they are added in a single row.

PR #2205, opened 2021-04-03
Bug Fixed : Clashing inputs on generating 2 combinational circuits

ZadeAbhishek:
Fixes  
This PR is for the issue #2203
In combinationalAnalysis.js file when function drawCombinationalAnalysis was call X,Y postion of circuit are static so each time the user import CombinationalAnalysis Circuit it import on same place.Problem comes when user export it more than one time.
In combinationalAnalysis in function drawCombinationalAnalysis i add findimension function to check if any circuit is present.If yes MaxWidth of that circuit is new X corrdinate of the CombinationalAnalysis Circuit. if no then default values

Previous 
Main(1)

Current
Main(2)

### DIFF
--- a/simulator/src/combinationalAnalysis.js
+++ b/simulator/src/combinationalAnalysis.js
@@ -11,9 +11,17 @@ import AndGate from './modules/AndGate';
 import OrGate from './modules/OrGate';
 import NotGate from './modules/NotGate';
 import { stripTags } from './utils';
+import { findDimensions } from './canvasApi';
+import simulationArea from './simulationArea';
 
 var inputSample = 5;
-var dataSample = [['01---', '11110', '01---', '00000'], ['01110', '1-1-1', '----0'], ['01---', '11110', '01110', '1-1-1', '0---0'], ['----1']];
+var posincrement = 0;
+var dataSample = [
+    ['01---', '11110', '01---', '00000'],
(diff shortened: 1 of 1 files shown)


pranay-simejia:
Good work! @ZadeAbhishek 
But it would still start overlapping when the `startPosX` reaches simulationarea.maxWidth`. In that case you need to increment `startPostY`

ZadeAbhishek:
> Good work! @ZadeAbhishek
> But it would still start overlapping when the `startPosX` reaches `simulationarea.maxWidth`. In that case you need to increment `startPostY`

@pranay-simejia  no I have kept the default coordinate also so this overlap never happen again

ZadeAbhishek:
@tachyons can you review this PR.

pranay-simejia:
@ZadeAbhishek , could we add circuits to the next row if the x coordinate reaches a certain limit? With the current functionality, the circuits get pretty small if they are added in a single row.

image

ZadeAbhishek:
> @ZadeAbhishek , could we add circuits to the next row if the x coordinate reaches a certain limit? With the current functionality, the circuits get pretty small if they are added in a single row.
> 
> What do you think?
> 
> image

In that case we have to do something like when user click on particular position circuit must spawn there because we cannot determine the max width of X. Also the main AIM was that circuit should not collapse which is solved i guess.

PR #2247, opened 2021-05-10
Fix: Improper forking of assignment submissions

gr455:
Fixes Bug where group mentor could not see submissions post assignment deadline

Proper forking of submission projects


satu0king (approved):

satu0king:
Seems to a good hot fix. 
@gr455 is it possible to reuse fork method in the model class? this way, we can avoid issues like this in the future?

gr455:
> Seems to be a good hot fix.
> @gr455 is it possible to reuse fork method in the model class? this way, we can avoid issues like this in the future?

Yeah we can just call fork from within the job

tachyons (approved):

PR #2273, opened 2021-06-06
backend independent I18n architecture for simulator

pavanjoshi914:
Fixes #
Backend independent I18n architecture for simulator

Dependencies:
i18n-js
yaml-loader

### DIFF
--- a/app/javascript/i18n-js/config.js
+++ b/app/javascript/i18n-js/config.js
@@ -1,0 +1,6 @@
+import I18n from "i18n-js"
+import translations from "./translations"
+
+I18n.locale = window.locale
+I18n.translations = translations
+export default I18n
--- a/app/javascript/i18n-js/translations.js
+++ b/app/javascript/i18n-js/translations.js
@@ -1,0 +1,4 @@
+import { en } from "../../../config/locales/simulator/en.yml"
+import { hi } from "../../../config/locales/simulator/hi.yml"
+
+export default { en, hi }
--- a/app/views/layouts/simulator.html.erb
+++ b/app/views/layouts/simulator.html.erb
@@ -4,6 +4,7 @@
     <title>CircuitVerse - Digital Circuit Simulator online</title>
     <%= csrf_meta_tags %>
     <script src="/js/pace.js"></script>
+    <script>window.locale = "<%= I18n.locale%>"</script>
     <%= stylesheet_link_tag  'simulator', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_pack_tag 'simulator', 'data-turbolinks-track': 'reload' %>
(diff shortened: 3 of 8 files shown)


app/javascript/i18n-js/translations.js
```diff
@@ -0,0 +1,4 @@
+import { en } from "../../../config/locales/simulator/en.yml"
```
tachyons, line 1: This still has dependency on rails i18n mapping cc @satu0king

pavanjoshi914: exactly how ? so that i can understand concept behind it and opt for solutions.

If we dont want to import translations from yml to json . then we can directly prepare json files for translations and can dump into ```I18n.translations```

pavanjoshi914: Any updates here @satu0king @tachyons ?

satu0king: @pavanjoshi914, yeah we don't want any relationship with rails i18n mapping. To make it simple, all the files related to the simulator should be in the`simulator/src` folder. 

The `.yml` files can be used as long as it is in the `simulator` directory. Because this directory will be moved away in the future to become a standalone repo. Does that make sense now?

app/views/layouts/simulator.html.erb
```diff
@@ -4,6 +4,7 @@
     <title>CircuitVerse - Digital Circuit Simulator online</title>
     <%= csrf_meta_tags %>
     <script src="/js/pace.js"></script>
+    <script>window.locale = "<%= I18n.locale%>"</script>
```
satu0king, line 7: This line should be the ONLY line that is rails specific.

app/javascript/i18n-js/config.js
```diff
@@ -0,0 +1,6 @@
+import I18n from "i18n-js"
```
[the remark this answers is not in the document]
satu0king: Please don't resolve codeclimate conversations without fixing! Atleast the simple ones need to be done.

satu0king (changes requested):

pavanjoshi914:
Yaa this was only dummy pr to introduce setup. I am taking care of code quality in mergable prs.

app/javascript/i18n-js/translations.js
pavanjoshi914: Yaa. I used yml only with yml loader only following rails conventions.

If I prepare json file (no requirement of yml loader) and dump them into ``|I18n.translations```  present in config.js in a similar fashion using import statements .. 

Does it will work?

pavanjoshi914: If not.... In any case I am opting  for banana i18n library by wikimedia... That might help us in creating better solution...

pavanjoshi914:
Closing as we shifted to banana-i18n library

PR #2275, opened 2021-06-09
Gsoc tb 1

satu0king:
Fixes #


satu0king:
Please add screenshots

config/routes.rb
```diff
@@ -121,6 +121,11 @@
     get "/", to: "simulator_old#new", as: "simulator_old_new"
     get "/embed/:id", to: "simulator_old#embed"
   end
+
+  scope "/testbench" do
+    get "/", to: "test_sets#new", as: "testbench_new"
+    post "/create", to: "test_sets#create", as: "testbench_create"
```
satu0king, line 127: I think you can just use resources for this, it will automatically add all CRUD routes.

app/views/test_sets/new.html.erb
```diff
@@ -0,0 +1,350 @@
+<% content_for :title, "CircuitVerse - Create Test Set" %>
+<body style="font-family: sans-serif;">
```
satu0king, line 2: Use partials. We need to reuse this in edit.html.erb as well right?

app/views/test_sets/new.html.erb
```diff
@@ -0,0 +1,350 @@
...
+            <%=  f.hidden_field 'data', id:'tb-data-field' %>
+            <%=  f.hidden_field 'testset_access_type', id:'tb-access-field', value: 'Public' %>
+            <div style="float: right; display: inline-block;">
+                <%=  f.button "Save", class: 'lower-button save-buton' %>
+            </div>
+        <% end %>
+    </div>
+
+</body>
+<style>
```
satu0king, line 59: Create a stylesheet

app/views/test_sets/new.html.erb
```diff
@@ -0,0 +1,350 @@
...
+    .inp {
+        bgcolor:"red";
+    }
+
+</style>
+
+
+<script>
+
+    /* Listen to submit click */
```
satu0king, line 182: Similarly, lets use asset pipeline for this.

satu0king:
@tachyons should we use asset pipeline for this or should we use webpack?

tachyons:
How are we planning to integrate workbench on offline app ?

shubhankarsharma00:
@tachyons I guess one simple way would be to make the same static HTML page open up in a new electron pop up window and whenever someone creates a testbench, just store it's json in the working directory (as a json file) and copy it onto the clipboard. So just like in the webpage the user can reuse the same testbench too as it will be stored locally. It would take just a few changes to the current html at `/testbench` endpoint.

PR #2279, opened 2021-06-11
Touch Compatibility to CircuitVerse -ZadeAbhishek || GSoC'21

ZadeAbhishek:
### **Feature**
The PR is for **touch compatibility to CircuitVerse**
CircuitVerese just supports the Mouse on the browser till now. In this PR the website will support both Touch and mouse on Brower simultanously. 

This is Done by adding touch listeners

- 1. touchstart
- 2. touchmove
- 3. touchend

1. Added **touch listeners** in listeners.js
2. Function for pan on the simulator is made common for both touch and mouse in listerners.js
3. All Mouse related variable name is changed for example **mouseDownX **will become **DownX** across complete Codebase 
> This is done because touch and mouse use the same variable so to avoid confusion in future above step is done.

4.**Pinch to Zoom** is implemented to zoom in and zoom out on the simulator using pinch.

### DIFF
--- a/simulator/src/embedListeners.js
+++ b/simulator/src/embedListeners.js
@@ -3,13 +3,19 @@
 // Refer listeners.js
 import simulationArea from './simulationArea';
 import {
-    scheduleUpdate, update, updateSelectionsAndPane,
-    wireToBeCheckedSet, updatePositionSet, updateSimulationSet,
-    updateCanvasSet, gridUpdateSet, errorDetectedSet,
+    scheduleUpdate,
+    update,
+    updateSelectionsAndPane,
+    wireToBeCheckedSet,
+    updatePositionSet,
+    updateSimulationSet,
+    updateCanvasSet,
+    gridUpdateSet,
+    errorDetectedSet,
 } from './engine';
 import { changeScale } from './canvasApi';
 import { copy, paste } from './events';
-import { ZoomIn, ZoomOut} from './listeners';
+import { ZoomIn, ZoomOut } from './listeners';
 
 var unit = 10;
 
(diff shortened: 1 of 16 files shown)


simulator/src/embedListeners.js
```diff
@@ -1,205 +1,204 @@
...
+    errorDetectedSet(false);
+    updateSimulationSet(true);
+    updatePositionSet(true);
+    updateCanvasSet(true);
+
+    simulationArea.lastSelected = undefined;
+    simulationArea.selected = false;
+    simulationArea.hover = undefined;
+    const rect = simulationArea.canvas.getBoundingClientRect();
+    simulationArea.DownRawX = (e.clientX - rect.left) * DPR;
```
ShreyaPrasad1209, line 37: change DownRawX to TouchDownX or something similar. Adding "touch" prefix in every variable will be easier to understand in the future.

ShreyaPrasad1209 (changes requested):
Can you add a GIF/screencast showcasing your changes?

simulator/src/embedListeners.js
ZadeAbhishek: @ShreyaPrasad1209 do you mean `TouchMouseDownX` like this

manjotsidhu:
@ZadeAbhishek First of all you have done a great job in implementing Touch Compatibility in such a short span of time. You have added approx. 4.4k lines and deleted around 4.3k, I appreciate the work done but since these all changes are just divided in just 2 commits, it becomes really hard to review stuff. Try to distribute your work in commits, your commits matter not only for the maintainers but also for the community to get idea of the changes.

https://medium.com/@nawarpianist/git-commit-best-practices-dab8d722de99

satu0king (changes requested):
Hey good job!

Yeah, we need to run linting on our entire code soon to avoid the issue @manjotsidhu mentioned. We also need to setup precommit hooks as well. @gr455 @shubhankarsharma00 can you check this. 

I apologize if this was discussed before and I missed adding my inputs. While the point of having a common name for the mouse to avoid confusion is valid, I do not think the current change is the right design decision 😛. 

Think about it, most of the other parts of the code just want coordinates, irrespective of whether it was clicked via mouse or via touch. And `mouseX` is an excellent variable name then. 

Yes if we can come up with a nice name that represents both without ambiguity that would be great but `touchMouse` is not that, in fact everyone will ask what is `touchMouse` the first time they read it. I couldn't come up with a name as well. 

Another issue is that `simulator.x` is not a good name at all. It just means x coordinate of something. How would a future contributor know it is x coordinate of a mouse?

Please revert the name change. I apologize for the extra work.

Also its hard to review currently, so can you tell if u have done any logical changes to any other parts of the code? Or have u just added new touch listeners?

ZadeAbhishek:
Its fine @satu0king i will revert changes.

shubhankarsharma00:
@ZadeAbhishek props to you because this looks like some really nice work. You need to take care of the variable names as satvik said. Also, please try to completely separate the linting commit and touch commit so it is a little easier to review. Also can I know how you tested this and on windows or ubuntu (if on laptop)?

shubhankarsharma00 (changes requested):
@ZadeAbhishek  props to you because this looks like some really nice work. You need to take care of the variable names as satvik said. Also, please try to completely separate the linting commit and touch commit so it is a little easier to review. Also can I know how you tested this and on windows or ubuntu (if on laptop)?

ZadeAbhishek:
> @ZadeAbhishek props to you because this looks like some really nice work. You need to take care of the variable names as satvik said. Also, please try to completely separate the linting commit and touch commit so it is a little easier to review. Also can I know how you tested this and on windows or ubuntu (if on laptop)?

yes @shubhankarsharma00 so I have tested on 
OS: Fedora OS 34
Browser: Firefox and chrome
Local devices: iPhone 7 and Andriod realm and one plus (only landscape mode because in low-end Andriod devices simulator does not work on  portrait mode its lags)

PR #2289, opened 2021-06-17
Testbench/Engine: Add Engine support for combinational and sequential tests

gr455:
Changes and flow:
 - UI for creating tests (for now served as a static html file)
 - Implemented testbench engine for Combinational Tests
 - Validation for I/O, Clock and Reset pins
 - Runs tests as fast as possible without rendering(to increase speed) on Simulator
 - Currently uses Labels for identification of I/O. To be deprecated in favour of identifiers

 * Testbench engine stops the clock and runs tests by controlling the clock itself.
   This is required so that tests are not throttled by clock time.
 * Testbench engine returns test results with observed output next to expected output
 * Testbench engine returns with failure message on validation failure

Testing:
 - Create a test at `/testbench`. Clicking the `Save` button will log the test JSON in the console
 - Paste this JSON on Simulator under Tools > Use TestBench
 - Click `Run All`
 - Test Results are displayed in the console for now. Console logs an Array of groups
 - for each group, group.outputs has two fields - values (expected values) and result (observed values)

Test  Creator UI:
Screenshot from 2021-06-17 12-38-54

TestBench Manual Mode UI
Screenshot from 2021-06-22 17-21-52

### DIFF
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -19,7 +19,7 @@
   <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
     <ul class="navbar-nav navbar-menu noSelect pointerCursor" id="toolbar">
       <li class="dropdown nav-dropdown">
-        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><%= t("simulator.nav_item.project") %><span></span></a>
+        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Project<span></span></a>
         <ul class="dropdown-menu">
           <% if not @project.nil? %>
              <li><a class="dropdown-item logixButton text-left pl-1" href="<%= user_project_url(@project.author, @project) %>" target="_blank">Project Page</a></li>
@@ -35,7 +35,7 @@
         </ul>
       </li>
       <li class="dropdown nav-dropdown">
-        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><%= t("simulator.nav_item.circuit") %><span class="caret"></span></a>
+        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Circuit<span class="caret"></span></a>
         <ul class="dropdown-menu">
           <li><a class="dropdown-item text-left logixButton pl-1" id="newCircuit">New Circuit + </a></li>
           <li><a class="dropdown-item text-left logixButton pl-1" id="newVerilogModule">New Verilog <br> Module</a></li>
@@ -43,9 +43,10 @@
         </ul>
       </li>
       <li class="dropdown nav-dropdown">
-        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><%= t("simulator.nav_item.tools") %><span class="caret"></span></a>
+        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Tools<span class="caret"></span></a>
         <ul class="dropdown-menu">
           <li><a class="dropdown-item text-left pl-1 logixButton" id="createCombinationalAnalysisPrompt">Combinational<br>Analysis</a></li>
+          <li><a class="dropdown-item text-left pl-1 logixButton" id="createTestBenchPrompt">Use<br>TestBench</a></li>
           <li><a class="dropdown-item text-left pl-1 logixButton" id="bitconverter">Hex-Bin-Dec<br>Converter</a></li>
           <li><a class="dropdown-item text-left pl-1 logixButton" id="createSaveAsImgPrompt">Download Image</a>
           <li><a class="dropdown-item text-left pl-1 logixButton" id="colorThemes">Themes</a>
@@ -54,7 +55,7 @@
         </ul>
       </li>
       <li class="dropdown tour-help nav-dropdown">
-        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><%= t("simulator.nav_item.help") %><span class="caret"></span></a>
+        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Help<span class="caret"></span></a>
         <ul class="dropdown-menu">
             <li><a class="dropdown-item text-left pl-1 logixButton" id="showTourGuide">Tutorial Guide</a></li>
             <li><a class="dropdown-item text-left pl-1" href="https://docs.circuitverse.org" class="" target="_blank" role="button" aria-haspopup="true" aria-expanded="false">User Manual</a></li>
@@ -522,6 +523,11 @@
   id="combinationalAnalysis"
   style="display: none;"
   title="BooleanLogicTable"
+></div>
+<div
+  id="testBench"
+  style="display: none;"
+  title="TestBench"
 ></div>
 <div
   id="insertSubcircuitDialog"
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -218,6 +218,46 @@
     </div>
   </div>
 </div>
+<div class="testbench-manual-panel draggable-panel">
+  <div class="panel-header">Testbench - Manual Mode
+    <span class="fas fa-external-link-square-alt maximize panel-button-icon"></span>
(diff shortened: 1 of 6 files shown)


app/views/simulator/edit.html.erb
```diff
@@ -19,7 +19,7 @@
   <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
     <ul class="navbar-nav navbar-menu noSelect pointerCursor" id="toolbar">
       <li class="dropdown nav-dropdown">
-        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><%= t("simulator.nav_item.project") %><span></span></a>
+        <a href="#" class="" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Project<span></span></a>
```
satu0king, line 22: Internationalization has been undone?

simulator/src/testbench.js
```diff
@@ -1,3 +1,230 @@
...
+function validate(data, scope=globalScope) {
+    if(!checkDistinctIdentifiers(data)) return { status: "fail", message: "Duplicate identifiers" };
+    for(let dataIO of data.groups[0].inputs.concat(data.groups[0].outputs)){
+        const matchIO = scope.Input.concat(scope.Output).some(function(simulatorIO) {
+            // TODO: label to identifier
+            return simulatorIO.label === dataIO.label && simulatorIO.bitWidth === dataIO.bitWidth;
+        });
+
+        if(!matchIO) return { status: "fail", message: "Some inputs/outputs not present in circuit or wrong bitwidths" };
+    }
```
satu0king, line 141: I think this can be refactored to be simpler to read. Also you need to check if it is dataIO is input and not output etc.

simulator/src/testbench.js
```diff
@@ -1,3 +1,230 @@
...
+
+    // Validate presence of reset if test is sequential
+    if(data.type === "seq"){
+        const resetPresent = scope.Input.some(function(simulatorReset) { return simulatorReset.label === "RST" });
```
satu0king, line 145: Check BitWidth to be 1

simulator/src/testbench.js
```diff
@@ -1,3 +1,230 @@
...
+    play();
+    reset.state = 0;
+    play();
+}
+
+/**
+ * Creates deep clone of scope given to the function
+ * @param {Scope} scope - the scope to be cloned
+ */
+function cloneScope(scope) {
```
satu0king, line 224: I think cloning scope is not a good idea. But let us see.

simulator/src/testbench.js
```diff
@@ -1,3 +1,230 @@
...
+    const results = runAll(data, scope);
+    console.log(results);
+}
+
+/**
+ * Run all the tests automatically. Called by runTestBench()
+ * @param {Object} data - Object containing Test Data
+ * @param {Scope=} scope - the circuit
+ */
+function runAll(data, scope) {
```
satu0king, line 89: Needs more review but looks good on first glance!

satu0king (changes requested):

satu0king:
Good work!

app/views/simulator/edit.html.erb
gr455: Not sure how this happened, but I think I need to rebase my branch

simulator/src/testbench.js
gr455: Yes, I dropped the idea, not using clone scope anymore

simulator/src/testbench.js
gr455: Okay :+1:

simulator/src/testbench.js
shubhankarsharma00: Can we also have a "warning" if there is no label named CLK?

shubhankarsharma00 (commented):
Looking really good for initial commits. Few things, like removing I/O fields on the testbench page felt missing but I am guessing you'll be working on that later. Error messages can be better explained. 

The code for the most part is really clean. Great work.

gr455:
> Looking really good for initial commits. Few things, like removing I/O fields on the testbench page felt missing but I am guessing you'll be working on that later. Error messages can be better explained.
> 
> The code for the most part is really clean. Great work.

Yes I will be working on error messages and finalizing the testbench creator, once done with the main testbench. Thanks a lot for the review!

simulator/src/testbench.js
gr455: We don't require to name the clock anything specific, we can issue a warning if there is no Clock element though

simulator/src/testbench.js
```diff
@@ -1,3 +1,512 @@
...
+    $('.testbench-manual-panel .tb-data#data-case').children().eq(1).text(1);
+    $('#tb-manual-table-inputs-head').attr('colspan', inputCount);
+    $('#tb-manual-table-outputs-head').attr('colspan', outputCount);
+
+    $('#tb-manual-table-labels').html('<th>Label</th>');
+    $('#tb-manual-table-bitwidths').html('<td>Bitwidth</td>');
+    for (const io of data.groups[0].inputs.concat(data.groups[0].outputs)) {
+        const label = `<th>${io.label}</th>`;
+        const bw = `<td>${io.bitWidth}</td>`;
+        $('#tb-manual-table-labels').append(label);
```
gr455, line 462: TODO: escape labels and bitwidths before appending

simulator/src/testbench.js
```diff
@@ -1,3 +1,512 @@
...
+    <p>Enter the Test JSON: <input id='testJSON' type='text'  placeHolder='{"type": "comb", ...}'></p>
+    `;
+    $('#testBench').dialog({
+        resizable: false,
+        width: 'auto',
+        buttons: [
+            {
+                text: 'Run All',
+                click() {
+                    const testJSON = $('#testJSON').val();
```
satu0king, line 49: Error handling of incorrect input

simulator/src/testbench.js
```diff
@@ -1,3 +1,512 @@
...
+
+/**
+ * Run all the tests automatically. Called by runTestBench()
+ * @param {Object} data - Object containing Test Data
+ * @param {Scope=} scope - the circuit
+ */
+function runAll(data, scope) {
+    // Stop the clocks
+    // TestBench will now take over clock toggling
+    changeClockEnable(false);
```
satu0king, line 108: Actually disabling clock is not needed as this is a blocking function but it is still nice to keep if the behavior changes in the future. But you missed an important thing. You have to make the initial clock value to 0.

simulator/src/testbench.js
```diff
@@ -1,3 +1,512 @@
...
+        for (let case_i = 0; case_i < group.n; case_i++) {
+            // Set and propagate the inputs
+            setInputValues(inputs, group, case_i, scope);
+            // If sequential, trigger clock now
+            if (data.type === 'seq') tickClock(scope);
+            // Get output values
+            const caseResult = getOutputValues(data, outputs);
+            // Put the results in the data
+            for (const outName of caseResult.keys()) {
+                const output = group.outputs.find((dataOutput) => dataOutput.label === outName);
```
satu0king, line 122: `.find()` is a linear time complexity. This might need refactoring.

simulator/src/testbench.js
```diff
@@ -1,3 +1,512 @@
...
+ * @param {number} groupIndex - Index of the group to be tested
+ * @param {number} caseIndex - Index of the case inside the group
+ * @param {Scope} scope - The circuit
+ */
+function runSingleSequential(data, groupIndex, caseIndex, scope) {
+    const { inputs, outputs, reset } = bindIO(data, scope);
+    const group = data.groups[groupIndex];
+
+    // Stop the clocks
+    changeClockEnable(false);
```
satu0king, line 233: Set clock to 0

simulator/src/testbench.js
gr455: Since we do not step the clock halfway for now, it shouldn't matter if we set the clock to 0 right because one of the two ticks will always be an up edge

satu0king (approved):
Looks good. Make the few changes I pointed out or put TODOs so we don't forget them. I suspect there will be more changes to come but its good for now.

simulator/src/testbench.js
gr455: But our sample size is pretty low for this to be a problem right. The output count is fairly low for this to be significant

gr455:
> An error occurred when fetching issues.
> 
> View more on [Code Climate](https://codeclimate.com/github/CircuitVerse/CircuitVerse/pull/2289).

Thank god

PR #2303, opened 2021-06-24
refactored js version1

ShreyaPrasad1209:

1. Refactored the JavaScript code in three files under simulator folder as of now. If this PR is approved then I will refactor rest of the JS code.
2. Used ES6 arrow notation and replaced var with let.
3. Used a `for-of` loop instead of a `for` loop.

None

### DIFF
--- a/simulator/src/canvas2svg.js
+++ b/simulator/src/canvas2svg.js
@@ -18,11 +18,11 @@
 
 "use strict";
 
-var STYLES, ctx, CanvasGradient, CanvasPattern, namedEntities;
+let STYLES, ctx, CanvasGradient, CanvasPattern, namedEntities;
 
 //helper function to format a string
-function format(str, args) {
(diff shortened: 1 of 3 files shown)


simulator/src/canvas2svg.js
```diff
@@ -18,20 +18,20 @@
```
tachyons, line 18: This is a thirdparty lib, better not to change this file

ShreyaPrasad1209: Okay which files should I not modify under simulator ?

tachyons: Everything except thirdparty scripts. there are only few such files in simulator folder, rest are moved to npm

ShreyaPrasad1209: Ok then I am closing this PR.

PR #2306, opened 2021-06-27
banana-i18n integration with simulator

pavanjoshi914:
### quick glance of the library
1. define constructor eg: ```banana```
2. access all the methods provided by library via ```banana.method_name()```
3. set locale value using ```setLocale```
4. setup ```fetch api``` accordingly to load translations
5. use methods in js files

### more  Setup information
1. ```banana-i18n.js``` library file```
2. ```config-i18n.js``` configuration for i18n.
3. translation loading is done using ```fetch API``` with ```async ```and``` await``` so that function is not completed till fetch request is completed.
4. cache is disabled during ```http request``` so that when ```json``` is modified ```json.response()``` will not contain old data and will render new data
5. locale is brought from backend to frontend using ```window.locale```
6. other functionalities such as plurals , interpolations/placeholders working fine 

### things needed to be handled further
1. page should not load until translations are loaded via fetch api.
sometimes program executes and delay in fetch can result in printing of key names only and not actual translations.
2. error handling 
fetch api doesnt handle 404 request. logic has to be build for handiling server and bad http request errors such as 404 file not found.
3. library is not resolving fallback language as well as fallback messages with webacker . (work in progress)


pavanjoshi914:
CC @satu0king @tachyons

public/js/config-i18n.js
```diff
@@ -0,0 +1,19 @@
+// intialize banana-i18n constructor
+const banana = new Banana({ finalFallback: 'en' });
+// set locale
+banana.setLocale(window.locale);
+// setup fetch api with async/await to load json files
+async function fetchJson() {
```
tachyons, line 6: Can we fetch the locales directly instead of making a http call ?

pavanjoshi914: Thats seems the drawback of the library they only provide ```banana.load()``` which can hold the constant having messages defined inside it . for loading messages from json explicitly they suggest fetch()

here is the documentation - https://github.com/wikimedia/banana-i18n#loading-the-messages
there demo example uses the same  - https://github.com/wikimedia/banana-i18n/blob/master/demo/js/demo.js

Current library is a model of``` jquery.i18n ```and purely written in js .  Jquery I18n provides load method to load translations but not this one
https://github.com/wikimedia/jquery.i18n#dynamic-loading-using-load-method

tachyons: You can load the i18n translations to messages object and pass it to the constructor. It is just that banana is doesn't dictate a directory structure. You need to implement loading yourself

pavanjoshi914: Yes, loading is to be done manually, so gone with fetch API 

Now I am thinking of bringing translations in form of J's objects and parse them using json.parse for  further use

Any suitable loading technique you want to suggest here? As there can be more than one techniques?

tachyons: Using fetch require http call, and it won't play nicely with the offline app. Here is one approach with load method https://github.com/zhaofengli/refill/blob/2ee4c00623f20bbf9b86e76679f763217f7a241e/web/src/main.js#L36

tachyons: Another approach https://github.com/smc/mlmorph-web/blob/5d741f617c0ed541fb65a249b039d74d1daf4d15/src/main.js#L9

public/i18n/en.json
```diff
@@ -0,0 +1,12 @@
+{
```
tachyons, line 1: this  should be in simulator folder

pavanjoshi914: yes , desired the same but fetch api makes requests such as ```localhost/i18n/en.json```  in that case had to use public directory .

will modify this also as i change loading mechanism

app/views/layouts/simulator.html.erb
```diff
@@ -3,7 +3,11 @@
   <head>
     <title>CircuitVerse - Digital Circuit Simulator online</title>
     <%= csrf_meta_tags %>
+    <!-- bring locale value from backend to frontend -->
+    <script>window.locale = "<%= I18n.locale %>"</script>
     <script src="/js/pace.js"></script>
+    <!--load banana-i18n library-->
+   <script src="/js/banana-i18n.js"></script>
```
tachyons, line 10: Do you still need this as we are fetching from npm?

pavanjoshi914: I did mistake here, now importing from node_modules. corrected in new commit!

simulator/src/combinationalAnalysis.js
```diff
@@ -11,6 +11,14 @@ import AndGate from './modules/AndGate';
 import OrGate from './modules/OrGate';
 import NotGate from './modules/NotGate';
 import { stripTags } from './utils';
+import banana from './config-i18n';
+
+
```
tachyons, line 16:
```suggestion

```

simulator/src/combinationalAnalysis.js
```diff
@@ -11,6 +11,12 @@ import AndGate from './modules/AndGate';
 import OrGate from './modules/OrGate';
 import NotGate from './modules/NotGate';
 import { stripTags } from './utils';
+import banana from './config-i18n';
+// library basic test
+console.log(banana.i18n('appname-title'));
+// interpolation and plurals check
+console.log(banana.i18n('result', 1));
+console.log(banana.i18n('result', 4));
```
tachyons, line 19:
```suggestion
```

tachyons: Remove console log

simulator/src/combinationalAnalysis.js
```diff
@@ -26,7 +27,7 @@ var sampleOutputListNames = ['X'];
 export function createCombinationalAnalysisPrompt(scope = globalScope) {
     scheduleBackup();
     $('#combinationalAnalysis').empty();
-    $('#combinationalAnalysis').append("<p>Enter Input names separated by commas: <input id='inputNameList' type='text'  placeHolder='eg. In A, In B'></p>");
+    $('#combinationalAnalysis').append(banana.i18n('appname-title'));
```
tachyons, line 30: You need proper translation here

pavanjoshi914:
I was thinking of creating fresh PR with single clean commit...
If discussion is required for future reference I will continue with this.

tachyons:
Feel free to create New PR

pavanjoshi914:
closing this , new pr for the same -  https://github.com/CircuitVerse/CircuitVerse/pull/2318

PR #2327, opened 2021-07-13
Featrecommendation system

DevanshD3:
Fixes #

* Temporary code for the scripts used in the recommendation system
NA


Recommendation-System/clean.py
```diff
@@ -0,0 +1,84 @@
...
+import string
+import re
+# from datetime import datetime
+
+# startTime = datetime.now()
+
+printable = set(string.printable)
+
+# Opening JSON file
+f = open(r'C:\Users\devan\OneDrive\Desktop\Results\dataset.json', encoding="utf8")
```
ayan-biswas0412, line 16: I think the file path given here should not be in codebase because the file path is your local environment specific and the file will not be fetched in the deployment environment.

DevanshD3: Yes Yes! You are absolutely right. This PR however is a draft PR, not the final one yet. I just wanted to move these codes from my local machine and @Shivansh2407 had suggested me to make a draft PR. I will edit all these things later on once the model is ready.

PR #2330, opened 2021-07-15
Testbench/UI: New UI for Testbench on Simulator and Creator

gr455:
- Implemented new UI
 - Prototype Detailed Results UI

New UI

Creator reordering


simulator/src/data/backupCircuit.js
```diff
@@ -32,6 +32,9 @@ export function backUp(scope = globalScope) {
     // Storing all nodes
     data.allNodes = scope.allNodes.map(extract);
 
+    // Storing test attached to scope
+    data.test = scope.test;
```
satu0king, line 36:
```suggestion
    data.testBenchData = scope.testBenchData;
```
Rename variable to be more descriptive

satu0king:
image
Theme needs to be corrected here

satu0king (changes requested):
Things to do:
1. Show error when validation fails (after testbench is created)
2. Allow testbench to be created even if validation fails
3. Fix the space issue in identifier name - use equivalent of `.strip()`
4. Edit Layout should not have testbench - Check where all timing diagram is toggled? Also replicate logic of timing diagram, or refactor both of them
5. When circuit is switched, testbench needs to be updated. 
6. Lots of small UI Improvements needed

simulator/src/testbench.js
```diff
@@ -73,32 +57,160 @@ export function createTestBenchPrompt() {
...
+
+        // Next Group Button
+        $('.tb-case-button#next-group-btn').on('click', () => {
+            if (currentGroup >= globalScope.test.groups.length - 1) return;
+            currentGroup++;
+            currentCase = 0;
+            setUICurrentCase(globalScope.test, currentGroup, currentCase);
+
+            result = runSingleTest(globalScope.test, currentGroup, currentCase, globalScope);
+            setUIResult(globalScope.test, currentGroup, currentCase, result);
```
satu0king, line 152: The above 3 lines seem to be repeated everywhere, can this be improved? Like say we have a function which says
`testbench.setCurrentTestCase(testbench.currentGroup + 1, testbench.currentCase + 1);`
This function itself can validate if that specific action is possible etc and update the UI and run the test accordingly?

I think `currentGroup` and `currentCase` needs to be scoped as a class object to achieve the above.

simulator/src/utils.js
```diff
@@ -299,3 +299,12 @@ export function promptFile(contentType, multiple) {
...
 }
+
+export function escapeHtml(unsafe) {
+    return unsafe
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;')
+        .replace(/"/g, '&quot;')
+        .replace(/'/g, '&#039;');
+}
```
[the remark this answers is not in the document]
shubhankarsharma00: So this function is only used once to sanitize label input defined here earlier:
https://github.com/CircuitVerse/CircuitVerse/blob/2e1ae715d0e67d5e3068fa52fd0deaa8fcc35cb6/simulator/src/ux.js#L416
We don't use it anywhere else afaik.

gr455:
Whoops messed up commit history. Closed in favor of https://github.com/CircuitVerse/CircuitVerse/pull/2366

PR #2347, opened 2021-07-19
Test addition for LTI

ayan-biswas0412:
lms parameters are generated successfully

image

### Issue 
I am sending the parameters in the post request but it always give Unauthorised 401 and the problem is genrating from ```app/controllers/lti_controller.rb``` line no 25-26 can you guys please help me in this issue
According to the documentation the request should be POST and the data which I generated in the spec should be send as ```application/x-www-form-urlencoded``` to the ```lti.launch``` , how to do that ?

### DIFF
--- a/spec/requests/lti_spec.rb
+++ b/spec/requests/lti_spec.rb
@@ -1,22 +1,66 @@
 # frozen_string_literal: true
-require 'rails_helper'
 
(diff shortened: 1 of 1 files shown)


ayan-biswas0412:
cc @tachyons @Shivansh2407 @satu0king

PR #2356, opened 2021-07-24
Stability(Simulator): Fix Subcircuits resulting in Queue Overflow

gr455:
* According to the design of subcircuits, the localscope is
   directly connected to the main scope. Earlier, output values of
   subcircuits were being set to undefined and propagated before
   the localscope resolved. This resulted in repeated adding of
   some nodes in the simulation queue, causing an overflow.

 * This commit fixes the above problem by making the subcircuits
   not propagate undefined values.

### DIFF
--- a/simulator/src/subcircuit.js
+++ b/simulator/src/subcircuit.js
@@ -564,17 +564,21 @@ export default class SubCircuit extends CircuitElement {
     }
 
     /**
-     * Not resolvable. Subcircuit's scope is added to the main scope directly
+     * By design, subcircuit element's input and output nodes are wirelessly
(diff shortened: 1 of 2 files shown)


simulator/src/subcircuit.js
```diff
@@ -571,35 +564,19 @@ export default class SubCircuit extends CircuitElement {
     }
 
     /**
-     * not used because for now everythiing is added onto the globalscope
+     * Not resolvable. Subcircuit's scope is added to the main scope directly
```
satu0king, line 567: Can you document this a bit more here, explaining about `makeConnections`?

satu0king (approved):

PR #2379, opened 2021-08-07
Feature flag added for LMS Integration Project

ayan-biswas0412:
1. LMS integration done under a feature flag `lms_integration`
2. Default condition of feature `lms_integration` is `true`

####  Group page
image

#### Assignment creation form
image

#### LMS side
image

#### Group page
image

#### Assignment creation form
image

#### LMS side 
image

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -6,7 +6,9 @@ class AssignmentsController < ApplicationController
   before_action :set_group
   before_action :check_access, only: %i[edit update destroy reopen]
   after_action :check_reopening_status, only: [:update]
-  after_action :allow_iframe_lti, only: %i[show]
+  if Flipper.enabled?(:lms_integration)
+    after_action :allow_iframe_lti, only: %i[show]
+  end
 
   # GET /assignments
   # GET /assignments.json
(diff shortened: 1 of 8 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -6,7 +6,9 @@ class AssignmentsController < ApplicationController
   before_action :set_group
   before_action :check_access, only: %i[edit update destroy reopen]
   after_action :check_reopening_status, only: [:update]
-  after_action :allow_iframe_lti, only: %i[show]
+  if Flipper.enabled?(:lms_integration)
```
tachyons, line 9: You have to pass this condition as a block, otherwise you will have to restart the server to take this effect

spec/requests/lti_spec.rb
```diff
@@ -3,88 +3,89 @@
 require "rails_helper"
 
 describe LtiController, type: :request do
+  before do
+    Flipper.enable(:lms_integration)
```
tachyons, line 7: Disable in after block

app/controllers/assignments_controller.rb
ayan-biswas0412: how to do that can you suggest ?

spec/requests/lti_spec.rb
ayan-biswas0412: in which place I have to disable it ?

ayan-biswas0412: yah done

app/controllers/assignments_controller.rb
ayan-biswas0412: done

tachyons (approved):

PR #2398, opened 2021-08-20
[POC] Enable Lms integration for selected users

tachyons:
POC for conditionally enable las integration for users

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -99,7 +99,7 @@ class User < ApplicationRecord
   end
 
   def flipper_id
-    "User;#{id}"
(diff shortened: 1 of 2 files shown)


ayan-biswas0412 (commented):
As per the LTI flow LMS first request a launch data to our CircuitVerse and then we start performing our logic i.e. 
1. Authenticating the teacher
2. Redirecting the student to the assignment

but here if the feature is enabled conditionaly based on the authentication then LMS will not be able to access the `/launch` path and our functionality will not be accomplished that I think .

ayan-biswas0412:
See the PR #2399

PR #2401, opened 2021-08-21
Security: Add request rate limiting

gr455:
Adds request based rate limitting

Current limits
 - No limits on assets
 - 5 requests every 20 seconds on /users/sign_in
 - 300 requests in 5 minutes everywhere else


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,46 @@
...
+  # Throttle all non asset requests
+  throttle('req/ip', limit: 300, period: 5.minutes) do |req|
+    req.remote_ip unless req.path.start_with?('/assets')
+  end
+
+  # More agressively throttle login requests
+
+  # Throttle by IP
+  throttle('logins/ip', limit: 5, period: 20.seconds) do |req|
+    if req.path == '/users/sign_in' && req.post?
```
tachyons, line 29: We have a login endpoint over api as well. Also limit password reset endpoint

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,46 @@
...
+  if !Rails.env.production?
+    Rack::Attack.enabled = !ENV['DISABLE_RACK_ATTACK']
+  end
+
+  # Use redis for rack attack since config.cache_store is not set
+  redis_client = Redis.new
+  Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(redis_client)
+
+  # Throttle all non asset requests
+  throttle('req/ip', limit: 300, period: 5.minutes) do |req|
```
tachyons, line 21: This is not needed as we are already under cloudflare, also this can be tricky when multiple students accessing circuitverse from college lab as external IP will be same for all

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+  # Disable if DISABLE_RACK_ATTACK and if env is not production
+  if !Rails.env.production?
+    Rack::Attack.enabled = !ENV['DISABLE_RACK_ATTACK']
+  end
+
+  # Use redis for rack attack since config.cache_store is not set
+  redis_client = Redis.new
+  Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(redis_client)
```
tachyons, line 18: Cache store is already set, so this config is redundant

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+class Rack::Attack
+  class Request < ::Rack::Request
+    # Take remote IP from Cloudfare's headers instead of rev proxy IP
+    def remote_ip
+      if ENV["CF_PROXY_DISABLED"]
```
tachyons, line 7: Assume CF is disabled by default

gr455: If the env var is for some reason not set on the server, then that will result in all requests getting blocked. So I assumed enabled by default. Should I change it to disabled?

tachyons: It is easy for us to set this ENV, but for people who self host need documentation if are enabling this by default

gr455: Okay I'll change it then :+1:

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,85 @@
...
+      # Cloudflare stores remote IP in CF_CONNECTING_IP header
+        @remote_ip ||= (env["HTTP_CF_CONNECTING_IP"] ||
+                        env["action_dispatch.remote_ip"] ||
+                        ip).to_s
+      end
+    end
+
+    # Hack to get JSON request params
+    # Reads from the IO stream and resets the stream
+    def json_params
```
tachyons, line 19: Why was this needed ?

gr455: Rack requests do not have any better way to parse JSON params

tachyons: In that case we need to handle `JSON::ParserError` as well

tachyons (approved):
Just one more nitpick before merging

gr455:
> Just one more nitpick before merging

It is sent as a 400 Bad Request by rails

PR #2426, opened 2021-08-23
feat(i18n): localize assignments module

pavanjoshi914:
Localize assignments module

image
image
image
image


pavanjoshi914 (commented):
@tachyons what is that error from hound? I can't see any mistyped non-whitespace character "d"?

PR #2427, opened 2021-08-23
feat(i18n): localize commontator gem

pavanjoshi914:
Localize commontator gem
change locale set method from with_locale to locale in order to make commontator and devise i18n work

image
image
image
image
image
image

image


Me-Priyank:
outdated

PR #2645, opened 2021-11-25
Edit Project screen open in new tab

### LINKED ISSUE #1358: Saving Online should open in new tab
**Describe the bug**
Whenever a user clicks to save his project [online], he gets redirected to the Edit Project Screen losing his work area. 

If he wants to work further on his project he needs to click on update project button from the Edit Project Screen then click on Launch Simulator to get his circuit back.

**Expected behavior**
The Edit Project Screen should open in a new page / After clicking on the Update Project button he should be redirected to the simulator screen back.

pulkit-30:
Fixes #1358 

#### Before
Whenever a user clicks to save his project [online], he gets redirected to the Edit Project Screen losing his work area.

If he wants to work further on his project he needs to click on the update project button from the Edit Project Screen then click on Launch Simulator to get his circuit back.

#### After Changes

The Edit Project Screen opens in a new tab, on clicking on the `save online` button.

https://user-images.githubusercontent.com/76155456/143478271-baac52f3-348f-4e59-8b52-a87ec74c10e5.mov

#### Code Reference:

```
target:'_blank'
```
To open the form in the new Tab.

-----

```
  setInterval(()=>{
            window.location.href =`/simulator/edit/${url_projectName}`;
        },1000);
```
`setInterval()` is used to redirect the page to `/simulator/edit/{project_name}` after clicking on the `save online` button . So that, the user can continue working on the project and save the changes in the current project.

Please provide me with appropriate feedback so that I can continue working on this

### DIFF
--- a/simulator/src/data/save.js
+++ b/simulator/src/data/save.js
@@ -326,7 +326,6 @@ export default async function save() {
 
     projectSavedSet(true);
 
-    $('.loadingIcon').fadeIn();
     const data = generateSaveData();
 
     const projectName = getProjectName();
@@ -343,6 +342,7 @@ export default async function save() {
         // Create new project - this part needs to be improved and optimised
         const form = $('<form/>', {
             action: '/simulator/create_data',
+            target:'_blank',
             method: 'post',
         });
         form.append(
@@ -377,6 +377,10 @@ export default async function save() {
 
         $('body').append(form);
         form.submit();
+        const url_projectName = projectName.replace(/\s+/g, '-').toLowerCase();
+        setInterval(()=>{
+            window.location.href =`/simulator/edit/${url_projectName}`;
+        },1000);
     } else {
         // updates project - this part needs to be improved and optimised
         $.ajax({


nitin10s:
@gr455 can you check this?

simulator/src/data/save.js
```diff
@@ -377,6 +377,10 @@ export default async function save() {
 
         $('body').append(form);
         form.submit();
+        const url_projectName = projectName.replace(/\s+/g, '-').toLowerCase();
```
gr455, line 380: This won't always be the project slug, say if two projects have the same name, then it will be suffixed by some random string.

gr455 (changes requested):
I think the better way to achieve it is to save and not redirect to the edit screen and just have a flash message on the simulator which has a button to go to the edit screen. What do you say @nitin10s

nitin10s:
The idea is really nice but one problem with this approach is that we are hiding a lot of additional important information that users can configure while creating a new project like description, project access. The adoption for adding this information by explicitly clicking on a button in the flash message would be very low. 

Not sure what should be done... whatsay? Your views @tachyons ?

pulkit-30:
What about a popup edit form?
When the user clicks on the save button, a popup edit form opens, User can update the details there and when he click on the update button he will be redirected to the `/simulator/edit/project_name` route. 
This way he can configure the important information like description, project access, etc. and he will not lose his work.

@gr455, @nitin10s,  @tachyons.

tachyons:
> The idea is really nice but one problem with this approach is that we are hiding a lot of additional important information that users can configure while creating a new project like description, project access. The adoption for adding this information by explicitly clicking on a button in the flash message would be very low.

That information is important for discoverability which is already not good in the platform. A popup with all the fields in the project creation page may work here

nitin10s:
Yeah popup could be a good way.

pulkit-30:
Shall I go ahead? @nitin10s @tachyons @gr455

Issue #1856, opened 2020-10-30
No redo option available


vkassingh:
**Is your feature request related to a problem? Please describe.**
There is only undo option in the user interface. So I would like to add a redo option there to make the circuit verse more user friendly.

**Additional context**
![bug1](https://user-images.githubusercontent.com/42089632/97660215-e5ef1680-1a96-11eb-8d56-dc7e70a1c047.JPG)

**Are you working on this? (Yes/No)**
Yes I would like to work on this issue

satu0king:
Duplicate of #376 #redo

Palaksharma23:
I want to work on this issue , Can you please assign it to me

Prerna-0202:
duplicate of  #376

PR #2654, opened 2021-11-27
Added Redo option for the Simulator #376 #1856

### LINKED ISSUE #376: Redo option
**Currently, there is no redo option for the Simulator.

No

pulkit-30:
Fixes #376 #1856 

There was no redo option available in the simulator
Added a `redo` option to make the CircuitVerse more user-friendly.

User can redo their work using option present in option bar or Using keyboard shortcuts :
`ctrl + z` for undo .
`ctrl  + y` for redo (windows).
`cmd + z` for undo (macOs).
`shift + cmd + z` for redo (macOs).

https://user-images.githubusercontent.com/76155456/143687180-dc93b876-85e8-4b9c-a2a2-31720185fafa.mov

Please provide me with appropriate feedback on this PR so that I can continue working on this.

### DIFF
--- a/simulator/src/data/backupCircuit.js
+++ b/simulator/src/data/backupCircuit.js
@@ -58,6 +58,7 @@ export function scheduleBackup(scope = globalScope) {
     var backup = JSON.stringify(backUp(scope));
     if (scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) {
         scope.backups.push(backup);
+        scope.history = [];
         scope.timeStamp = new Date().getTime();
         projectSavedSet(false);
     }
--- a/simulator/src/data/backupCircuit.js
+++ b/simulator/src/data/backupCircuit.js
@@ -56,7 +56,7 @@ export function backUp(scope = globalScope) {
 
 export function scheduleBackup(scope = globalScope) {
     var backup = JSON.stringify(backUp(scope));
-    if ((scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) && backUp(scope).allNodes.length !== 0 ) {
+    if (scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) {
         scope.backups.push(backup);
         scope.history = [];
         scope.timeStamp = new Date().getTime();
--- a/simulator/src/data/backupCircuit.js
+++ b/simulator/src/data/backupCircuit.js
@@ -56,7 +56,7 @@ export function backUp(scope = globalScope) {
 
 export function scheduleBackup(scope = globalScope) {
     var backup = JSON.stringify(backUp(scope));
-    if (scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) {
+    if ((scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) && backUp(scope).allNodes.length !== 0 ) {
         scope.backups.push(backup);
         scope.history = [];
         scope.timeStamp = new Date().getTime();
--- a/simulator/src/data/undo.js
+++ b/simulator/src/data/undo.js
@@ -9,7 +9,7 @@ import Scope, { scopeList } from '../circuit';
 import { loadScope } from './load';
 import { updateRestrictedElementsInScope } from '../restrictedElementDiv';
 import { forceResetNodesSet } from '../engine';
-
+import { scheduleBackup } from './backupCircuit';
 /**
  * Function called to generate a prompt to save an image
  * @param {Scope=} - the circuit in which we want to call undo
@@ -19,6 +19,7 @@ import { forceResetNodesSet } from '../engine';
 export default function undo(scope = globalScope) {
     if (layoutModeGet()) return;
     if (scope.backups.length === 0) return;
+    scheduleBackup(); // Backup current globalscope
     const backupOx = globalScope.ox;
     const backupOy = globalScope.oy;
     const backupScale = globalScope.scale;
@@ -26,8 +27,11 @@ export default function undo(scope = globalScope) {
     globalScope.oy = 0;
     const tempScope = new Scope(scope.name);
     loading = true;
-    loadScope(tempScope, JSON.parse(scope.backups.pop()));
+    const undoData = scope.backups.pop();
+    scope.history.push(undoData);
+    scope.backups.length !== 0 && loadScope(tempScope, JSON.parse(scope.backups[scope.backups.length - 1]));
     tempScope.backups = scope.backups;
+    tempScope.history = scope.history;
     tempScope.id = scope.id;
     tempScope.name = scope.name;
     scopeList[scope.id] = tempScope;
--- a/simulator/src/data/undo.js
+++ b/simulator/src/data/undo.js
@@ -17,9 +17,9 @@ import { scheduleBackup } from './backupCircuit';
  * @exports undo
  */
 export default function undo(scope = globalScope) {
-    if (layoutModeGet()) return;
-    if (scope.backups.length === 0) return;
     scheduleBackup(); // Backup current globalscope
+    if (layoutModeGet()) return;
+    if (scope.backups.length < 2) return;
     const backupOx = globalScope.ox;
     const backupOy = globalScope.oy;
     const backupScale = globalScope.scale;
--- a/simulator/src/data/undo.js
+++ b/simulator/src/data/undo.js
@@ -9,7 +9,6 @@ import Scope, { scopeList } from '../circuit';
 import { loadScope } from './load';
 import { updateRestrictedElementsInScope } from '../restrictedElementDiv';
 import { forceResetNodesSet } from '../engine';
-import { scheduleBackup } from './backupCircuit';
 /**
  * Function called to generate a prompt to save an image
  * @param {Scope=} - the circuit in which we want to call undo
@@ -17,7 +16,6 @@ import { scheduleBackup } from './backupCircuit';
  * @exports undo
  */
 export default function undo(scope = globalScope) {
-    scheduleBackup(); // Backup current globalscope
     if (layoutModeGet()) return;
     if (scope.backups.length < 2) return;
     const backupOx = globalScope.ox;
--- a/app/views/simulator/edit.html.erb
+++ b/app/views/simulator/edit.html.erb
@@ -125,6 +125,10 @@
         </button>
       </div>
       <div>
+        <button type="button" class="quick-btn-redo" title="Redo" id="redoButton">
+        </button>
+      </div>
+      <div>
         <button type="button" class="quick-btn-view" title="Preview Circuit" id="viewButton"><i style="color: #ddd;" class="fas fa-expand-arrows-alt"></i>
         </button>
       </div>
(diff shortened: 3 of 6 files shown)


pulkit-30:
@tachyons @nitin10s @satu0king,
Please review this PR and provide me with appropriate feedback on this so that I can continue working on this.

nitin10s:
@gr455

gr455 (changes requested):
If you undo enough times so that the scope.backups array becomes empty, the `scheduleBackup()` method forces a new backup into scope.backups and hence emties the scope.history array. Can you think of a workaround this. 

Currently, because of this, if the scope.backups becomes empty (most amount of undos achieved), then redo is not possible

pulkit-30:
Hey @gr455 ,

What I understand is that when we undo many times, the length of `scope.backups` array becomes 0, then redo is not possible?

>Currently, because of this, if the scope.backups becomes empty (most amount of undos achieved), then redo is not possible

That what you said?

But when the length of `scope.backups` array becomes 0 then undo functions didn't run, hence `scheduleBackup()` is not called.


https://user-images.githubusercontent.com/76155456/144640441-1bb4ec03-a5d1-4483-ac84-03f9c8a7f6b0.mov


Here I console the value of `scope.backups` for every undo, and when the length of `scope.backups` become 0 then the further undo is not possible and hence it returns at the starting check ` if (scope.backups.length === 0) return; ` in undo function and the redo function is working properly.

#### Sorry, if I am wrong, and please explain to me more about the problem so that I can work on it.

gr455:
> Hey @gr455 ,
> 
> What I understand is that when we undo many times, the length of `scope.backups` array becomes 0, then redo is not possible?
> 
> >Currently, because of this, if the scope.backups becomes empty (most amount of undos achieved), then redo is not possible
> 
> That what you said?
> 
> But when the length of `scope.backups` array becomes 0 then undo functions didn't run, hence `scheduleBackup()` is not called.
> 
> 
> https://user-images.githubusercontent.com/76155456/144640441-1bb4ec03-a5d1-4483-ac84-03f9c8a7f6b0.mov
> 
> 
> Here I console the value of `scope.backups` for every undo, and when the length of `scope.backups` become 0 then the further undo is not possible and hence it returns at the starting check ` if (scope.backups.length === 0) return; ` in undo function and the redo function is working properly.
> 
> #### Sorry, if I am wrong, and please explain to me more about the problem so that I can work on it.

That's right, undo doesn't do anything anymore. But say after scope.backups is empty, if you click on the canvas, it'll call the scheduleUpdate func which will reset the history array

pulkit-30:
> That's right, undo doesn't do anything anymore. But say after scope.backups is empty, if you click on the canvas, it'll call the scheduleBackup func which will reset the history array. If the array wasn't empty, the check in the function checking `if the top element of scope.backups is same as current backup, then don't do anything` will not do anything. But in case of empty backups array, it ends up appending the current backup to backups and resetting the history array

If the history array is not empty and the user does some work and accidentally presses the redo button, he will finish up with his progress. That's why the line `scope.history = []` is important inside the scheduleBackup function.
But what you said is correct , I will look into this matter and try to come up with a solution soon.
Thank you @gr455 for your review.

pulkit-30:
Hey @gr455 ,

Finding a solution for the above problem ... I encounter one thing,
If the simulator work area is empty and we click on the canvas it'll call the scheduleBackup function and perform the backup.
But, according to me, if the simulator work area is empty then it should not make a backup. What do you think about it?

If the User doesn't work or does not have anything at the workspace then the backup for the empty workspace is meaningless and that creates the above problem. 
> That's right, undo doesn't do anything anymore. But say after scope.backups is empty, if you click on the canvas, it'll call the scheduleBackup func which will reset the history array. If the array wasn't empty, the check in the function checking if the top element of scope.backups is same as current backup, then don't do anything will not do anything. But in case of empty backups array, it ends up appending the current backup to backups and resetting the history array

---

https://user-images.githubusercontent.com/76155456/144707800-80771d1d-c6ab-4a26-aca8-c97df59b5ce2.mov

In the reference Video, I click on the empty canvas and check the globalscope in the console, the `backups` array in globalscope has one element (the empty workspace backup), which should not be there, according to me.

What do you think about it @gr455 and @nitin10s, is this backup of the empty workspace of the simulator should be there or not? Is there any use of that backup?

gr455:
> Hey @gr455 ,
> 
> Finding a solution for the above problem ... I encounter one thing,
> If the simulator work area is empty and we click on the canvas it'll call the scheduleBackup function and perform the backup.
> But, according to me, if the simulator work area is empty then it should not make a backup. What do you think about it?
> 
> If the User doesn't work or does not have anything at the workspace then the backup for the empty workspace is meaningless and that creates the above problem. 
> > That's right, undo doesn't do anything anymore. But say after scope.backups is empty, if you click on the canvas, it'll call the scheduleBackup func which will reset the history array. If the array wasn't empty, the check in the function checking if the top element of scope.backups is same as current backup, then don't do anything will not do anything. But in case of empty backups array, it ends up appending the current backup to backups and resetting the history array
> 
> ---
> 
> https://user-images.githubusercontent.com/76155456/144707800-80771d1d-c6ab-4a26-aca8-c97df59b5ce2.mov
> 
> In the reference Video, I click on the empty canvas and check the globalscope in the console, the `backups` array in globalscope has one element (the empty workspace backup), which should not be there, according to me.
> 
> What do you think about it @gr455 and @nitin10s, is this backup of the empty workspace of the simulator should be there or not? Is there any use of that backup?
> 

No, if the backups array is empty then it automatically creates a backup. For example if I just loaded an existing project, my backups array starts empty. So if I press ctrl+Z i shouldn't lose the whole circuit so it auto backs up at empty array. I think this behaviour could be changed but this is the reason why it's like that

pulkit-30:
Hey @gr455,
I tried to solve the problem with a backups array in the recent commit, Please review.

Added a condition `backUp(scope).allNodes.length !== 0 `, to check if User clicks on the empty canvas then `scheduleBackup` should not perform.

```
export function scheduleBackup(scope = globalScope) {
    var backup = JSON.stringify(backUp(scope));
    if ((scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) && backUp(scope).allNodes.length !== 0 ) {
        scope.backups.push(backup);
        scope.history = [];
        scope.timeStamp = new Date().getTime();
        projectSavedSet(false);
    }

    return backup;
}
```

simulator/src/data/backupCircuit.js
```diff
@@ -56,7 +56,7 @@ export function backUp(scope = globalScope) {
 
 export function scheduleBackup(scope = globalScope) {
     var backup = JSON.stringify(backUp(scope));
-    if (scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) {
+    if ((scope.backups.length === 0 || scope.backups[scope.backups.length - 1] !== backup) && backUp(scope).allNodes.length !== 0 ) {
```
gr455, line 59: This change will actually not fix anything. It does fix the condition that if the undo results in empty canvas, then backup does not happen. But say if I load a saved circuit, then the max undos(till backups array is empty) will set the canvas in a nonempty state and still the history array will be overwritten

simulator/src/data/undo.js
```diff
@@ -17,9 +17,9 @@ import { scheduleBackup } from './backupCircuit';
  * @exports undo
  */
 export default function undo(scope = globalScope) {
-    if (layoutModeGet()) return;
-    if (scope.backups.length === 0) return;
     scheduleBackup(); // Backup current globalscope
+    if (layoutModeGet()) return;
+    if (scope.backups.length < 2) return;
```
gr455, line 22: I think this condition is good enough, we don't need to check `backUp(scope).allNodes.length !== 0`. What do you say?

pulkit-30:
Yes, you are right,

First I have to find a way to fix the problem of clicking on empty canvas when the `backups` array is empty, then it makes a `scheduleBackup` and reset the `history` array and after that, I noticed the Project loading bug. That's why I used both 😅.

But the question arises for the empty canvas backup. Do we need that?

I am committing the changes @gr455, Please review.

simulator/src/data/undo.js
```diff
@@ -9,25 +9,29 @@ import Scope, { scopeList } from '../circuit';
...
-
+import { scheduleBackup } from './backupCircuit';
 /**
  * Function called to generate a prompt to save an image
  * @param {Scope=} - the circuit in which we want to call undo
  * @category data
  * @exports undo
  */
 export default function undo(scope = globalScope) {
+    scheduleBackup(); // Backup current globalscope
```
gr455, line 20: I see why this is a good idea, but this is actually causing some problems. Sometimes, the circuit data is different so the scheduleBackup adds it to the backups array although the circuit corresponding to the data is the same. This then stalls the undos because everytime you undo, a new data is added and it is then popped, this keeps the backups array same forever. 
So I think it's best to remove this line

gr455 (changes requested):
@pulkit-30 I have left a comment for a small change, after that try to fix as many codeclimate probelms as possible, then we can merge this

simulator/src/data/undo.js
pulkit-30: But this line is important, without `scheduleBackup()` the functionality of undo and redo function will be disturbed.
Have a look at this, 

https://user-images.githubusercontent.com/76155456/145218888-a1984ae8-8a6b-4bf9-a152-3717eb2761ad.mov

without `scheduleBackup()` we would not able to redo the last work done by the user. 
We can go for some alternative.
I’m eager to receive your feedback.

pulkit-30:
Hey @gr455, May you please help me out with this problem of codeclimate ...
`Expected an assignment or function call and instead saw an expression.`

Screenshot 2021-12-08 at 7 53 39 PM

simulator/src/data/undo.js
gr455: From what I understand, the reason that line is important is because we need to make sure that right before we undo, the top state of backups array is the current state. Without this it's possible that you pop a previous state that techincally leaves you 2 states behind in one undo (the final state is now unreachable). Right?

Then we might need to figure out some other way to achive that

gr455:
> Hey @gr455, May you please help me out with this problem of codeclimate ... `Expected an assignment or function call and instead saw an expression.`
> 
> Screenshot 2021-12-08 at 7 53 39 PM

Don't worry about it. It looks good now

simulator/src/data/undo.js
pulkit-30: Yes, I meant to say the same.

gr455: I guess you can remove that line for now and we can merge this, we can think about a better solution in another PR. Something is better than nothing. What do you say?

pulkit-30: Yes right @gr455, I will make a new commit after removing this line...

gr455 (approved):
Looks good! Thanks a lot for working on this.
We will also try to figure out a way to backup right before undo without the undo loop

pulkit-30:
> Looks good! Thanks a lot for working on this. We will also try to figure out a way to backup right before undo without the undo loop

Welcome @gr455,
Currently working on this and opening PR soon.

PR #2025, opened 2020-12-11
Dashboard now shows user has no projects in case of private projects

sarthakm21:
Fixes #2001 

#### Describe the changes you have made in this PR -
The dashboard no longer shows a blank screen in the projects section when the user's projects are private. Instead, it now shows the user has no projects image.

### Screenshots of the changes (If any) -
private1
private2

Note: Please check **Allow edits from maintainers.** if you would like us to assist in the PR.

PR #2718, opened 2021-12-15
Fix : Blank Space when the User Has all private circuits #2001

### LINKED ISSUE #2001: Dashboard : Blank Space when the User Has all private circuits
**Describe the bug**
When the user has no circuits, there is an indication shown by image that the user does not have any circuits. But if the user creates a new Circuit and Keeps the Project Acces type to be private, Then in that case ( if the user has all the circuits as private ) there is no image shown and the space remains blank.

**To Reproduce**
1. See this Profile https://circuitverse.org/users/59394 . It has only one circuit that too with Private access. If we add more private access projects then the blank space remains.

**Expected behavior**
The same Image with No Circuits Indication Can be displayed.

**Screenshots**
No Circuits : 

image

All Circuits with Private Access : 
 
From Main User Dashboard :
image

From Other User's End : 

image

**Device Information [optional]:**
 - OS: [Windows]
 - Browser [Chrome]

No

pulkit-30:
Fixes #2001 

The same Image with No Circuits Indication is displayed when the User has all Private Circuits.

#### From the Main User Dashboard
Screenshot 2021-12-15 at 8 30 10 PM

#### From theGuest User Dashboard
Screenshot 2021-12-15 at 9 00 34 PM

Inspired from #2025 

Please provide me with appropriate feedback so I can continue working on this.

### DIFF
--- a/app/controllers/users/circuitverse_controller.rb
+++ b/app/controllers/users/circuitverse_controller.rb
@@ -11,8 +11,8 @@ class Users::CircuitverseController < ApplicationController
   def index
     @profile = ProfileDecorator.new(@user)
     @projects = @user.rated_projects
-    @is_Admin = @current_user ? @profile.id == @current_user.id : false
-    @public_projects_exist = @is_Admin || @user.projects.open.exists?
+    @is_admin = @current_user ? @profile.id == @current_user.id : false
+    @public_projects_exist = @is_admin || @user.projects.open.exists?
   end
 
   def edit; end
--- a/app/controllers/users/circuitverse_controller.rb
+++ b/app/controllers/users/circuitverse_controller.rb
@@ -11,8 +11,8 @@ class Users::CircuitverseController < ApplicationController
   def index
     @profile = ProfileDecorator.new(@user)
     @projects = @user.rated_projects
-    @isAdmin = @current_user? @profile.id === @current_user.id: false
-    @public_projects_exist = @isAdmin || @user.projects.open.exists?
+    @is_Admin = @current_user ? @profile.id == @current_user.id : false
+    @public_projects_exist = @is_Admin || @user.projects.open.exists?
   end
 
   def edit; end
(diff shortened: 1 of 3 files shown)


app/controllers/users/circuitverse_controller.rb
```diff
@@ -11,6 +11,10 @@ class Users::CircuitverseController < ApplicationController
   def index
     @profile = ProfileDecorator.new(@user)
     @projects = @user.rated_projects
+    public_projects = @user.projects.map do |project|
+      policy(project).check_direct_view_access?
+    end
+    @public_projects_exist = public_projects.include?(true)
```
tachyons, line 17: If you want to check if public projects exists, most perfomant way is to use

```suggestion
   public_projects_exist = @user.projects.open.exists?
```

pulkit-30: Yes Sir this is correct,
But with this change, the admin user will not able to see his own private projects(if he has only private projects) instead it will show the `No result image`.

https://user-images.githubusercontent.com/76155456/146793412-fdbca57f-5a17-4cc9-b76a-687f91210c8f.mov

See, when the User has 1 private and 1 public project, there is no problem,
But deleting the public one, then it will not show the private project to the User.

Please check.

tachyons: That means you were setting wrong value to public_projects_exist, right ?

pulkit-30: No, I think this is correct.

pulkit-30: Hey @tachyons,
I have updated this PR,
I was wrong, about the above statement ( > No sir, I think this is correct.)
I am sorry for that 😔, learning from my mistakes
Please review 
Thanks

tachyons:
Thanks

PR #2761, opened 2022-01-06
Fix: Make Devise not store location if it's too big

gr455:
*ISSUE: Devise storing URLs too long in return_to causes CookieOverflow to be thrown.*

Overrides Devise's internal store_location_for function to not store URLs that are too big


gr455:
> _ISSUE: Devise storing URLs too long in return_to causes CookieOverflow to be thrown._
> 
> Overrides Devise's internal store_location_for function to not store URLs that are too big

@tachyons Is this feasible?

tachyons:
Can you try to do this on application controller instead

```ruby
class ApplicationController 

def store_location_for(resource_or_scope, location)
  return if location.length > 100
  super
end
end
```

Alternate fix is to store session in the backend using database or redis. But that seems overkill for this use case

tachyons:
Can you also add a test to validate this behaviour ?

tachyons:
Also a sanity test to ensure that our regular urls are below 150 bytes ?

gr455:
> Also a sanity test to ensure that our regular urls are below 150 bytes ?

Should we just use MAX_COOKIE_SIZE / 2 or something for this

tachyons:
MAX cookie size is 4096 by default. it needs to store to many values encoded in json, so better not to use half of it just store urls

app/controllers/application_controller.rb
```diff
@@ -35,6 +35,12 @@ def switch_locale(&action)
     end
   end
 
+  # Overrides Devise::Controller::StoreLocation.store_location_for to check if
+  # URL is too long to store in the session
+  def store_location_for(resource_or_scope, location)
+    super unless location && location.length > 150
```
tachyons, line 41: What about setting location=request.host if location > OUR_MAX_LIMIT. So that to avoid potential infinite loop and unexpected behaviour?

gr455:
> Also a sanity test to ensure that our regular urls are below 150 bytes ?

The longest URL that make sense to be redirected to are project pages, To be safe I think 200 bytes seems like a good number.

app/controllers/application_controller.rb
gr455: I think devise automatically redirects to the home page if there is nothing in session?

tachyons: But what if there are some from previous requests which aren't cleared yet ?

PR #2794, opened 2022-01-17
feat:added card for delete account div

shukapurv:
Added bootstrap card around delete account div

**Before**
Screenshot from 2022-01-17 07-03-46

**After**
Screenshot from 2022-01-17 07-03-50


Kartikay304 (commented):
Well, it is a good idea to emphasize the delete account button from the rest of the content on the page. But I think it is not consistent with rest of the page,WDYT @vedant-jain03.

PR #2820, opened 2022-01-22
Move to esbuild for entire bundling process

tachyons:
Using webpack with esbuild-loader is still slower than using esbuild for whole process.

Build time:

Webpack with babel : 20 Seconds
Webpack with esbuild-loader: 5.5 seconds
esbuild: 2.5 seconds


vedant-jain03 (approved):

vedant-jain03:
I am using windows to run the CV environment, the JS Bundling is taken a lower time than as compared to webpacker for me.

gr455 (approved):
Works great!

Issue #2787, opened 2022-01-15
Profile picture should get removed instead of the button being selected.


Kartikay304:
**Describe the bug**
Profile picture should get removed instead of the button being selected.

**To Reproduce**
1. Go to 'Dashboard'
2. Click on 'Edit details'                     
3. Click on 'Remove Picture'

**Expected behavior**
Current behavior when the user wants to remove his profile pic :
1. Clicks on the remove profile pic button.
2. The button is selected. (but the pic is not removed).
3. Clicks on save.
4. Now, the pic gets removed.

The correct behavior should be: The pic getting removed in step 2.

**Screenshots**

https://user-images.githubusercontent.com/40604582/149613854-e562949a-d563-4d73-8dfb-029384b12162.mp4

**Are you working on this issue? (Yes/No)**
No

Palaksharma23:
I would love to take up this issue

Kartikay304:
> I would love to take up this issue

Thanks for your initiative but, this issue is only open for beginners and new contributors. You can check other unassigned issues.

mudassir1231:
can i work on this issue

Kartikay304:
> can i work on this issue

Sure, go ahead

dikshashukla455:
Greetings, can I work this issue if there's no problem

arnavchhokrajuit:
Greetings,  I would love to take this issue, if it's no problem.

ashwani3011:
I would love to work on this, should I proceed ahead?

Kartikay304:
@dikshashukla455 @ashwani3011 @arnavchhokrajuit , this issue is already assigned, You can check other unassigned issues

FIZA-khan7:
Hello, dear @Kartikay304 sir, can you plz assign me any good first issue. I didn't find any unassigned good first issue in circuitverse repositories.

Kartikay304:
Hi @FIZA-khan7 ,there is already a PR for this issue. You can go through other issues.

mudassir1231:
#2824

suyashdube:
Can I work on this issue? i am a beginner

vedant-jain03:
@suyashdube It is already been assigned checkout another issues.

ashish960:
@Kartikay304 can i work on this issue

Arijit6258:
@Kartikay304  Is this issue still open and can I work on this?

codewithnadeem14502:
[...]

PR #2824, opened 2022-01-24
remove button removes the profile picture

mudassir1231:
Before : previously the remove button used to be like a check box ,the profile pic is not removed unless we save the changes 
After : but this line of code is better than the previous one, as we click on remove button the profile pic gets removed instantly 
i can send you the working video on slack if you want

https://user-images.githubusercontent.com/88880988/150799190-e37a5e2a-f685-4a7a-bfcd-23e4cfd22ffd.mp4

### DIFF
--- a/app/views/users/circuitverse/edit.html.erb
+++ b/app/views/users/circuitverse/edit.html.erb
@@ -24,7 +24,7 @@
             <label for="profile_picture" class="users-edit-image-button"><%= t("users.circuitverse.edit_profile.choose_file") %></label>
             <input tabindex="0" type="file" accept="image/png, .jpeg, .jpg" class="users-edit-image-button-hidden" id="profile_picture" name="user[profile_picture]" onchange="readURL(this);" value="<%= @profile.profile_picture %>">
             <%= f.text_field :remove_picture, id: "remove_pic", class:"remove-pic",value:"1" %>
-            <label for="remove_pic" class="users-remove-image-button" onclick="document.getElementById('imgedit').src='/assets/thumb/Default.jpg'; document.getElementById('remove_pic').value = '1'"><%= t("users.circuitverse.edit_profile.remove_picture") %></label>          </div>
+            <label for="remove_pic" class="users-remove-image-button" onclick="rempic()"><%= t("users.circuitverse.edit_profile.remove_picture") %></label>
           </div>
           <div class="field form-group">
             <h6><%= f.label :country %></h6>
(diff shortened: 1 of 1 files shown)


vedant-jain03 (changes requested):
After removing the profile pic, you should be on the edit page only so that to edit other details as well.

app/views/users/circuitverse/edit.html.erb
```diff
@@ -23,8 +23,8 @@
             <p><%= image_tag @profile.profile_picture.url(:thumb), id:'imgedit', class:"users-edit-image", alt:'User Profile Image' %></p>
             <label for="profile_picture" class="users-edit-image-button"><%= t("users.circuitverse.edit_profile.choose_file") %></label>
             <input tabindex="0" type="file" accept="image/png, .jpeg, .jpg" class="users-edit-image-button-hidden" id="profile_picture" name="user[profile_picture]" onchange="readURL(this);" value="<%= @profile.profile_picture %>">
-            <%= f.check_box :remove_picture, id: "remove_pic", class:"remove-pic" %>
-            <label for="remove_pic" class="users-remove-image-button"><%= t("users.circuitverse.edit_profile.remove_picture") %></label>
+            <%= f.text_field :remove_picture, id: "remove_pic", class:"remove-pic",value:"1" %>
+            <label for="remove_pic" class="users-remove-image-button" onclick="document.getElementById('imgedit').src='/assets/thumb/Default.jpg'; document.getElementById('remove_pic').value = '1'"><%= t("users.circuitverse.edit_profile.remove_picture") %></label>          </div>
```
Kartikay304, line 27: Remove the div tag, which is at the end of this line. And I think the js code should be under a function to maintain code readability.

Kartikay304 (changes requested):

vedant-jain03 (approved):
Great work @mudassir1231 , LGTM, Let's wait for maintainers review.

Kartikay304 (approved):

Kartikay304:
#2787

gr455:
Extremely sorry for a late review. Will review this asap

app/views/users/circuitverse/edit.html.erb
```diff
@@ -77,8 +77,13 @@
       .attr('src', e.target.result);
     };
     reader.readAsDataURL(input.files[0]);
+    document.getElementById("remove_pic").value = "0";
     }
   }
+  function rempic(){
```
tachyons, line 83: This won't remove the pic if the user click on the button and refresh the page, also the image path given below won't work in production

vedant-jain03: I have suggestion. What if we can implement a popup for profile image similar as we have in social media.

We can remove the scope of editing profile image from edit.

WDYT?

aryanndwi123:
Closing the issue !!! As this is a stale PR !!! and has merge conflicts 🍡

PR #2921, opened 2022-02-11
Fix Bug Related to the Splitter and its Prompt Window

### LINKED ISSUE #2920: Project Breaking Problem/Bug Caused by Splitter Element
**Describe the bug**
Clicking cancel in splitter prompt window causes an error to occur. This ends up breaking the project, and if saved, the project will never be able to load.

**To Reproduce**
1. Launch a project
2. Click on Circuit Elements > Misc > Splitter
3. Click cancel through the two prompt windows
4. See errors in dev tools' console
5. Try to save project online
6. Unable to save; if you are, project becomes corrupted

**Expected behavior**
In the simulator, there's no immediate indicator that error occurred. However, moving and placing elements become very troublesome to do. Expected behavior should be that it properly cancels the creation of the splitter and causes no problems.

**Screenshots**
bug

**Device Information [optional]:**
 Tried the steps on Chrome, Safari, and Firefox and all show similar errors in the console.

**Additional context**
Problem needs to be fixed or else users' projects may get corrupted without them knowing; resulting in so much hours of work lost.

Yes

Zhongy1:
Fixes #2920

- Fixed a bug that prevents the Splitter element from properly getting created when a user cancels the prompt window.
- As shown by the screenshot, the change will default to a `1` input if a user were to cancel the prompt window.
- A few whitespace character changes in the code

Screen Shot 2022-02-10 at 7 53 23 PM

### DIFF
--- a/simulator/src/modules/Splitter.js
+++ b/simulator/src/modules/Splitter.js
@@ -38,7 +38,7 @@ export default class Splitter extends CircuitElement {
         */
         this.rectangleObject = false;
 
-        this.bitWidthSplit = bitWidthSplit || (prompt('Enter bitWidth Split') || `${'1 '.repeat((bitWidth || 1) - 1)}1`).split(' ').filter((lambda) => lambda !== '').map((lambda) => parseInt(lambda, 10) || 1);
+        this.bitWidthSplit = bitWidthSplit || (prompt('Enter bitWidth Split') || `${'1 '.repeat((this.bitWidth || 1) - 1)}1`).split(' ').filter((lambda) => lambda !== '').map((lambda) => parseInt(lambda, 10) || 1);
         this.splitCount = this.bitWidthSplit.length;
 
         this.setDimensions(10, (this.splitCount - 1) * 10 + 10);
--- a/simulator/src/modules/Splitter.js
+++ b/simulator/src/modules/Splitter.js
@@ -38,7 +38,7 @@ export default class Splitter extends CircuitElement {
         */
         this.rectangleObject = false;
 
-        this.bitWidthSplit = bitWidthSplit || (prompt('Enter bitWidth Split') || '1').split(' ').filter((lambda) => lambda !== '').map((lambda) => parseInt(lambda, 10) || 1);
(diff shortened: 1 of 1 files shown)


vedant-jain03:
@gr455 Any Views here?

simulator/src/modules/Splitter.js
```diff
@@ -38,7 +38,7 @@ export default class Splitter extends CircuitElement {
         */
         this.rectangleObject = false;
 
-        this.bitWidthSplit = bitWidthSplit || prompt('Enter bitWidth Split').split(' ').filter((lambda) => lambda !== '').map((lambda) => parseInt(lambda, 10) || 1);
+        this.bitWidthSplit = bitWidthSplit || (prompt('Enter bitWidth Split') || '1').split(' ').filter((lambda) => lambda !== '').map((lambda) => parseInt(lambda, 10) || 1);
```
gr455, line 41: Thank you for this, I have a few concerns with this
I am not sure if this is helpful. If the splitter has a bitwidth of `n`. Then we want all splits to add up to `n`. If this does not happen, then the splitter behaves in an unsoundly.

You can maybe change this to have `'1'.repeat(bitWidth)`?

Zhongy1: Thanks for the advice. I went and changed it to `'1 '.repeat((bitWidth || 1) - 1) + '1'`. Basically, this will give it the default behavior of splitting a bus of size `n` to `n` number of splits if user chooses to cancel the prompt or enters nothing.

Zhongy1:
@vedant-jain03 @gr455 Can I get a status on this?

gr455:
> @vedant-jain03 @gr455 Can I get a status on this?

Oh so sorry, I missed it. I'll review it now

Zhongy1:
@gr455

Zhongy1:
@gr455 Review request reminder

Zhongy1:
@vedant-jain03 This PR is in need of a little more love. Btw, I see AppVeyor build failed, but I don't think I've done anything that should cause it to fail. Is that going to be a problem?

Zhongy1:
@gr455

gr455:
> @gr455

Wow, I am soo sorry for such a late review. I have to do many pending reviews :sweat_smile: 

Looks good. There are a few more checks that we should do in splitter but for now, this solves the problem!
Thanks a lot!!

gr455 (approved):

PR #2961, opened 2022-02-26
Add: Search using the part/component/element number

### LINKED ISSUE #2112: introduce refreshToken to keep user logged in from browser
In most of the websites we see after log in for the first time it keeps user in logged in state until and unless any unusual activity happens from user's account but in the CircuitVerse website it not happens , one user has to give his/her credential every time he/she visits the website.

I would like to update the backend side the login api , when one user will login/signup for the first time it will return one ```refreshToken``` along with the ```token``` now we have to store the refreshToken in local storage and everytime user visit the website refreshToken will be exchanged to get a new token + refreshToken and this process continues .
If any unusual activity happens refreshToken will be automatically revoked .

**Describe alternatives you've considered**
Alternative solution is already running in CircuitVerse which is the current login/signup flow which is very fine , I just want to enhance the user experience 

**Additional context**
The proper implementation can be done using the google auth API and here is the reference guide [here](https://firebase.google.com/docs/reference/rest/auth)

the previous login/signup api will remain as it is only change is needed to return the refreshToken in login/signup both cases 
and also we have to write another api for exchanging between refreshToken and token

image

No, but looking for approval from mentors and other contributors , if I get approval then I can implement this

Dnouv:
Fixes #2112

In this PR a new node is added to the `elementHierarchy` Object in addition to the `name` & `label`, also added a new custom filter function to filter the component based on their IC numbers. This enables the user to search the component using the component number.

search ALU using part number
search XNOR GATE using part number

Feedback is most welcomed. Thank you!


vedant-jain03:
@Dnouv kindly remove 2150 issue fixing as the issues mentioned are exact replicas of each other, fixing one of them we can close both.

Dnouv:
Ok, I will edit the comment.
Thank you!

ayan-biswas0412:
@Dnouv can you please provide a video file of the demo?
and also can you please explain the logic involved here?

Dnouv:
@ayan-biswas0412  the basic idea behind the search is, a node `type` is added in the element Object, in the `setup.js` a new list is introduced which contains Object with keys `label` and `type`, this array (list) is further on made accessible to the functions in the `listener.js`. 
In the `listener.js`'s `customFilter` function when the search input value is changed, it compares the changed value with the circuit element names or labels, then if it's true it stores this boolean value in a constant `labelCheck` then proceeding further the changed value is compared with the component numbers stored in the `type` key and the returned boolean value is stored in `typeCheck` variable. Then an OR statement is executed on the `labelCheck` and the `typeCheck` and a value is returned accordingly.

For example, suppose the user enters the value "gate" then  the following line of code returns `true` for every element containing the "gate" word in the label,
```
const labelCheck = ele.toLowerCase().includes(value)
```
since there are none of the components with "gate" in the `type` node, the following code remains `false`
```
let typeCheck = false
```
then it returns `true` from the filter function.

Now a user enters the value "74266" so the `labelCheck` will always remain `false` as there are none matching the searched value and for each circuit element the following line
```
elementType.forEach((inTyp) => {
                if (Object.values(inTyp)[0] == ele && Object.values(inTyp)[1].toLowerCase().includes(value)) {
                    typeCheck = true
                }
            })
```
 compares the searched value with the `type` of the component if it matches then it sets the `labelCheck` as `true` and the filter returns true and the component is added in the `results` array.
Thank you!

simulator/src/metadata.json
```diff
@@ -118,42 +118,42 @@
...
             {"name":"MSB", "label":"MSB(Most Significant Bit)"},
             {"name":"LSB", "label":"LSB(Least Significant Bit)"},
-            {"name":"PriorityEncoder", "label":"Priority Encoder"},
+            {"name":"PriorityEncoder", "label":"Priority Encoder", "type": "IC74147"},
             {"name":"Decoder", "label":"Decoder"}
         ],
         "Sequential Elements": [
-            {"name":"DflipFlop", "label":"D flip Flop"},
-            {"name":"Dlatch", "label":"D latch"},
+            {"name":"DflipFlop", "label":"D flip Flop", "type": "IC74LS74"},
```
ayan-biswas0412, line 145:
```suggestion
            {"name":"DflipFlop", "label":"D flip Flop", "type": "IC7474"},
```
here LS stands for - low power Schottky diode, so the IC number is basically 7474 so please keep it as 7474 only and also change others as well.

ayan-biswas0412 (changes requested):
I don't know if this hardcoding the only ic number as `type` best fits the solution or not, because there may be similar IC numbers that also refers to the same gate/flip-flop but with different IC number, so better if those made an array instead of a single value.

can you guys please have a look @CircuitVerse/issues-team

Dnouv:
Thanks a lot for the review, I have made the required changes, regarding the array instead of `type` how about going with the most used component numbers? 

> ...please have a look https://github.com/orgs/CircuitVerse/teams/issues-team

Sorry but this link is broken could you please share the correct link, once again.
Thank you!

ayan-biswas0412:
> Thanks a lot for the review, I have made the required changes, regarding the array instead of `type` how about going with the most used component numbers?
> 
> > ...please have a look https://github.com/orgs/CircuitVerse/teams/issues-team
> 
> Sorry but this link is broken could you please share the correct link, once again. Thank you!

no no the link is not for you, i just tagged our team such that everyone can give review.

ayan-biswas0412:
> Thanks a lot for the review, I have made the required changes, regarding the array instead of `type` how about going with the most used component numbers?
> 
> > ...please have a look https://github.com/orgs/CircuitVerse/teams/issues-team
> 
> Sorry but this link is broken could you please share the correct link, once again. Thank you!

how to determine which number is most used ?

Dnouv:
Ah, that one by the most I meant something like standard literature, which most of the users follow such as in the following pdf link there are some of the ICs number mentioned. So maybe some standards like this.
https://organization.soe.ucsc.edu/sites/default/files/Digital_IC.pdf
Thank you!

Dnouv:
Hello @ayan-biswas0412 so shall we proceed with the most common names or the array idea? Please let me know.
Thank you!

ayan-biswas0412:
> Hello @ayan-biswas0412 so shall we proceed with the most common names or the array idea? Please let me know. Thank you!

i think you can proceed with the array, because most common may not be common for a particular individual

Dnouv:
Thank you for confirming @ayan-biswas0412 I've got a request if it's ok, shall we continue our conversation on Slack, since there might be many names(code) I find and it might require verification and you could also provide me some codes.
Please let me know then I'll ping you. Looking forward to your reply.
Thank you!

ayan-biswas0412:
> Thank you for confirming @ayan-biswas0412 I've got a request if it's ok, shall we continue our conversation on Slack, since there might be many names(code) I find and it might require verification and you could also provide me some codes. Please let me know then I'll ping you. Looking forward to your reply. Thank you!

yah you can continue on slack

gr455:
@Dnouv Any updates on using the array for IC numbers?

PR #3011, opened 2022-03-07
Fix nodes overflow when object input size > 5

### LINKED ISSUE #3009: Gates style when no of inputs > 5
**Describe the bug**
The design on the input gate becomes not good looking when the number of inputs to it becomes > 5

**To Reproduce**
1. Go to Simulator
2. Click on Gates
3. Choose AND Gate
4. Make no of inputs > 5
5. See error

**Expected behavior**
The gate should become greater whenever the number of inputs increases, so that the points of the input don't overflow the design

**Screenshots**
a

Yes

ahmedibrahim404:
Fixes #3009 

I made the width/height of the input change w.r.t. input size as shown below

I implemented this only for AND, since it's easy to
If it's okay, I will do it for the rest of the objects

final

### DIFF
--- a/simulator/src/modules/AndGate.js
+++ b/simulator/src/modules/AndGate.js
@@ -39,7 +39,9 @@ export default class AndGate extends CircuitElement {
         this.inputSize = inputLength;
 
         this.yDim = 20 + (this.inputSize - 2) * 4;
-        this.xDim = 15 + (this.inputSize-2) * 5;
+        this.xDim = 15 + (this.inputSize - 2) * 5;
+
+        // this is added due to small shifts in the output node, and the cursor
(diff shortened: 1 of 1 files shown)


gr455:
This might break existing circuits (the nodes are moved around and might go over other nodes), so it doesn't seem worth doing

ahmedibrahim404:
> This might break existing circuits (the nodes are moved around and might go over other nodes), so it doesn't seem worth doing

Yeah, right
I think this could be done later, when there is a version control
Thanks

gr455:
> > This might break existing circuits (the nodes are moved around and might go over other nodes), so it doesn't seem worth doing
> 
> Yeah, right I think this could be done later, when there is a version control Thanks

Yep! Keep the issue open then, you can close the PR

PR #3086, opened 2022-04-26
Added Pretty Url for Users

### LINKED ISSUE #2055: Pretty URL for Users
Having pretty URLs is a must for a website. I personally, don't like seeing random numbers appear in the address bar. It would be better if we have pretty URLs for a user.

`/users/123434234` can be replaced by `/users/:user-name` by adding a slug to users' schema or by using [friendly_id](https://github.com/norman/friendly_id) which has already been used a few times in this project.

**Additional context**

 _Before_
> before

_After_
> 
Screenshot from 2021-01-03 01-00-30

Yes

Prerna-0202:
Fixes #2055 

I have added the feature "Pretty Url for Users" as per issue #2055 

**Before:-**
Screenshot (340)

**After:-**
Screenshot (339)

### DIFF
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -44,7 +44,7 @@ FriendlyId.defaults do |config|
   # performance because it will avoid Rails-internal code that makes runtime
   # calls to `Module.extend`.
   #
-  config.use :finders
+  # config.use :finders
   #
   # ## Slugs
   #
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -44,14 +44,14 @@ FriendlyId.defaults do |config|
   # performance because it will avoid Rails-internal code that makes runtime
   # calls to `Module.extend`.
   #
-  # config.use :finders
+  config.use :finders
   #
   # ## Slugs
   #
   # Most applications will use the :slugged module everywhere. If you wish
   # to do so, uncomment the following line.
   #
-  # config.use :slugged
+  config.use :slugged
   #
   # By default, FriendlyId's :slugged addon expects the slug column to be named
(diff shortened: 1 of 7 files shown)


config/initializers/friendly_id.rb
```diff
@@ -44,14 +44,14 @@
   # performance because it will avoid Rails-internal code that makes runtime
   # calls to `Module.extend`.
   #
-  # config.use :finders
+  config.use :finders
```
tachyons, line 47: Don't use finders module, it does have side effects

Prerna-0202: Okay, I removed it and checked that it is working fine...

tachyons:
The main reason we kept holding this issue was that urls can be too long as names are not unique, could you check how the URLs looks like there are two people with same name and same project name ?

Prerna-0202:
> The main reason we kept holding this issue was that urls can be too long as names are not unique, could you check how the URLs looks like there are two people with same name and same project name ?

yaa right, URLs are too long for two same users. 
Screenshot (347)
Screenshot (348)
Screenshot (351)
Screenshot (354)

vedant-jain03:
> > The main reason we kept holding this issue was that urls can be too long as names are not unique, could you check how the URLs looks like there are two people with same name and same project name ?
> 
> So, we have to make URLs short for the same name users?

What if we can do this:
if the user has the same name then URL could be `${url}/name-${user.created_at}`
created_at field will be different and it will not make the URL larger as it does in above-mentioned sample.
WDYT? @tachyons

tachyons:
@vedant-jain03 That will make URLs unnecessarily complex, I am closing this for now. We can revisit them once we have unique username kind of mechanism

PR #3138, opened 2022-05-26
Add arm64 support for docker builld

tachyons:
This PR is to add arm64 support for docker image make circuitverse image accessible from Mac M1 and arm based linux distributions


gr455 (approved):
LGTM :+1:

tachyons:
Merging as explicitly specifying --platform seems to have reduced the extra build time. This will get reverted if it was a false positive

PR #3145, opened 2022-05-28
[Infra] Migrate docker to alpine base image

### LINKED ISSUE #3142: Reduce the size of production docker image
Our current docker image takes around 600MB, we might be able to reduce this by using small base images like alpine.

gr455:
Fixes #3142

This PR migrates the existing docker build to apline based image

Slashes 600 MB from the image size

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,14 @@
-FROM ruby:3.1.2
+FROM ruby:3.1.2-alpine
 
 # set up workdir
 RUN mkdir /circuitverse
 WORKDIR /circuitverse
 
 # install dependencies
-RUN apt-get update -qq && apt-get install -y imagemagick shared-mime-info && apt-get clean
+RUN apk update -qq && apk add imagemagick shared-mime-info postgresql-dev
 
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash \
- && apt-get update && apt-get install -y nodejs && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,17 +1,14 @@
...
 # install dependencies
-RUN apt-get update -qq && apt-get install -y imagemagick shared-mime-info && apt-get clean
+RUN apk update -qq && apk add imagemagick shared-mime-info postgresql-dev openssl
 
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash \
- && apt-get update && apt-get install -y nodejs && rm -rf /var/lib/apt/lists/* \
- && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
- && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
- && apt-get update && apt-get install -y yarn && rm -rf /var/lib/apt/lists/*
+RUN apk --no-cache add nodejs yarn --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community \
```
tachyons, line 10: AFAIK this will always install latest node version, can we get more control on this or use latest LTS version here. Because latest version node can have incompatibilities with the packages we use

gr455: The alpine repository only has the edge channel releases it seems
https://pkgs.alpinelinux.org/packages?name=yarn

tachyons: We may have to find better alternatives, merging this PR for now

PR #3148, opened 2022-05-30
fix: import theme file breaks custom theme dialog

### LINKED ISSUE #2923: Importing theme file breaks customTheme dialog
**Describe the bug**
When we import a custom Theme file, we cannot change the color Input values

**To Reproduce**
1. Go to circuitverse simulator
2. Click on tools > themes > customThemes
3. Import a theme file
4. try to change color
5. see error

**Expected behavior**
After importing the theme file, the user is able to change input color values

Jaagrav:
Fixes #2923 

#### Changes I made in this PR -
The issue was caused because the event listener responsible to update the colors/theme on the simulator page was stopping to work after importing a json file for setting a custom theme. Therefore I added the event listener inside another function which is called automatically when the exported function `CustomColorThemes` is called in order to set the event listener. Now I am using this same function signature to call and reset the event listener every time a new custom theme file is imported, inside the function `receivedText`

### Screencast -
Basically proof that it works...

https://user-images.githubusercontent.com/52719271/170951656-2c6edba4-5d4d-416f-b6a2-4990f2db715f.mp4

### DIFF
--- a/simulator/src/themer/customThemer.js
+++ b/simulator/src/themer/customThemer.js
@@ -1,3 +1,5 @@
+/* eslint-disable import/prefer-default-export */
+/* eslint-disable import/no-cycle */
 import { dots } from '../canvasApi';
 import themeOptions from './themes';
 import updateThemeForStyle from './themer';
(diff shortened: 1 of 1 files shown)


pulkit-30 (approved):
Hey @Jaagrav, Welcome to CircuitVerse 🎉 
Thank you so much. This looks great!

gr455 (approved):

PR #3179, opened 2022-06-11
Revert "[Infra] Migrate docker to alpine base image"

tachyons:
Reverts CircuitVerse/CircuitVerse#3145 as it seems to have broke some workflows


gr455:
Aw man, it seems like the workflow is deleting the tagged image before it's used in other builds

PR #3188, opened 2022-06-15
fix: setup documentation

vedant-jain03:
- Updated `setup.md` for gitlab.


tachyons:
Could you update the text as per language cheker bot

vedant-jain03:
> Could you update the text as per language cheker bot

the second check is not needed as it is a callback URL that needs to be `github` only!
Updated first!

PR #3242, opened 2022-07-18
fixes : embed feature broken

vedant-jain03:
Done by @pulkit-30 @gr455 @tachyons

### DIFF
--- a/app/views/layouts/simulator.html.erb
+++ b/app/views/layouts/simulator.html.erb
@@ -1,4 +1,3 @@
-<!DOCTYPE html>
 <html>
   <head>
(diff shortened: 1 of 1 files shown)


pulkit-30 (approved):
great 🎉

tachyons:
Good catch, but removing this line means circuit will render in compatiability mode, not proper html mode, we may have to create an issue to address the actual issue as well

vedant-jain03:
> Good catch, but removing this line means circuit will render in compatiability mode, not proper html mode, we may have to create an issue to address the actual issue as well

Yaa, totally agree. It is just a temporary solution!

PR #3276, opened 2022-09-01
fix: dependent notifications on project while destroy

### LINKED ISSUE #3275: fix: ActionView::Template::Error: Couldn't find Project with 'id'
**Describe the bug**
New unhandled error in production from [circuitverse](https://app.bugsnag.com/circuitverse/circuitverse?i=sk&m=nw) in projects#destroy [(details)](https://app.bugsnag.com/circuitverse/circuitverse/errors/630f03a63d053400098bc0bf?event_id=630f03a6009657c426200000&i=sk&m=nw)
Unhandled error
ActionView::Template::Error: Couldn't find Project with 'id'=792042
Location
app/notifications/fork_notification.rb:8 - message

Yes

vedant-jain03:
Fixes #3275 

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -9,6 +9,7 @@ class ProjectsController < ApplicationController
 
   before_action :check_access, only: %i[edit update destroy]
   before_action :check_delete_access, only: [:destroy]
+  before_action :filter_dependent_notifications, only: [:destroy]
   before_action :check_view_access, only: %i[show create_fork]
   before_action :sanitize_name, only: %i[create update]
   before_action :sanitize_project_description, only: %i[show edit]
@@ -142,6 +143,15 @@ class ProjectsController < ApplicationController
 
     def sanitize_name
       params[:project][:name] = sanitize(project_params[:name])
+    end
+
+    # destroy all dependent notifications
+    def filter_dependent_notifications
+      NoticedNotification.find_each do |notification|
+        if notification.params[:project_id] == @project.id
+          notification.destroy
+        end
+      end
     end
 
     # Sanitize description before passing to view
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -9,7 +9,6 @@ class ProjectsController < ApplicationController
 
   before_action :check_access, only: %i[edit update destroy]
   before_action :check_delete_access, only: [:destroy]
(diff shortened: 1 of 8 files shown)


app/controllers/projects_controller.rb
```diff
@@ -144,6 +145,15 @@ def sanitize_name
       params[:project][:name] = sanitize(project_params[:name])
     end
 
+    # destroy all dependent notifications
+    def filter_dependent_notifications
+      NoticedNotification.find_each do |notification|
+        if notification.params[:project_id] == @project.id
```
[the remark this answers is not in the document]
tachyons: This is super inefficient, just add dependent: destroy to notifications associations in the project model

vedant-jain03:
@tachyons Kindly Review
`dependent: true` default in `has_noticed_notification` helper association!

vedant-jain03:
## What I did:
- Initially, we were storing `project_id` which was causing an error in destroying the dependent notifications on deleting the project, so now we are storing the object with the param name `project`.
- The data migration for old notifications has not been run yet, so modified that file too!
- Added new data migration for existing `noticed_notifications` where the notifications for which Project does not exist will get deleted and for existing will update the params with the `Project` and `user` object.

@tachyons @abstrekt @pavanjoshi914

app/controllers/users/noticed_notifications_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
   def mark_as_read
     notification = NoticedNotification.find(params[:notification_id])
     notification.update(read_at: Time.zone.now)
-    project = Project.find(notification.params[:project_id])
+    project = notification.params[:project]
```
tachyons, line 14: What is the difference does this change make ?

tachyons, line 14: What is the difference does this change make ?

vedant-jain03: So previously we were finding the record using `project_id` in `Project` but now the object is already saved in `params[: project]` so there is no need to find it and complicate it once again!

vedant-jain03:
@tachyons we can merge this!
The error in code climate is recursive, even if we merge the `if` in a single line, the line is becoming long and it is suggested to use the `block` statement only!

tachyons:
@vedant-jain03 Then disable specific rubocop cop with rubocop:disable command

vedant-jain03:
@tachyons Kindly Review!

PR #3480, opened 2023-01-09
search item was not is in line with respect to each other

weberankit:
Fixes #search item was not  in line with respect to each other

bug--
bug-1
 bug-fixed-

bug-fixed

### DIFF
--- a/app/assets/stylesheets/search.scss
+++ b/app/assets/stylesheets/search.scss
@@ -199,3 +199,8 @@
     width: 90%;
   }
 }
+
+.search-user-container .center-row{
+display: flex;
+flex-direction: column;
+}


vedant-jain03:
@weberankit Thanks for the PR, could you open an issue for it!

app/assets/stylesheets/search.scss
```diff
@@ -199,3 +199,8 @@
     width: 90%;
   }
 }
+
+.search-user-container .center-row{
```
vedant-jain03, line 203: You can use bootstrap class for this: `flex-column`

vedant-jain03 (changes requested):

app/assets/stylesheets/search.scss
vedant-jain03: Just add `flex-column` class!

vedant-jain03: @weberankit

ThatDeparted2061:
@Arnabdaz the issue has been resolved

PR #3483, opened 2023-01-09
feat: comment and new thread notification

### LINKED ISSUE #3470: feat: attach notification events to forum threads and comments

Building notification events for forum (simple_discussion) for which the user has subscribed. 

**Additional context**

Yes, currently working on it.

VaibhavUpreti:
Signed-off-by: VaibhavUpreti <vaibhav.upreti16@gmail.com>

Fixes #3470

Added two notification events for forum

1. Forum Comment notification:
-  for all the users that have subscribed to the thread.
- eg: Users that have subscribed to a conversation receive notifications below...
conversation
convo
notifications
b
a

2. New Thread notification:
-  for all admin users, moderators so that they can resolve the query asap.
- eg: ref, both the admin user receive notification when `user1` posted a new thread.
admin-thread
meow-thread

### Redirects 
1. Comment notification redirects to the comment
after-redirect
2. Thread notification redirects to the thread

 Have disabled checks for files that were directly added from simple discussion(as they had many robocop violations... resolved most of them... still some are left).
### Future Scope
To add a central notification preference (new_thread_notification )for all users (not just admin users) to receive notification on a new thread... default false .... In customise event notification PR


config/locales/views/users/en.yml
```diff
@@ -91,4 +91,6 @@ en:
       see_all : "See All"
       fork_notification: "%{user} forked your Project %{project}"
       star_notification: "%{user} starred your Project %{project}"
-      new_assignment_notification: "You have been assigned a new assignment '%{assignment_name}'"
\ No newline at end of file
+      new_assignment_notification: "You have been assigned a new assignment '%{assignment_name}'"
+      forum_thread_notification: "%{user} posted a new thread '%{thread}'"
+      comment_notif: "%{user} commented '%{mssg}' on thread '%{thread}'"
```
vedant-jain03, line 96:
```suggestion
      comment_notif: "%{user} commented on thread '%{thread}'"
```

The message could be large. Let's keep it just to notify you if there is a new comment on the thread!

app/notifications/forum_comment_notification.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def message
+    user = params[:user]
+    post = params[:forum_post]
+    thread = params[:forum_thread]
+    t("users.notifications.comment_notif", user: user.name, thread: thread.title, mssg: post.body.truncate_words(4))
+  end
+
+  def icon
+    "fa fa-comment"
```
vedant-jain03, line 14:
```suggestion
    "far fa-comment"
```
Let's use this icon with `padding: 12px;` style.

vedant-jain03 (changes requested):
Awesome, added my initial reviews!

config/locales/views/users/en.yml
VaibhavUpreti: Currently its truncated to merely 4 words [permlink](https://github.com/CircuitVerse/CircuitVerse/pull/3483/files#diff-2ba232ca3549893005e0136addf5f50454622b5f49ddfc0bf88d98f7d2d603feR10).
Is that limit good or should I change it? WDYT @vedant-jain03

vedant-jain03: I think that notifying users regarding events is the primary task, WDYT?
What if the thread heading is so large?

VaibhavUpreti: Yes you are right, Thread heading should be truncated too... because questions would be generally longer.

For mssg, it would look something like this. (adding screenshots... will update in next commit).
no-comment

comment

app/models/forum_post.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ForumPost < ApplicationRecord
```
tachyons, line 3: Just override the relevant methods, no need to override entire methods, refer https://guides.rubyonrails.org/engines.html#overriding-models-and-controllers

Issue #3621, opened 2023-02-27
[ui] : Remove extra space above footer.


vedant-jain03:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

There is an additional space above the footer.

![Screenshot from 2023-02-27 17-35-05](https://user-images.githubusercontent.com/76901313/221559538-b7d16777-0b49-4a01-899a-4e354812c59a.png)

**Key Deliverables:**
- Remove this extra space

### Additional context

This feature is under the fosshack, you need to register [here](https://fossunited.org/login#login). You can join #fosshack-students channel in slack for any doubts and support

### Are you working on this issue?

No

thisiskaransgit:
Hey, I would like to work on this issue

vedant-jain03:
@thisiskaransgit Go ahead!

Aniket1299:
Hi guys, instead of removing the footer div completely, would it be good if we reduced the size to 25px? Looks neat.

ShivamKaushik316:
Greetings ..  I am new to Open Source Contribution . Can I work on this issue?

Gourav2580:
Hi I am  new open sorece contribution please help me to understande this issue.

ShivamKaushik316:
Bro You may find these funny but I myself started yesterday only 🤣 So if u
find some help plz let me know as well

ShivamKaushik316:
But if u want to understand this issue...the problem was in the UI . The
space between footer and section above footer was very much (if u look at
my pull request , u can see the before and after pics as well).

mohitboyidapu:
Hey @vedant-jain03, Is this issue still open? If it is open I would like to work on this issue.

B1TSH3LL:
Hey @vedant-jain03 i want to contribute to this issue. Can you please assign this to me?

Rishabh-mikku:
I want to contribute on this project, can you please assign this to me.

rishuraj1:
Hey there, I believe this issue should be fixed using the above fix #3700 . Thanks :)

mandliyarajendra11:
is open ???

siddharthiitian:
@vedant-jain03  is this issue still open ?

divyanshu255:
i want to work on this issue ,please assign it to me.

DkRadhey:
Hey @vedant-jain03  is this issue still open. If yes, could you please assign this to me.

parthksingh1:
[...]

PR #3624, opened 2023-02-27
feat: updated footer component

thisiskaransgit:
Fixes #3621 

Removed Extra Space from the footer component

### DIFF
--- a/app/assets/stylesheets/footer.scss
+++ b/app/assets/stylesheets/footer.scss
@@ -1,5 +1,5 @@
 .footer-empty-div {
-  height: 250px;
+  height: 25px;
 }
 
 .footer-container-fluid {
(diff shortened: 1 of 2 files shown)


vedant-jain03:
cc @Prerna-0202

Prerna-0202:
After completely removing the extra space, the footer is looking like this. I think we can just decrease this space instead of removing the tag.
WDYT @vedant-jain03 @thisiskaransgit ?
Screenshot (1202)

thisiskaransgit:
Hey @Prerna-0202, so rather than removing the whole container, I've reduced the size of it to 25px, hope this will suffice the need

app/assets/stylesheets/footer.scss
```diff
@@ -1,5 +1,5 @@
 .footer-empty-div {
-  height: 250px;
+  height: 25px;
```
vedant-jain03, line 2:
```suggestion
  height: 50px;
```

vedant-jain03 (changes requested):

VaibhavUpreti:
Fixed via https://github.com/CircuitVerse/CircuitVerse/pull/5278

PR #3629, opened 2023-03-01
Removed extra space above Footer

ShivamKaushik316:
.footer-empty-div now have the height from 250px to 40px

Fixes :  Access Empty Space above the Footer has been reduced to 40px

Before : 

before

After:

after


naman79820:
no issue attached

PR #3663, opened 2023-03-11
chore: bump mailkick gem to v1

### LINKED ISSUE #3662: chore: Bump mailkick gem to v1
## Is your feature request related to a problem? Please describe

Mailkick gem needs to be upgraded to latest v1. Some major changes have been included in the newer version and many methods like `mailkick_user`, `subscribed?`,etc are now deprecated.

reference [link](https://github.com/ankane/mailkick#upgrading)

## Are you working on this?

Yes

VaibhavUpreti:
Signed-off-by: VaibhavUpreti <vaibhav.upreti16@gmail.com>

Fixes #3662 

Upgraded mailkick gem to v1 [Reference](https://github.com/ankane/mailkick#upgrading)

Updated the following code: 
- model
- spec
- views
- added migration
- data_migration for production data

### Testing mails after upgrading to v1
 
https://user-images.githubusercontent.com/85568177/224471115-2698430b-9f34-4550-afdf-1898fcd3156e.mp4

### Only Subscribed Users are sent mails

https://user-images.githubusercontent.com/85568177/225252736-eeda6ede-8019-4628-9aee-2e9ed603c33e.mp4

### Data Migration

https://user-images.githubusercontent.com/85568177/224471132-6a659149-ef64-4844-996e-5a547b5e0941.mp4

### Test Cases

test

## Final Step

Run a migration to drop table mailkick_optouts after we have made sure all previous data is populated in new table


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
```
vedant-jain03, line 1: @VaibhavUpreti we need to update the `mailkick` and it's dependent gems only!
@tachyons

vedant-jain03 (changes requested):

Gemfile.lock
VaibhavUpreti: Sure, will resolve it by EOD!

VaibhavUpreti:
It appears that the task is completed. Earlier, the gem stored the details of users who chose to not receive emails in the `mailkick_opt_outs` table. However, the newer version now stores details of users who choose to receive emails in the `mailkick_subscriptions` table [reference](https://github.com/ankane/mailkick#upgrading), and the migration will populate the data of all such users.

Thus subsequently we create subscriptions whenever a new user is created and update the records accordingly.

app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module MailerHelper
+  def mail_if_subscribed(recipient_email, subject, user)
```
tachyons, line 4: Can we use [before_action](https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks) as most of our email need to have this check ?
```suggestion
  def mail_if_subscribed(recipient_email, subject, user)
```

VaibhavUpreti: Yup, we use this for 3 mailer classes, I was earlier thinking of defining a method in each block or using this helper method. Will update in next commit 👍

VaibhavUpreti: @tachyons looks like before_action approach wont work here, unable to get user through params nor can we pass params in before action .

If we use a private method to check this condition then again we will require another conditional inside the mailer methods.

So I think the former approach, before this helper method one is better. I have updated approach like it was done before.

PR #3800, opened 2023-06-14
Fix erb_lint #3711

tanmoysrt:
Fixes #3711 

Fixed the lining issues

### DIFF
--- a/app/views/editor/_index.html.erb
+++ b/app/views/editor/_index.html.erb
@@ -1,6 +1,6 @@
 <textarea id="trumbowyg"> <%= content %> </textarea>
 
-<script type="text/javascript" nonce="<%= request.content_security_policy_nonce %>">
+<%= javascript_tag type: 'text/javascript', nonce: true do -%>
   $(document).ready(function() {
     $('#trumbowyg').trumbowyg({
       svgPath: "https://cdnjs.cloudflare.com/ajax/libs/Trumbowyg/2.20.0/ui/icons.svg",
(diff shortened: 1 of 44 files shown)


app/views/editor/_index.html.erb
```diff
@@ -1,6 +1,6 @@
 <textarea id="trumbowyg"> <%= content %> </textarea>
 
-<script type="text/javascript">
+<script type="text/javascript" nonce="<%= request.content_security_policy_nonce %>">
```
tachyons, line 3: Can we extract this into a helper? 

<script type="text/javascript" <%= default_nonce_tag %> >

Or use [javascript_tag](https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html) like this


<%= javascript_tag type: 'application/javascript' do -%>
  alert('All is good')
<% end -%>

tanmoysrt: ok! 
Will replace that with `javascript_tag` if no problem arise.

tanmoysrt:
@tachyons, it's now fixed.

I want to mention some issues.
a. In `/app/views/editor/_index.html.erb` there is some regex , which erb_lint can't parse, so I have excluded this file
b. In `/app/views/assignments/_form.html.erb` and `/app/views/about/index.html.erb` ,also has been excluded for checks. if there is `for(var i=0; I<n; i++)` inside `javascript_tag` , erb_lint is warning

---
Can't use something like ``` <%= default_nonce_tag %>``` what you have mentioned before, because erb_lint either search for `javascript_tag` / `javascript_include_tag` with nonce: true or `script` tag with nonce: reqeust.content_security_policy_nonce

Issue #3816, opened 2023-06-21
Integrate Undercover CI


tanmoysrt:
Integrate Undercover CI to check untested code of commit

PR #3818, opened 2023-06-22
Fix CI issue of Docker Build in Test Environment

tanmoysrt:
#### Issue
After the recent merge of this PR #3816  , the CI checks failing to build Docker Image for test environment
undercover require rugged gem and rugged gem required `cmake` at installation time

- Modified `Dockerfile` to install cmake and uninstall it after running `bundle install --without production`
- Move `undercover` from `development` group to `test`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ RUN gem install bundler
 RUN bundle install  --without production
 RUN yarn install
 
-# Remove CMAKE
 # copy source
 COPY . /circuitverse
 RUN yarn build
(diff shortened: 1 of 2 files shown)


tanmoysrt:
cc @tachyons

Dockerfile
```diff
@@ -22,6 +23,9 @@ RUN gem install bundler
 RUN bundle install  --without production
 RUN yarn install
 
+# Remove CMAKE
+RUN apt-get purge -y --auto-remove cmake
```
tachyons, line 27: Keep this as this won't save space anyway, also create one more docker layer for no good reason. Could you check if the same is needed for the production docker file?

Dockerfile
```diff
@@ -22,6 +23,9 @@ RUN gem install bundler
...
+
```
tachyons, line 28:
```suggestion
```

Dockerfile
```diff
@@ -22,6 +23,7 @@ RUN gem install bundler
 RUN bundle install  --without production
 RUN yarn install
 
+# Remove CMAKE
```
tachyons, line 26:
```suggestion
```

Dockerfile
tanmoysrt: for production this will not required `rugged` is only using by `undercover` gem, so in production environment the gem is not going to be installed

PR #3835, opened 2023-06-26
Unit test for models

tanmoysrt:
- Added some Unitests for models


vedant-jain03 (approved):
LGTM 🚀

spec/models/announcement_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      announcement.save
+
+      expect(announcement.exists?).to be true
+    end
+
+    it "returns false for a new announcement" do
+      announcement = described_class.new
+
+      expect(announcement.exists?).to be false
+    end
```
tachyons, line 38: This is a rails method, no need to test this

spec/models/announcement_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
```
tachyons, line 39: We don't need to test this method

PR #3837, opened 2023-06-26
Test undercover

tachyons:


tachyons:
@tanmoysrt Seems like undercover is not working as expected

tanmoysrt:
> @tanmoysrt Seems like undercover is not working as expected

Will check out

tanmoysrt:
@tachyons
1.  I have done test #3838 and reviewdog analysis working. I think with a blank function, it will not give any issue. 
2. It seem to be that the github secret has not PR review permission. Please check the CI log of #3838
image

tanmoysrt:
Also, defining permissions in workflow does not work. The maximum permission is read for all.
So, at the administration level, we need to enable the` write` permission of `GITHUB_TOKEN` [workflow] for the `pull-request` type.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

PR #3860, opened 2023-07-01
feat: Serve assets using active storage

### LINKED ISSUE #3765: step2: Serve assets using ActiveStorage
After Validating all assets have migrated to S3 then 

- Use ActiveStorage for serving assets & remove sync, thus update
   - models
   - views
   - controllers
- Add ActiveRecord Validations
- Update test cases
- Archive Paperclip and Carrierwave data.
  
Are you working on this? (Yes/No)
Yes

VaibhavUpreti:
Fixes #3765 

- Added migration to rename `image_preview` records to `circuit_preview`.
- Feature flag at view level, uploads to both configurations for now.


vaibhavupreti:
After merging this change ,For Project API, instead of calling `project.attributes.imagePreview.url` we would have to use something like `project.attributes.imagePreview`

I think we are using projects api only in our flutter app to get image_preview for the circuit

So the above mentioned would have to be changed in our mobile app,  [example- here](https://github.com/CircuitVerse/mobile-app/blob/cba195a0afc143748d18aba92047c96e79960591/lib/ui/views/projects/components/project_card.dart#L98
):

vaibhavupreti:
This PR is ready. cc @tachyons

spec/spec_helper.rb
```diff
@@ -132,6 +132,10 @@
   # include spec utils
   config.include SpecUtils
 
+  config.before(:suite) do
+    Flipper.disable(:active_storage_s3)
+  end
```
tachyons, line 137: Add spec for both cases instead of globally disabling the spec

vaibhavupreti: My bad, this popped up from local testing. Added extensive test cases for both  the cases 👍

VaibhavUpreti:
@tachyons can we merge this, changes are at view level we will still be mirroring both configs, can easily revert using flipper flag.

app/serializers/api/v1/project_serializer.rb
```diff
@@ -41,6 +41,18 @@ class Api::V1::ProjectSerializer
     project.commontator_thread.is_closed?
   end
 
+  # :nocov:
+  if Flipper.enabled?(:active_storage_s3)
```
tachyons, line 45: @VaibhavUpreti  Since this condition is on class level, application reboot is required after changing this flag

app/serializers/api/v1/project_serializer.rb
```diff
@@ -41,6 +41,18 @@ class Api::V1::ProjectSerializer
...
+    attributes :image_preview do |project|
+      if project.circuit_preview.attached?
+        {
+          url: Rails.application.routes.url_helpers.rails_blob_url(project.circuit_preview, only_path: true)
+        }
+      end
+    end
+  end
```
tachyons, line 53:
```suggestion

    attributes :image_preview, if: -> {  Flipper.enabled?(:active_storage_s3) },  do |project|
      if project.circuit_preview.attached?
        {
          url: Rails.application.routes.url_helpers.rails_blob_url(project.circuit_preview, only_path: true)
        }
      end
  end
```

PR #3894, opened 2023-07-18
Github Codespace as Remote Development Environment

tanmoysrt:
- Introduce Github Codespaces as Remote Development Environment
  - Includes all the features same as Gitpod Remote Development 
  - Ruby Debugger supported


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    ports:
+      - "3000:3000"
+    depends_on:
+      - db
+      - redis
+    environment:
+      REDIS_URL: "redis://redis:6379/0"
+      CIRCUITVERSE_USE_SOLR: "false"
+      DOCKER_ENVIRONMENT: 'true'
+    command: sleep infinity
```
tachyons, line 21: Why do we need this command?

tanmoysrt: As we are using docker-compose with devcontainer, The primary process will attach to the service after it is created. Devcontainer will attach to the service and run the scripts and manage it. To prevent it from terminating before it can attach,  we are using `sleep infinity`.

tanmoysrt: @tachyons also official documentation suggests to use some sleep command to prevent it from terminating
https://code.visualstudio.com/docs/devcontainers/create-dev-container#_extend-your-docker-compose-file-for-development

PR #3913, opened 2023-07-21
Docker based development improvement

### LINKED ISSUE #3914: Imporve  Development Experience with Docker
Improvement that can be achieved :
- Sidekiq, Asset build to run as a separate container currently, which can be run with Foreman in the main container
- Dockerfile consists the steps to prepare full image including install gems, node module installation, which means any changes force to rebuild the image which take lot of time. So we should use Docker for building environment [using same concept of Remote Development Tools like Codespaces / Gitpod]
- As we run the container as the root user and access files from host via volume mounting that results in various Permission issues.
- Currently, after each update on gemfile, node module we need to rebuild image, which takes again much time.We should make it easy to stop the server and give user ability to run commands inside container without stopping the container.
- Gem installation usually lost while recreating the image
- Not possible to restart server without restart the container
- Not easy terminal access to perform various actions

tanmoysrt:
Fixes #3914 

- Dockerfile to serve as an environment [Like another remote development platform]
- Run the application with a non-root user with the same uid and gid as the host machine to resolve any Permission related issue. Also, the non-root user has been given passwordless sudo access to use if required,
- Persist PostgreSQL data using named volume
- Persist ruby bundle data using named volume
- Run setup scripts after the container become ready
- Move to foreman to have an identical setup
- The developer can restart the web service without restarting the container
- Provide bash access to the developer to interact as a regular os terminal
- Ruby debugger, Vite, Solargraph, and Siidekiq is working fine with the current setup
- The developer can do both unit testing and integration testing inside the docker environment

---

Run `bin/docker_run` to start the app in docker `[Linux & Mac]`
Run `bin/docker_run.ps1` to start in `Windows`


VaibhavUpreti (commented):
Code is impressive 🚀  but just like Aboobacker highlighted during the meeting, I am also in favour of having these services(sidekiq, etc) in separate containers instead of a single one.

tanmoysrt:
> Code is impressive 🚀  but just like Aboobacker highlighted during the meeting, I am also in favour of having these services(sidekiq, etc) in separate containers instead of a single one.

I am looking at that , there is a small problem with solargraph while splitting in service. I am trying out workaround to fix that

tanmoysrt:
@tachyons updated the configuration with seperate container for each service.Kindly review this PR

VaibhavUpreti (approved):

PR #3955, opened 2023-08-18
feat: add rbs for models

tanmoysrt:
- Removed `rbs_rails` & keep `rbs` only
- Updated steep gem
- Supress NoMethodError to raise warnings if only signature mismatch
- Added RBS for models


tanmoysrt:
Can you review this PR ? I have only added signatures for defined functions only and this will not raise error if it didn't find any method, So we dont need the auto generated stuffs.
If this PR look fine, then I can proceed to add rbs for other part of codebase

cc @tachyons @smritigarg @vedant-jain03

sig/app/models/collaboration.rbs
```diff
@@ -0,0 +1,2 @@
+class Collaboration
+end
```
vedant-jain03, line 2: Just a little concern, do we need these empty class?

vedant-jain03 (changes requested):

sig/app/models/collaboration.rbs
tanmoysrt: Assume, From a method of User class it's returning Array[Project]. 
For the same purpose, We can keep this class , so that we can use it later for return  type of other methods.
Maybe in the models RBS file this class is not referred anywhere. but in other part of the source code , it can be referred. So i think, we need to keep that

tachyons: @tanmoysrt Add empty classes when needed, remove empty classes which aren't referenced yet

tachyons:
> Supress NoMethodError to raise warnings if only signature mismatch

Could you explain this part ?

tanmoysrt:
> > Supress NoMethodError to raise warnings if only signature mismatch
> 
> Could you explain this part ?

Steep normally check each point of code and if it found any class, which we have reference but no method defined for it , that will raise NoMethodError. Now the ActiveModel has many functions like find, where, exists [many functions that was actually auti genrated in previous PR]. So, if it found an class User and it does not find that class as that is not defined in our rbs files, that will throw Error

When we supress this `NoMethodError` , it will not give us this error any more, it will only raise Error if the Type Signature does not match

tanmoysrt:
@tachyons pls take a look at this pr. If it's completely fine, I can add RBS for some other part of codebase.

tanmoysrt:
@tachyons pls take a look at this pr. If it's completely fine, I can add RBS for some other part of codebase.

Me-Priyank:
never merged, 3+ years stale, zero YARD annotations in current main

PR #3961, opened 2023-08-19
chore: disable generating spans for default settings

VaibhavUpreti:

By default, we should refrain from generating spans when there is no exporter configured, as this will lead to  higher resource consumption.

spans


config/initializers/opentelemetry.rb
```diff
@@ -1,24 +1,26 @@
 # frozen_string_literal: true
 
-require "opentelemetry/sdk"
-require "opentelemetry/exporter/otlp"
-require "opentelemetry/instrumentation/all"
+if Flipper.enabled?(:opentelemetry)
```
VaibhavUpreti, line 3: Should we opt for an ENV variable or a Feature flag? What could be the optimal approach to handle this, Aboobacker?

PR #3993, opened 2023-08-26
chore: make maintenance tasks migrations safe

VaibhavUpreti:

Marking potentially dangerous migrations introduced by [maintenance_tasks](https://github.com/Shopify/maintenance_tasks) as safe in to order to prevent exception thrown by [strong_migrations](https://github.com/ankane/strong_migrations) gem.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,8 +42,6 @@ jobs:
       - name: Deploy with kamal
         run: |
           kamal lock release
-          kamal accessory boot yosys2digitaljs-server
-          kamal traefik boot
           kamal redeploy -vvvv --skip_push
         env:
           RAILS_SERVE_STATIC_FILES: true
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -42,8 +42,6 @@ jobs:
       - name: Deploy with kamal
         run: |
           kamal lock release
-          kamal accessory boot yosys2digitaljs-server
-          kamal traefik boot
```
tachyons, line 46: Unrelated

VaibhavUpreti: we can remove this as we have already booted those containers.

VaibhavUpreti: Here you can see, this is my personal server:
If we run this again then that container already exists. Hence the action will fail.

test-servre

tachyons: Sorry, merged the PR without checking this thread. Our deployment script should be smart enough to check if it is already booted or not.

PR #3997, opened 2023-08-27
Revert "Docker based development improvement"

tachyons:
Reverts CircuitVerse/CircuitVerse#3913


tachyons:
@tanmoysrt we had to revert this as the CI failed with new setup

tanmoysrt:
> @tanmoysrt we had to revert this as the CI failed with new setup

Yes got it, will update the workflow with the new setup

PR #4000, opened 2023-08-27
chore: mark maintenance tasks migrations safe

VaibhavUpreti:

Mark migrations introduced by maintenance_tasks gem as safe.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,8 +42,8 @@ jobs:
       - name: Deploy with kamal
         run: |
           kamal lock release
-          kamal accessory boot yosys2digitaljs-server || true
-          kamal traefik boot || true
+          kamal accessory boot yosys2digitaljs-server
+          kamal traefik boot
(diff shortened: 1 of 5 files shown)


VaibhavUpreti:
We will get a true exit code if the container already exists, otherwise kamal will start it.

.github/workflows/deploy.yml
```diff
@@ -42,8 +42,8 @@ jobs:
       - name: Deploy with kamal
         run: |
           kamal lock release
-          kamal accessory boot yosys2digitaljs-server
-          kamal traefik boot
+          kamal accessory boot yosys2digitaljs-server || true
```
tachyons, line 45: Move this to separate PR

tachyons (changes requested):

.github/workflows/deploy.yml
VaibhavUpreti: Done @tachyons, first need to merge this then the other one https://github.com/CircuitVerse/CircuitVerse/pull/4001

PR #4011, opened 2023-09-03
fix: use proxy url for assets for cdn caching

VaibhavUpreti:


VaibhavUpreti:
@tachyons we can merge this, in the last patch I missed this change, Instead of directly loading the url of the object from the s3 bucket(expires in 5mins), we use a proxy url by rails that is fixed for the attachment, so that the cdn can cache it.

Something like
 ```url
https://circuitverse.org/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejZWQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ae99fcfc3164c1ded570463f305041bf8a7ffe1e/avatar.jpeg
```

PR #4012, opened 2023-09-03
fix: specific check for use foreman in github codespaces

tanmoysrt:
#### Fix
In github codespaces, it is running the configuration for docker environment but it should start the processes using foreman.

- In `/bin/dev` checks added for `REMOTE_DEVELOPMENT_PLATFORM` env variable. If it's set it will not run the command for docker one . Instead it will start all processes using foreman


.devcontainer/docker-compose.yml
```diff
@@ -17,7 +17,8 @@ services:
     environment:
       REDIS_URL: "redis://redis:6379/0"
       CIRCUITVERSE_USE_SOLR: "false"
-      DOCKER_ENVIRONMENT: 'true'
+      DOCKER_ENVIRONMENT: "true"
+      REMOTE_DEVELOPMENT_PLATFORM: "true"
```
tachyons, line 21: dev containers are used for local development as well right ?

.devcontainer/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ruby:3.2.1
 # install dependencies
 RUN apt-get update -qq && apt-get install -y imagemagick shared-mime-info libvips && apt-get clean
 
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash \
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash \
```
tachyons, line 6: Could you use move node update to separate PR ?

tanmoysrt: Ok :+1:

.devcontainer/docker-compose.yml
tanmoysrt: For locally, the recommended method was to use the docker environment by `bin/docker_run`. 

But anyone can use the provided configuration to start the dev container in vscode. We haven't tested or documented this. This configuration was specifically added for Github Codespaces

.devcontainer/Dockerfile
tanmoysrt: updated !

.devcontainer/docker-compose.yml
tachyons: @tanmoysrt then Can we us the env DEV_CONTAINER to make this more specific ?

tanmoysrt: @tachyons as there is no pre-existing environment variable like DEV_CONTAINER in the codespaces environment. Added the DEV_CONTAINER variable in docker-compose and updated

PR #4039, opened 2023-09-11
chore: update kamal config

VaibhavUpreti:

- Custom workflow to run `kamal` commands
- use `host.docker.internal` 
- write VAPID keys before running container


Me-Priyank:
kamal config exists and has since been updated directly.

PR #4096, opened 2023-09-30
Revert "Change the implementation of the priority queue to use a heap data structure"

Prerna-0202:
Reverts CircuitVerse/CircuitVerse#3916 

The older circuits made with delays breaks with the new queue implementation since the delays create a different resolution order.


Prerna-0202:
@tachyons

PR #4155, opened 2023-10-08
fix(emailbox): fix extra size of emailbox

### LINKED ISSUE #4146: 🐞 Bug:email box is not responsive

### Describe the bug

the email container is not responsive at all
Screenshot (321)

### Steps To Reproduce

.

### Expected Behavior

.

### Screenshots

_No response_

```markdown
- OS:
- Browser:
- version:
```

### Additional context

_No response_

Yes

harshsharma-11:
Fixes #4146 

Screenshot (330)

### DIFF
--- a/app/assets/stylesheets/select2_overrides.scss
+++ b/app/assets/stylesheets/select2_overrides.scss
@@ -1,5 +1,5 @@
 .select2-container span.select2-selection--multiple {
-   max-width: 450px;
+  max-width: 450px;
   min-height: 150px;
   width: 70vw;
 }
--- a/app/assets/stylesheets/select2_overrides.scss
+++ b/app/assets/stylesheets/select2_overrides.scss
@@ -1,6 +1,20 @@
 .select2-container span.select2-selection--multiple {
   min-height: 150px;
-  min-width: 450px;
+  width: 350px;
+}
+@media screen and (max-width: 402px) {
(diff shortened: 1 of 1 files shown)


harshsharma-11:
@VaibhavUpreti @vedant-jain03 @tachyons please review the pr

app/assets/stylesheets/select2_overrides.scss
```diff
@@ -1,6 +1,23 @@
 .select2-container span.select2-selection--multiple {
   min-height: 150px;
-  min-width: 450px;
+  width: 350px;
```
tanmoysrt, line 3: Hardcoded `width` with `px` unit will not work as expected !
With this approach we need to have breakpoint for each type of screen which is obviously not feasible.

It will be better if you can anyhow related the width with the parent of the element

tanmoysrt (changes requested):
With the current changes

image
You can see , the label of input is coming in same row. So we should move the parent of those to `flex` and use `flex-col`

app/assets/stylesheets/select2_overrides.scss
harshsharma-11: ok let me try something ..

harshsharma-11:
@vedant-jain03 @VaibhavUpreti please review

PR #4257, opened 2023-11-06
feat: added toggle for user registration in flipper

### LINKED ISSUE #4236: Feature: disable signup

I am trying to deploy CircuitVerse within my university for a course. The local IT people would prefer that we limit account creation to comply with our security policy.

It would be nice if there was some way for the administrator of a CircuitVerse instance to disable signup, i.e., to prevent new users from registering an account. This could happen either via setting or a switch on the admin panel.

I considered integrating with my university's SAML provider, and still intend to do this in the long run. Going by the interaction I had on Slack, however, that still requires a different change on your end. For now, disabling registration is enough for me.

### Additional context

_No response_

### Are you working on this?

No

niladrix719:
Fixes #4236 

1. Added block_registration toggle in flipper configuration
2. Checks if flipper :block_registration is enabled / disabled for Signup, Auth provider signup and Api signup
3. Wrote test cases for registration featured flag
4. Removed the signup link and accessing the signup page will now redirect the user to login page when :block_registration enabled

https://github.com/CircuitVerse/CircuitVerse/assets/91966855/dfc9c98e-3129-431a-924a-1e79bf329da9


niladrix719:
cc @tachyons

tachyons:
Great job @niladrix719 . Could you make feature flag "block_registration" and block signup only if this toggle is on ?

niladrix719:
@tachyons i made signup flag by default true which means it allows the signup but it can be toggled false in flipper, am i missing something?

niladrix719:
i have changed it to :block_ registration @tachyons

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -29,24 +29,44 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+      return redirect_to(new_user_session_path, alert: "Registration is currently blocked")
+    end
+
     generic_callback("facebook")
   end
 
   def google_oauth2
+    if Flipper.enabled?(:block_registration)
+      return redirect_to(new_user_session_path, alert: "Registration is currently blocked")
+    end
```
tachyons, line 42: Will this prevent social sign-in as well? Because the endpoint is the same, could you please validate the flow?

spec/requests/api/v1/authentication_controller/oauth_signup_spec.rb
```diff
@@ -6,6 +6,7 @@
   describe "oauth user signup" do
     context "when user already exists" do
       before do
+        allow(Flipper).to receive(:enabled?).with(:block_registration).and_return(false)
```
tachyons, line 9: We don't need to stub this as this is the default behaviour, right ?

spec/system/signup_spec.rb
```diff
@@ -50,12 +55,43 @@
...
+    end
+  end
+
+  context "when registration is disabled" do
+    before do
+      allow(Flipper).to receive(:enabled?).with(:block_registration).and_return(true)
+      allow(Flipper).to receive(:enabled?).with(:recaptcha).and_return(false)
+      allow(Flipper).to receive(:enabled?).with(:gitlab_integration).and_return(false)
+      allow(Flipper).to receive(:enabled?).with(:sso_integration).and_return(false)
+      allow(Flipper).to receive(:enabled?).with(:forum).and_return(false)
```
tachyons, line 82: We can disable this by default in spec_helper so that we don't need to duplicate these every specs.

app/controllers/users/omniauth_callbacks_controller.rb
niladrix719: i checked it, and yes It is preventing the signin functionality also , i am looking for a solution for this

TobiasKappe:
Hi @niladrix719, thanks for picking this up so quickly!

This looks great. To answer @tachyons's question: I think it'd be good if we could also hide the registration form (as well as the link on the login page) from the UI.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -55,6 +55,8 @@ def generic_callback(provider)
     if @user.persisted?
       sign_in_and_redirect @user, event: :authentication
       set_flash_message(:notice, :success, kind: provider.capitalize) if is_navigational_format?
+    elsif Flipper.enabled?(:block_registration)
+      redirect_to new_user_session_path, alert: "Registration is currently blocked"
```
tachyons, line 59: We need to make these texts i18n compatible. We can do that in the follow-up issue.

tachyons (approved):

tachyons:
Great job @niladrix719 👍

app/controllers/users/omniauth_callbacks_controller.rb
niladrix719: sure i am opening one

PR #4336, opened 2023-12-02
fix: disabled CSRF check for github codespaces

### LINKED ISSUE #4334: Invalid AuthenticityToken Error in Github Codespaces

### Describe the bug

After launching the project in CodeSpaces , While registering or logging in with a user , it's providing the below mentioned error.
image

### Steps To Reproduce

1. Open the project in CodeSpace [[Follow Guide](https://github.com/CircuitVerse/CircuitVerse/blob/master/installation_docs/remote_development.md#github-codespaces)]
2. Register a user by visiting this route `http://localhost:3000/users/sign_up`
3. You will see the error

### Expected Behavior

It should not provide any error.

### Additional context

This error only comes when Github Codespaces run in Browser, and If you can open Codespaces in local VS Code, you will not face this issue.

No

niladrix719:
Fixes #4334 

disabled CSRF check for github codespaces

https://github.com/CircuitVerse/CircuitVerse/assets/91966855/a48680ed-61d6-451d-94eb-1edf7ac88cea


tachyons:
We need to add codespaces URL here https://github.com/CircuitVerse/CircuitVerse/blob/ee6fdf29621559a93c735bd69962eeaaa66239e3/config/environments/development.rb#L84 . Removing CSRF check all together makes it harder to test on code spaces.

niladrix719:
but in dev environment we are already whitelisting everything - https://github.com/CircuitVerse/CircuitVerse/blob/ee6fdf29621559a93c735bd69962eeaaa66239e3/config/environments/development.rb#L85

i have tried using `config.hosts << /.*\.app\.github\.dev\Z/ `which seems not to work

upon research i found out this issue 
https://github.com/github/codespaces-rails/issues/37

will investigate more if there is a solution without disabling it

tanmoysrt:
If you want to just check whether it's a codespace, you can use the `DEV_CONTAINER` environment variable
Ref - https://github.com/CircuitVerse/CircuitVerse/blob/master/.devcontainer/docker-compose.yml#L21

tanmoysrt:
> Hi @niladrix719 I checked the issue. It seems that Github Codespaces has a reverse proxy setup which pass the requests to the containers.
> 
> Now, while doing the reverse proxy they don't rewrite the Host Header's. As a result - Request URL becomes localhost:3000/* But Http Host Header also changes to localhost, but the X-Forwarded-Host header contains the actual URL (*.github.dev).
> 
> So, as a result it fails CSRF verification (request base !== X-Forwarded-Host).
> 
> In older Rails, It gives preference to X-Forwarded-Host header instead of Host Header for decide actual Host In latest patches, they have given preference to Host header to resolve cache poisoning attack. Ref - [rails/rails#29893](https://github.com/rails/rails/issues/29893) [rails/rails#22965](https://github.com/rails/rails/issues/22965)
> 
> Usually, while using nginx/haproxy/traefik we can rewrite this header but in this Codespaces we can't have that flexibility.
> 
> So, we may disable CSRF for only Github Codespaces
> 
> We can add this configuration to disable the CSRF check in development
> 
> ```
> config.action_controller.forgery_protection_origin_check = false
> ```

cc @tachyons

niladrix719:
@tanmoysrt thanks that was great help, just one question can we use CODESPACES environment variable instead of  DEV_CONTAINER to identify codespaces

tanmoysrt:
> @tanmoysrt thanks that was great help, just one question can we use CODESPACES environment variable instead of DEV_CONTAINER to identify codespaces

Yes you can
If you got time, just verify whether Dev container actually working In local environment.
In VS Code, you can run the same dev container. In codespaces that's actually running on the cloud.

tachyons:
Fantastic 🥳

niladrix719:
https://github.com/CircuitVerse/CircuitVerse/assets/91966855/bc7a07a9-53cf-4adb-84fe-70fc5f655f30

i think its working fine

tachyons (approved):

PR #4345, opened 2023-12-06
fix: moved stylelint-config-standard-scss to dev dependency

niladrix719:
Fixes Codeclimate check

Moved stylelint-config-standard-scss to dev dependency


tanmoysrt:
@niladrix719 It seems that Codeclimate may have some issue, so we have removed that package for now

PR #4374, opened 2023-12-14
fix: docs deployment cd

tanmoysrt:
- It seems that [andstor/jsdoc-action](https://github.com/andstor/jsdoc-action/tree/master) is not maintained for long time and using `node12` as per it's [source code](https://github.com/andstor/jsdoc-action/blob/master/action.yml)
- So, removed `andstor/jsdoc-action` from workflow, and generating docs using js only


tanmoysrt:
cc @tachyons @vedant-jain03

vedant-jain03 (approved):

PR #4386, opened 2023-12-16
fix: Text-position#4369

### LINKED ISSUE #4369: The position of ancho-text "Learn logic design"  is not responsive

### Describe the bug

In the home page the anchor text "Learn logic design" is not is the in the position when we change the aspect ratio . the design text is changing  its position 

### Steps To Reproduce

Go to https://circuitverse.org/
Change the aspect ratio 
And see the error

### Expected Behavior

image

### Screenshots

image

```markdown
- OS:
- Browser:
- version:
```

### Additional context

_No response_

Yes

upsaurav12:
Fixes #4369 

I used the media query in breakpoint 992px so that the text design does not change its position when width is reduced

Problem:-
image

Solution Should be :-
https://github.com/CircuitVerse/CircuitVerse/assets/116784047/679239a4-63b6-4790-b1b4-a70652d87787

### DIFF
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -49,7 +49,7 @@
 }
 
 //breakpoints
-@media (max-width: 990px){
+@media (max-width: 991px){
   .home-main-content{
     width: 345px;
   }
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -3,7 +3,6 @@
   margin-bottom: 60px;
   margin-top: 45px;
   text-align: left;
-  border: 2px solid black;
 }
 
 .home-image-cover {
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -3,6 +3,7 @@
   margin-bottom: 60px;
   margin-top: 45px;
   text-align: left;
+  border: 2px solid black;
 }
 
 .home-image-cover {
@@ -49,7 +50,7 @@
 }
 
 //breakpoints
-@media (max-width: 991px){
+@media (max-width: 992px){
   .home-main-content{
     width: 345px;
   }
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -49,6 +49,11 @@
 }
 
 //breakpoints
+@media (max-width: 990px){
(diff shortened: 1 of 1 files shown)


upsaurav12:
cc @tanmoysrt

app/assets/stylesheets/home.scss
```diff
@@ -49,6 +49,11 @@
 }
 
 //breakpoints
+@media (max-width: 992px){
+  .home-main-content{
+    width: 345px;
```
tanmoysrt, line 54: It's not a good idea to set a fixed width on a specific breakpoint. That will break responsiveness.

The issue was that `Learn Logic Design` was broken in two lines,,
There is 2 solution -
1. Instead of space between the words, use `&nbsp;` tag. If you use this tag this will provide spacing, also the word will never break , so all the three words will be in same line. Although, you need to check whether it works with i18n.
2. Currently those button is in HTML as `<a></a>` tag. You can move them to `button` tag which will also resolve this issue.

I think Solution 2 will be better in this case.

tanmoysrt (changes requested):

app/assets/stylesheets/home.scss
upsaurav12: ok i will try both solution

upsaurav12: i tried with 1st approach but it does not look good then i applied 2nd approach 
https://github.com/CircuitVerse/CircuitVerse/assets/116784047/dfd7691f-d3c7-446c-97c7-c93b4c40a599
if any suggestion about this solution ??

upsaurav12: @tanmoysrt  instead of making the anchor-text to button we can change the padding-right and padding-left of home-main-content container the result is like this 
**before:-**
image
**after:-**
image

upsaurav12: instead of making anchor-text to button we can change padding left and right the result is this 

https://github.com/CircuitVerse/CircuitVerse/assets/116784047/70dace23-5926-4d4d-b427-44d31181240f

tanmoysrt: Push the code, so that I can also check

tanmoysrt:
Closing this as you have raised new PR #4472

PR #4449, opened 2023-12-28
fix(comments): makes the urls clickable

fillingtothemomo:
## Fixes 
Earlier the URLs in the comments were not clickable and user had to copy paste them to go to that link, now the url is clickable and redirects you to the link.

In the `app/helpers/projects_helper.rb` I declared a helper function which checks for a url using simple regex  and in the `app/views/commontator/comments/_body.html.erb` I rendered the comment using the helper.

https://github.com/CircuitVerse/CircuitVerse/assets/122573982/78cf4b46-7d8c-4152-950d-673be4a58d0b


fillingtothemomo:
@tanmoysrt I used the sanitize method to clean the html content and the render the link as html safe still codeclimate detects a bug , I can use the raw method but that will also give the same error I guess, can you suggest any other alternate approaches?

app/views/commontator/comments/_body.html.erb
```diff
@@ -3,4 +3,6 @@
   comment
 %>
 
-<%= commontator_simple_format comment.body %>
+<div id="comment-body-<%= comment.id %>">
+  <%= auto_link_comments(comment) %>
```
VaibhavUpreti, line 7:
```suggestion
  <%= auto_link(comment) %>
```
Why not use the inbuilt method from the `rails_autolink` gem? Seems better than writing a custom helper function `auto_link_comments`.

https://github.com/tenderlove/rails_autolink#basic-usage

fillingtothemomo: Thanks a lot for correcting my approach, I looked into the method and it is much better than helper method. I have made the changes in my new commit.

fillingtothemomo:
@tanmoysrt can you kindly review the PR and suggest changes, if any.

tanmoysrt:
> @tanmoysrt can you kindly review the PR and suggest changes, if any.

Fix the issues of failed CI, check the CI logs

PR #4544, opened 2024-01-08
fix: Improved responsiveness by setting some minwidth to application

### LINKED ISSUE #4496: 🐞 Bug:Not Responsive Properly after some minwidth (UI/UX  enhancement )

### Describe the bug

Like bigger organisation which overflows in right x direction after some minwidth for responsiveness is not found here which will definitely increase user exxperience
Screenshot 2024-01-03 030932

### Steps To Reproduce

1. Go to any pages
2. Decrease it width using developer tools

### Expected Behavior

It should  overflows in right x direction after some minwidth for responsiveness
Screenshot 2024-01-03 031257
Screenshot 2024-01-03 031308

### Screenshots

_No response_

```markdown
- OS:
- Browser:
- version:
```

### Additional context

_No response_

Yes

priyanshagra:
Fixes #4496

Set the minwidth to application for improving responsiveness in some minwidth hence increase user experience

https://github.com/CircuitVerse/CircuitVerse/assets/127022024/940b77fc-414d-4e60-9811-43cdcdabc9bb

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -27,3 +27,7 @@
 @import 'announcements';
 @import 'select2.css';
 @import 'select2_overrides';
+
+body {
+    min-width: 320px;
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,11 +20,6 @@
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <%= render partial: "layouts/google_analytics" if Rails.env.production? %>
     <%= stylesheet_link_tag "application" %>
-    <style>
-        body {
-            min-width:320px;
-        }
-    </style>
   </head>
 
   <body>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,6 +20,11 @@
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <%= render partial: "layouts/google_analytics" if Rails.env.production? %>
     <%= stylesheet_link_tag "application" %>
+    <style>
+        body {
+            min-width:320px;
+        }
+    </style>
   </head>
 
   <body>


priyanshagra:
please review it @tanmoysrt @nitin10s

tachyons:
Hey @priyanshagra , Welcome to Circuitverse. In order to keep our code maintainable, we separate styles to css files,  could you please update the PR with CSS instead of using style tag in HTML? 

Also don't make changes to cv-frontend-vue dir to the PR.

priyanshagra:
@tachyons  please check changes done

app/assets/stylesheets/application.scss
```diff
@@ -26,4 +26,8 @@
 @import 'commontator';
 @import 'announcements';
 @import 'select2.css';
-@import 'select2_overrides';
\ No newline at end of file
+@import 'select2_overrides';
+
+body {
+    min-width: 320px;
+}
```
tanmoysrt, line 33: It's not a good idea to set min-width to body. Instead we can try to make other stuffs responsive.
`body` is the top layout, if your device has less than 320px width, the layout will break and show scrollbars everywhere.

tanmoysrt (changes requested):

app/assets/stylesheets/application.scss
priyanshagra: sir in the body further responsiveness is not possible according to images and text 
and also google and other big org. follows this type of styling you can check youtube by doing so it is better to do this as this is a standard practise

tanmoysrt: If it's not possible, we can skip that. We don't need to squeeze page below 280 px(galaxy fold), we can keep other stuffs as responsive keeping 280px as our target.

Share some official docs (maybe of mdn if available), which can confirm this as a standard practice.

priyanshagra: so then i apply apply minwidth after 280px till that i will make reponsive

priyanshagra: is that okay @tanmoysrt

PR #4634, opened 2024-01-22
style: removes extra space above footer

Pulkit0729:
Fixes #3621

I have decreased the space above footer to 48px, which is a multiple of 4 and closest to 50, which was discussed in previous comments.
Screen Shot 2024-01-22 at 10 27 56 PM

### DIFF
--- a/app/assets/stylesheets/footer.scss
+++ b/app/assets/stylesheets/footer.scss
@@ -1,5 +1,5 @@
 .footer-empty-div {
-  height: 250px;
+  height: 48px;
 }
 
 .footer-container-fluid {


tanmoysrt (changes requested):
We can't reduce the height directly. The issue you can see on the picture. If content is less footer will come also upper.

image

Either
1. We may need to some special class like `-webkit-fill-available` to auto fill required height.
2. We need to add additional spacing in required pages.

VaibhavUpreti:
Fixed via https://github.com/CircuitVerse/CircuitVerse/pull/5278

PR #4790, opened 2024-03-02
Revert "build(deps): bump rails from 7.0.8 to 7.0.8.1"

tanmoysrt:
Reverts CircuitVerse/CircuitVerse#4776


tanmoysrt:
@tachyons can we revert this for now, as it has broken the circuitverse installation/setup, contributors are not able run the project in Gitpod/Codespaces.
As there are warnings from Security Audit, We can check and upgrade the versions of gem manually.

vedant-jain03 (approved):

tachyons:
It should be fixed by now

PR #4882, opened 2024-04-09
Deferred Contentions

gr455:
Before this commit, contentions were checked at output nodes and if a disagreeing value was found (only exception being Tristates), the simulation would error out with a contention error. This is not correct since some contentions might occur temporarily and might resolved in the same play() cycle down the resolution queue.

This commit maintains an [effectively] set of all contentions that are found during resolution. Some contentions might get resolved in the simulation and they are removed from the set. At the end of the play cycle, if any contentions remain, then only those are real contentions and are reported.

Before: Contention Error
Screenshot from 2024-04-09 19-51-39

After:
Screenshot from 2024-04-09 19-52-12

### DIFF
--- a/simulator/src/contention.js
+++ b/simulator/src/contention.js
@@ -14,8 +14,8 @@ export default class ContentionPendingData {
 	// Adds 
 	add(ourNode, theirNode) {
 		if (this.contentionPendingMap.has(ourNode)) {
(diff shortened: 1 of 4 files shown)


gr455:
Should be back compatible, so changing base branch to master.

PR #4917, opened 2024-04-30
fix: notifications when the projects or users are deleted

### LINKED ISSUE #4904: Fix notifications when the projects or users are deleted
When a project is forked, a notification is generated. But if that project is deleted, error is displayed since the notification unable to find the notification.

Solution: 

Use null object pattern

Ref:

- https://dev.to/daviducolo/exploring-the-null-object-pattern-in-ruby-k5l
- https://thoughtbot.com/blog/rails-refactoring-example-introduce-null-object

```diff
  def message
-    user = params[:user]
-    project = params[:project]
+    user = params[:user] || DeletedUser.new
+    project = params[:project] || DeletedUser.new
    t("users.notifications.fork_notification", user: user.name, project: project.name)
  end
```

Malavi1:
Fixes #4904

* Introduced a new **`DeletedUser`** class it is called whenever the user or project is nil/false.


app/models/deleted_user.rb
```diff
@@ -0,0 +1,8 @@
+# app/models/deleted_user.rb
```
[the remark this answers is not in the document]
Malavi1: added

app/models/deleted_user.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class DeletedUser
+  def name
+    "Unknown User"
```
tachyons, line 5:
```suggestion
    "Deleted User"
```

app/notifications/fork_notification.rb
```diff
@@ -4,9 +4,9 @@ class ForkNotification < Noticed::Base
   deliver_by :database, association: :noticed_notifications
 
   def message
-    user = params[:user]
-    project = params[:project]
-    t("users.notifications.fork_notification", user: user&.name, project: project&.name)
+    user = params[:user] || DeletedUser.new
+    project = params[:project] || DeletedUser.new
+    t("users.notifications.fork_notification", user: user.name, project: project.name)
```
tachyons, line 9: Could you apply the same logic for other notifications too ?

Malavi1: Sure I do

app/models/deleted_user.rb
Malavi1: Done with change

naman79820:
stale and duplicate pr

Issue #4928, opened 2024-05-13
🐞 Bug : Corner Triangular Edge Display Issue Of Logic Gates


Darshan808:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When viewing circuit elements such as *OR Gate*, *NOR Gate*, and *X-OR* Gate in the simulator , their **upper edge** appears slightly messed up on default zoom. This issue affects the UI as it causes the corners of these gates to flicker and change their appearance when the position of gates is changed.

### Steps To Reproduce

1. Open https://circuitverse.org/simulator
2. Add *OR Gate* to the simulation Area.
3. Keep on changing the position of OR gate.
4. Observe the *upper edge* of the gate at default zoom level.

### Expected Behavior

The corner triangular edges of the gates should appear smooth and consistent at all zoom levels, without any distortion or flickering at every position.

### Screenshots

![Screenshot 2024-05-13 174822](https://github.com/CircuitVerse/CircuitVerse/assets/127468609/0d79a7c5-e512-4933-ac2b-c48a9d80b07d)
![Screenshot 2024-05-13 174850](https://github.com/CircuitVerse/CircuitVerse/assets/127468609/ef40a2f3-d7c7-4b42-8a0c-2534a7ce41bd)

### Device Information [optional]

```markdown
- OS:
- Browser:
- version:
```

### Are you working on this issue?

Yes

Jay2006sawant:
Would be taking up this issue

vermagaurav24:
i think it is looking good on moving OR gate or other gates.

PR #4929, opened 2024-05-13
fix(UI): gates upper edge drawing issue fix

Darshan808:
#### Issue
This pull request addresses issue #4928 :  Corner Triangular Edge Display Issue Of Logic Gates.

#### Problem
The issue arises due to the upper edges of circuit elements, particularly gates with triangular corners like OR Gate, NOR Gate, X-OR Gate, etc., being slightly distorted on the default zoom level. This distortion affects the UI negatively, causing an inconsistent appearance.

#### Root Cause
Upon investigating, it was found that the distortion occurs during the drawing of the gates. While drawing curves, the `Math.round()` function is correctly used to round off decimal numbers. However, when moving the canvas context (ctx), `Math.round()` is not applied to the moving coordinates. As a result, when the `ctx.closePath()` method is called, the presence of decimal numbers, particularly .5 in both x and y coordinates, causes the upper edge distortion.

#### Solution
To resolve this issue, `Math.round()` is applied to the values before moving the ctx. This ensures that the coordinates are always rounded to whole numbers, resulting in smooth curve creation for the gates drawing.

#### Impact
Implementing this solution ensures consistent and distortion-free rendering of the circuit elements' upper edges, enhancing the UI experience for users.

### Changes Made

- Applied `Math.round()` to coordinates before moving the canvas context to ensure whole number values.
- Tested the changes thoroughly to ensure they resolve the issue without introducing any other problems.
  
### Testing

Manual testing was done to validate the changes.

### Screenshots
####Before
Screenshot 2024-05-13 174850
Screenshot 2024-05-13 174822
Screenshot 2024-05-13 203051

####After (in every condition)
Screenshot 2024-05-13 174906
Screenshot 2024-05-13 221921

## Summary by CodeRabbit

- **New Features**
	- Improved rendering accuracy on the canvas by modifying the coordinate handling for the `moveTo` function when the `bypass` parameter is true.
  
- **Bug Fixes**
	- Enhanced precision in drawing, particularly in scenarios requiring exact positioning.


Darshan808:
@Prerna-0202 Please review my PR. Or please tell me if i am missing something.

Prerna-0202 (approved):
Great 🚀 @Darshan808 . Thanks for this fix.

PR #4939, opened 2024-05-15
Update docker.md for setup on windows using WSL

Darshan808:
Added documentation for setting up the project using WSL on Windows

### DIFF
--- a/installation_docs/docker.md
+++ b/installation_docs/docker.md
@@ -2,7 +2,34 @@
 
 ---
 
-### Windows
+### Windows Using WSL (Recommended)
+**Prerequisites**
+|  Name | Version | Installation |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


subhamkumarr:
The instructions for setting up the project on `Windows using WSL` seems unnecessary. Since we can already set up the project using Docker on Windows, so why do we need to use WSL? We should not add this section to the documentation.
To keep the setup instructions simpler.

Darshan808:
I understand. I faced the issue of slow running docker on windows which was
solved by setting up the project through WSL which is why i thought the
docs should be updated. But that might be my device issue and the
instruction may cause complications in setup.
Thank you for your time.
I'll close my pull request.

subhamkumarr:
@Darshan808 Yes

PR #5367, opened 2025-02-01
Upgrade to Node.js 22

gitsofaryan:
Fixes #5304 

#### Describe the changes you have made in this PR -
- Updated all configurations, Dockerfiles, and documentation to use Node.js 22.
- Tested the changes locally and with Github Actions to ensure compatibility and smooth functionality with the new Node.js version.
- their were dependency issues conflict between canvas and jest solved using ```--legacy-peer-deps``` across percy.yml, docs.yml and package.json.
- removed ```--frozen-lockfile --immutable``` in percy.yml

### Screenshots of the changes (If any) -
image

Note: Please check **Allow edits from maintainers.** if you would like us to assist in the PR. 

## Summary by CodeRabbit

- **Chores**
	- Upgraded the Node.js version to 22 across container builds, CI workflows, and testing environments.
	- Updated several development dependencies for improved stability.
- **Documentation**
	- Revised installation guides for Linux, macOS, and Windows to reflect the new Node.js version requirement.
	- Enhanced clarity in documentation regarding dependency installation links.

PR #5376, opened 2025-02-03
Fix: doc.yml node update

gitsofaryan:
Fixes #5367 

- pre-install build tool :
```
        run: |
          sudo apt-get update
          sudo apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
```

## Summary by CodeRabbit

- **Chores**
  - Enhanced the documentation build process by adding an automated step to install essential build dependencies, ensuring smoother documentation generation.

- **Refactor**
  - Improved the simulation validations by refining how input and output labels are compared, resulting in more consistent behavior.


simulator/src/testbench.js
```diff
@@ -536,7 +536,7 @@ function runSingleSequential(testbenchData, scope) {
  */
 function setInputValues(inputs, group, caseIndex, scope) {
     group.inputs.forEach((input) => {
-        inputs[input.label.trim()].state = parseInt(input.values[caseIndex], 2);
+        inputs[input.label].state = parseInt(input.values[caseIndex], 2);
```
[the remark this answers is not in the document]
gitsofaryan: LGTM

.github/workflows/docs.yml
```diff
@@ -17,6 +17,12 @@ jobs:
         with:
           node-version: '22'
 
+# This is required for jsdoc
+      - name: Install build tools 
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
```
tachyons, line 24: Why jsdoc requires cairo? Are we installing canvas package which is only needed for testing.

gitsofaryan: The cairo package is getting pulled in because of the canvas dependency, which is needed for building assets or running tests. Since we're upgrading to Node.js 22, it looks like these dependencies are important for making sure everything works smoothly with the new version. If we don't need canvas in production, I can make sure it only installs in the dev environment. Let me know if you want me to tweak that!

tachyons: Yes, it is dev dependency. No need to install it everywhere

gitsofaryan: Got it! Thanks for pointing that out. I'll make sure to update the setup so that the canvas dependency, and its associated packages like cairo, are only installed in the dev environment. I’ll get that adjusted!

PR #5379, opened 2025-02-03
Fix: doc.yml node 22 update

gitsofaryan:
Fixes #5367

- Removed the unnecessary Cairo dependencies from the build tools as they weren’t required for our current setup.
- Kept only the Canvas-related build tools, which are essential for building and testing the project.
- This ensures a cleaner and more efficient workflow while upgrading to Node.js 22.

N/A

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,6 +17,11 @@ jobs:
         with:
           node-version: '22'
 
+      - name: Install build tools 
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
+
       - name: jsdocs-dep
         run: npm install --legacy-peer-deps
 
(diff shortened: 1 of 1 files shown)


gitsofaryan:
@tachyons please review!!

.github/workflows/docs.yml
```diff
@@ -17,6 +17,11 @@ jobs:
         with:
           node-version: '22'
 
+      - name: Install build tools 
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
```
tachyons, line 23: Hey you seem to have removed Cairo, but you still have other build dependencies. Why those weren't needed before ?

gitsofaryan: I removed Cairo because it’s no longer needed for docs. However, the other dependencies (libpango, libjpeg, libgif, librsvg2) are still required for Canvas to work properly. Previously, they were installed as indirect dependencies, but now I'm explicitly adding them to avoid potential issues across different environments.

PR #5480, opened 2025-03-02
test(percy): testing percy functionality

salmoneatenbybear:
Testing the Percy functionality. 

Made a change in the Login page in the UI to test whether percy is catching it correctly. 

Before the change i made the ui looked like this ( notice the login text header ) : 

image

After : 

image

## Summary by CodeRabbit

- **New Features**
  - Updated the login screen's main heading to display "Sign In" instead of the previous generic label. This change offers clearer guidance for users, making it easier to identify the authentication process. Users will now experience a more intuitive entry point when signing in.


salmoneatenbybear:
image


Currently percy is not able to detect the change as the integration is not yet fully complete, as you can see from ci/cd log for the test above, percy is not active.

salmoneatenbybear:
cc : @vedant-jain03 @tachyons

salmoneatenbybear:
image


Percy is working but on PRs merged in the master branch because the token isn't accessible to the PRs, let me try and figure out a way around this... as the whole point of percy is to detect changes before merge in master 

CC : @tachyons @vedant-jain03 @Asrani-Aman

salmoneatenbybear:
right now Percy only runs after PRs are merged. This is the standard “secrets aren’t exposed to forked pull-requests” issue on GitHub Actions.

these are according to me the possible fixes for this with the first one being the most appropriate according to me...

- Switch to pull_request_target (with a careful workflow so secrets aren’t leaked).
- Require branches be created in this repo (instead of forks).
- Require manual approval for forked PRs so Percy can access secrets.

Which approach works best for us? @tachyons @vedant-jain03 @Asrani-Aman

salmoneatenbybear:
IMP security notice if we go with option 1 : we need to ensure the  tests are locked down so they don’t accidentally leak PERCY_TOKEN.

salmoneatenbybear:
should i contact the percy support engineer i had a zoom call with for initial setup related questions ?

tachyons:
Try tagging them here, they might be able to help

salmoneatenbybear:
@parth-bstack

Can you please help with above ? 

would be extremely grateful...

salmoneatenbybear:
Next step would be to decide critical UI flows we anticipate changes in this summer... and add percy snapshot commands there to test them when running the percy tests.


CC : @vedant-jain03 @tachyons @Asrani-Aman

PR #5552, opened 2025-03-11
Restrict Percy Tests to Exact Matching "check-visual-regression" Label

salmoneatenbybear:
## Summary
This pull request updates the condition for running Percy visual tests so that it matches the PR label `"check-visual-regression"` **more precisely**. 

Previously, the workflow just searched for the substring `check-visual-regression` anywhere in the labels’ JSON. Now, it looks explicitly for the snippet `"name":"check-visual-regression"`, ensuring we only trigger Percy tests if a PR has that **exact** label name.

## Details
- **Why this helps**: Prevents any partial or accidental match and guarantees only the `"check-visual-regression"` label triggers the test.
- **Outcome**: Percy tests won’t appear for a PR unless it is explicitly labeled `"check-visual-regression"`, or if it’s a direct push to `master`/`main`.

---------------------

This is to iterate over #5547 to solve the issue encountered in #5476 

## Summary by CodeRabbit

- **Tests**
	- Enhanced the triggering logic for visual regression checks to ensure more consistent quality verification.

- **Style**
	- Introduced minor formatting refinements in configuration for improved clarity.

### DIFF
--- a/.github/workflows/percy.yml
+++ b/.github/workflows/percy.yml
@@ -14,9 +14,12 @@ jobs:
   percy-tests:
     
     if: >
-      ${{ github.event_name == 'push' 
-         || (github.event_name == 'pull_request_target' 
-             && contains(toJson(github.event.pull_request.labels), 'check-visual-regression')) }}
+      github.event_name == 'push'
+      OR (
+        github.event_name == 'pull_request_target'
+        AND
+        contains(toJson(github.event.pull_request.labels), '"name":"check-visual-regression"')
+      )
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


salmoneatenbybear:
CC : @vedant-jain03 @tachyons

salmoneatenbybear:
Something related to an async op / callback not firing is causing the notification related rspec to fail alongwith which the Percy is bundled... Which causes the Percy check to fail here... Looking into it, we need to separate those two somehow because I think the time taken by the snapshot command is disturbing the notification selectors

vedant-jain03 (approved):

salmoneatenbybear:
@vedant-jain03 thanks for approving... if you could please take a look at #5553 which fixes the ci/cd which is causing the indirect trip over here... once that is merged i can update this pr and the ci/cd should pass smoothly.

PR #5580, opened 2025-03-16
fix: percy label not being detected

salmoneatenbybear:
## Description
This PR updates the Percy GitHub Actions workflow to correctly trigger visual regression tests only when a PR has the **check-visual-regression** label. Previously, the workflow was not running even when the label was present. 

### Key Changes:
- Added `labeled` to `pull_request_target.types` to ensure the workflow runs when the label is added.
- Improved the condition checking for PR labels in the `if:` clause.
- Added a debug step to print the PR labels at runtime to help diagnose label detection issues.

With this change, Percy visual regression tests should correctly run when the `check-visual-regression` label is present on a pull request.

## Summary by CodeRabbit

- **Chores**
  - Expanded workflow triggers to cover more pull request events.
  - Streamlined label checks for visual regression tests.

### DIFF
--- a/.github/workflows/percy.yml
+++ b/.github/workflows/percy.yml
@@ -6,7 +6,7 @@ on:
       - master
       - main
   pull_request_target:
-    types: [opened, reopened, synchronized, labeled]
+    types: [opened, reopened, synchronize, labeled]
     branches:
       - master
       - main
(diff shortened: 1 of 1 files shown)


salmoneatenbybear:
Before adding "check-visual-regression" Label Percy workflow *not triggered* as desired : 


https://github.com/user-attachments/assets/56193e70-f6f3-4d1c-b6ba-7ece90434e20



After adding "check-visual-regression" Label Percy workflow *triggered* as desired :

PR #5644, opened 2025-04-15
bug: improved search rendering function

mahmoodalisha:
Fixes #

 

The Current Rendering logic inside the search function appends elements
one by one using (searchResults.append()) Here $.each(results, function
(index, result) {...} iterates over results and appends each li separately
which can make UI sluggish. I have also noticed there is calling .empty() before appending results.
searchResults.empty(); clears the existing results. Then, $.each(results,
function (index, result) {...}) iterates over each result and appends it one by
one using searchResults.append(...) This causes multiple individual updates
to the DOM instead of a single update. Thus .append() is called multiple times, once for each result which is a bad performance. highlight.call(searchInput); is called after every single li is appended. This
means each time a search result is added, it recalculates the highlight effect.
So if we have 20 results, highlight() runs 20 times instead of just once at the end. 

Instead of appending elements one by one,I have built the entire HTML for all
results as a single string. Then, updated the DOM only once by setting
searchResults.html(resultHTML). so I have made sure to reduce DOM operations from O(n) to O(1).

https://github.com/user-attachments/assets/4e41591a-af5f-4d7c-bf44-21c6b9237df6

- [ ] If it is a core feature, I have added thorough tests.

## Summary by CodeRabbit

- **Refactor**
  - Improved the performance and clarity of the search feature, resulting in faster and more efficient search results display.
  - Streamlined the way "No Results Found" messages are shown for a cleaner user experience.

### DIFF
--- a/docs/source/javascripts/app/_search.js
+++ b/docs/source/javascripts/app/_search.js
@@ -56,36 +56,42 @@
   }
 
   function search(event) {
-
     var searchInput = $('#input-search')[0];
-
+    var searchTerm = searchInput.value;
+  
     unhighlight();
     searchResults.addClass('visible');
-
+  
     // ESC clears the field
     if (event.keyCode === 27) searchInput.value = '';
-
-    if (searchInput.value) {
-      var results = index.search(searchInput.value).filter(function(r) {
+  
+    if (searchTerm) {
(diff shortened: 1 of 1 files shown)


naman79820:
no issue tagged

PR #5815, opened 2025-06-15
Migrate Project Card to View Component

senbo1:
Fixes #5814

 

### Screenshots of the UI changes (If any) -

Screenshot 2025-06-15 at 2 57 14 PM

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.

Note: Please check **Allow edits from maintainers.** if you would like us to assist in the PR.

PR #5886, opened 2025-07-12
fix: remove obsolete 2-arg calls to project_image_preview

### LINKED ISSUE #5885: Bug: ArgumentError (given 2, expected 1) in Circuitverse#index
## Summary
After merging PR #5815 ("Migrate Project Card to View Component"), the helper method `project_image_preview` was modified to accept only one positional argument (`project`). Several legacy views (including `app/views/circuitverse/index.html.erb`) still call this helper with two arguments (`project, current_user`), causing a runtime argument mismatch.

## Where the Error Shows Up
- **File:** `app/views/circuitverse/index.html.erb`
- **Line:** ~216
- **Offending Call:** 
  ```erb
  <%= project_image_preview(project, :thumb) %>
  ```

Image

## Root Cause
Commit `c3284f21` (part of PR #5815) changed the method signature:
```diff
- def project_image_preview(project, current_user)
+ def project_image_preview(project)
```

salmoneatenbybear:
## Summary

Fixes #5885 

## Context 

| Commit / PR | What changed | Impact |
|-------------|--------------|--------|
| **PR #5815** – “Migrate Project Card to View Component” | Removed the `current_user` positional parameter from `project_image_preview(project)` | Any view still calling the *old* two-arg form now fails with an arity mismatch. |

---

## Changes in This PR

| File | Old Call | New Call |
|------|----------|----------|
| `app/views/circuitverse/index.html.erb` | `project_image_preview(circuit, current_user)`<br>`project_image_preview(project, current_user)` | `project_image_preview(circuit)`<br>`project_image_preview(project)` |
| `app/views/featured_circuits/index.html.erb` | `project_image_preview(project, current_user)` | `project_image_preview(project)` |

---

### DIFF
--- a/app/views/circuitverse/index.html.erb
+++ b/app/views/circuitverse/index.html.erb
@@ -181,7 +181,7 @@
             <% @featured_circuits.each do |circuit| %>
               <div class="col-12 col-sm-12 col-md-6 col-lg-4">
                 <div class="card circuit-card">
-                  <%= image_tag project_image_preview(circuit, current_user), alt: circuit.name, class: "card-img-top circuit-card-image", height: "180", width: "288" %>
+                  <%= image_tag project_image_preview(circuit), alt: circuit.name, class: "card-img-top circuit-card-image", height: "180", width: "288" %>
                   <div class="card-footer circuit-card-footer">
                     <div class="circuit-card-name">
                       <p><%= circuit.name %></p>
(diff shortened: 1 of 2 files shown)


tachyons:
We can remove this helper and inline it.

salmoneatenbybear:
@tachyons 

The CI failure is the same intermittent Selenium/Capybara flake we keep seeing across various system specs (today it surfaced in assignment_spec.rb, other runs have flagged different files). Chrome re-renders the page between the moment Selenium locates an element and the instant it checks it, so the node ID is suddenly “not in the document” and the test explodes. It’s purely a timing issue.

The solid fix I think is to refactor the affected steps to use Capybara’s synchronised helpers (click_link, accept_confirm, follow-up expect(page).to have_text / have_no_selector, or a change { Model.count } assertion) instead of raw element calls; that forces Capybara to wait for the DOM to settle and should eliminate these flakes project-wide.

**Do you think we should make this fix ?**

senbo1:
LGTM, Sorry for the mistake in #5815, will take care next time when updating existing methods/variables.
Please review @tachyons

salmoneatenbybear:
> LGTM, Sorry for the mistake in #5815, will take care next time when updating existing methods/variables. Please review @tachyons

No worries @senbo1 no need for apologies, thanks for being open to collaborating.

PR #5891, opened 2025-07-13
 setup playright

tachyons:
TODO


tachyons:
@abhisek regarding the playright package

tachyons:
cc @blazethunderstorm

salmoneatenbybear:
@tachyons after this PR, contest specs started to fail which I think is because of reliance on Selenium and headless chrome.

PR #5964, opened 2025-08-11
fix(users): make votes use ActsAsVotable::Vote (as: :voter) to prevent NameError on user destroy

salmoneatenbybear:
Fixes : #5961


salmoneatenbybear:
@tachyons please review this

PR #5995, opened 2025-08-23
fix: add regex to exclude unwanted tags + sort key

### LINKED ISSUE #5980: Enhance Logic for browse by tags for the tags which are shown to remove non useful tags.
https://github.com/CircuitVerse/CircuitVerse/blob/baa1b5dbdd3a7beb3aeed2fa6268e59107f2558c/app/controllers/explore_controller.rb#L94

Image

There are tags like 1,2 listed which are not really signifying anything.

salmoneatenbybear:
Fixes #5980

## Summary by CodeRabbit

- **Bug Fixes**
  - Top Tags now excludes tags composed only of non-letter characters (e.g., symbol-only tags), reducing noise and improving discoverability.
  - When multiple tags share the same usage count, they are sorted alphabetically (case-insensitive) for a stable, predictable order across Explore and tag listings.


salmoneatenbybear:
@tachyons I haven't bumped the cache key version so after deploy it can take up to 5 hrs for the changes to be reflected if there is a fresh cache at the time of deploy

PR #5966, opened 2025-08-11
il8n: remove hard coded english + make UI RTL safe

salmoneatenbybear:
Fixes #5955

PR #6000, opened 2025-08-23
fix: RTL search input placeholder alignment

### LINKED ISSUE #5999: Text Overlapping with Search Icon in Arabic (RTL) Layout
**Description:**
When the site language is switched to Arabic, the text in the search input field overlaps with the search icon. 

**Steps to Reproduce:**
- Log in to CircuitVerse.
- Go to Dashboard.
- Click on edit details.
- Go to Locale input and select Arabic language then save it.
- In the left side the text overlapping input field will show

**Expected Behavior:**
The placeholder text should be properly aligned and not overlap with the icon in RTL languages.

**Actual Behavior:**
The placeholder text overlaps with the search icon in Arabic (see attached screenshot).

**Screenshot:**
Image

akramcodez:
Fixes #5999  

Before: Text overlapped with search icon in Arabic locale.  
input text overlapping

After: Proper alignment with sufficient padding.  
Screenshot 2025-08-23 161140

- [x] Added proper PR title and linked the issue  
- [x] Performed a self-review of code  

Summary by CodeRabbit
---
- Fixed placeholder text alignment in RTL (Right-to-Left) locales (e.g., Arabic).  
- Added proper padding to ensure text does not overlap with the search icon.  
- No impact on LTR (Left-to-Right) layouts.  

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed misalignment of search icon and navbar so elements are correctly positioned across breakpoints.

* **Style**
  * Improved search-bar and navbar layout with adjusted alignment, spacing, and responsive positioning; search icon now vertically centered and consistently offset, and spacing increased on small screens for better readability.


salmoneatenbybear:
Can you check this file : 


app/components/admin/contest_close_modal_component.html.erb

In the context of the change made to the close btn in this PR : #5966

I think search button should switch it's postion from far right to far left for RTL language if I am not wrong. We should avoid using custom padding and use as much standard bootstrap styling as possible and ensure multi device compatibility, also pls share UI screenshots for most common screen sizes. That would be great. 

CC : @tachyons

akramcodez:
@salmoneatenbybear sorry I can't get it. I want to do same thing using bootstrap styling?

salmoneatenbybear:
> @salmoneatenbybear sorry I can't get it. I want to do same thing using bootstrap styling?

Instead of positioning with custom CSS, try using Bootstrap’s logical utilities (start-0, end-0, ps-, pe-) so the icon flips automatically between LTR and RTL.
Also, make sure the input uses consistent line-height/padding so text and the clear button stay vertically centered.
Bootstrap’s default .form-control sizing is a good reference here.

It should look something like this : 

image

image

akramcodez:
If I am using bootstrap classes like px-2 py-2 ps-4
image

it doesn’t give me enough space I actually need 35px. Normally, I would have to create a custom class for that. But instead, if I just use:
    padding-block: 7px;
    padding-inline: 10px;
    padding-inline-start: 35px;

then the problem is solved entirely.
   
image
image
image
image

So, I think in this case, instead of using Bootstrap styling, just adding these three lines is fine.

salmoneatenbybear:
@akramcodez okay let's wait for org admin to review. Thanks for the contribution

tachyons:
The search icon should be placed on left side on RTL interface

salmoneatenbybear:
> > @salmoneatenbybear sorry I can't get it. I want to do same thing using bootstrap styling?
> 
> Instead of positioning with custom CSS, try using Bootstrap’s logical utilities (start-0, end-0, ps-, pe-) so the icon flips automatically between LTR and RTL.
> Also, make sure the input uses consistent line-height/padding so text and the clear button stay vertically centered.
> Bootstrap’s default .form-control sizing is a good reference here.
> 
> It should look something like this : 
> 
> image
> 
> image

@tachyons is this what I suggested the correct way to do it ?

akramcodez:
@tachyons I fixed the position of the search icon, updated it with Bootstrap styling, and tested it across different screen sizes.

 image
image

Issue #643, opened 2019-11-23
If there are no favourites, show a message rather than showing a blank page.


PragatiVerma18:
**Describe the bug**

If there are no favourites added by a user, show a message rather than showing a blank page or either open the favourites page and there show a message saying there are no favourites selected by the user yet and give an option to view more projects by that author.

**To Reproduce**
1. Go to 'https://circuitverse.org/users/13353'
2. Click on 'View Favourites'
3. See error

**Expected behaviour**

An empty page doesn't look good. In the first place, it gives an impression of what is wrong with the page. So, better write that there re no favourites yet.

**Screenshots**

![1](https://user-images.githubusercontent.com/42115530/69483266-21c33a00-0e4b-11ea-80f3-7690e0d1e124.PNG)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 78.0.3904.108

**Are you working on this issue? (Yes/No)**
No. For GCI 2019.

sal2701:
@PragatiVerma18  There is an empty screen in projects section as well. We can add a message for Favourites as well as Projects. Also the message should be different for the account owner and a viewer. The owner should preferably receive a motivational message :smile: and a link to the simulator. What do you think?

PragatiVerma18:
Nice idea @sal2701!!! 😀 That would be great considering user experience. Even when there are favorites there should be a link to view more projects from the same author or a link to simulator to create some of their own. Just the proper linking of pages can work wonders for user experience here.

Issue #573, opened 2019-11-08
RECAPTCHA


dev9414:
Please implement a Recaptcha for the following activities:

Password Reset
First login failed (the second and following requests should have a recaptcha)
**Are you working on this? (Yes/No)**no,
This is for GCI Students

PR #6072, opened 2025-09-17
chore: fix embed failure

ThatDeparted2061:
Fixes [#643 ](https://github.com/CircuitVerse/cv-frontend-vue/issues/643)
Fixes [#573](https://github.com/CircuitVerse/cv-frontend-vue/issues/573)

 

- [ ] I have added proper PR title and linked to the issue
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -41,7 +41,6 @@ class ProjectsController < ApplicationController
       end
   end
 
-
   # GET /projects/1/edit
   def edit; end
 
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -31,7 +31,20 @@ class ProjectsController < ApplicationController
     @collaboration = @project.collaborations.new
     @admin_access = true
     commontator_thread_show(@project)
+
+    # Resolve simulator embed path
+    raw_data = @project.project_datum&.data
+    parsed_data = raw_data.present? ? JSON.parse(raw_data) : {}
+    sim_version = parsed_data["simulatorVersion"] || "legacy"
+
+    @embed_path =
+      if sim_version == "v0" || sim_version == "v1"
+        simulatorvue_path(@project)
+      else
+        simulator_path(@project)
+      end
   end
+
 
   # GET /projects/1/edit
   def edit; end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -33,12 +33,8 @@ class ProjectsController < ApplicationController
     commontator_thread_show(@project)
 
     # Resolve simulator embed path
-    raw_data = @project.project_datum&.data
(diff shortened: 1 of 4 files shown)


ThatDeparted2061:
@tachyons @aryanndwi123 
PTAL

ThatDeparted2061:
Okay the reason for using parse in the beginning of the code piece here is, so view simulator uses variable called simulator version which is stored in its data to set the value as v0 and V1. now when we are interacting with the API to store the circuits on the primary code base we divide the data that we get into two parts the  project metadata and the other parts of the project. The simulator vision is stored not in the primary table but rather in a text format which is why we are not able to access the simulator version directly from the code in show html.erb so to fix that we have two probable solutions the first one is the shorter one which I have raised it in this PR when parsing that text field the data is not in jsonb format or json that's why I had to parse it since it's in text format, I checked it on the rail server and postgres server. The Other method is basically where we convert that whole text column which we have in the database we create a new column and we transfer all of the text data into JSONB format and then remove the old column that's the second way but that is a super long way so @tachyons tell me which one is the way that we want our files to be and will do that, this one was the easier one so currently raised a PR for this.

tachyons:
Could you move the conditional logic to the controller. Something like this https://github.com/CircuitVerse/CircuitVerse/pull/4864/files

ThatDeparted2061:
Okay

app/controllers/projects_controller.rb
```diff
@@ -31,8 +31,21 @@ def show
     @collaboration = @project.collaborations.new
     @admin_access = true
     commontator_thread_show(@project)
+
+    # Resolve simulator embed path
+    raw_data = @project.project_datum&.data
+    parsed_data = raw_data.present? ? JSON.parse(raw_data) : {}
+    sim_version = parsed_data["simulatorVersion"] || "legacy"
```
tachyons: Move this logic to the project model

aryanndwi123 (approved):
LGTM

tachyons (changes requested):
Please fix test failures and lint failures

tachyons:
Tests are still failing, please ask for review once it is resolved

salmoneatenbybear:
@ThatDeparted2061 

I think the failures are because the files are missing, commit the simulator Vue files (index-cv.html) into public/simulatorvue/v0/ and public/simulatorvue/v1/.

And then try

ThatDeparted2061:
The current failure in reviewdog is not from my PR but its own breakdown
`XML syntax error on line 6: expected attribute name in element`
**CI is failing because of a reviewdog XML parse error**

ThatDeparted2061:
@tachyons, kindly refer to the logs of the ci failure

aryanndwi123:
There is only one failure(reviewdog) which has already been resolved in the latest commit in the main branch
cc: @tachyons

Issue #6437, opened 2025-12-06
🐞 Bug :  On click over the logo icon it does not redirect to Home page


ritiksingh3202:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The issue is that once you click on blog tab it opens but once you click on the logo icon in the blog tab it doesn't redirect you to the Home page of circuitverse instead it reloads the same page. I am attaching a video for the same.

### Steps To Reproduce

I feel i am potential candidate to solve the bug hours just after it is assigned to me. So i request maintainers to assign me this issue.

### Expected Behavior

The Expected behavior is that it should redirect you to the home page of circuitverse.

### Device Information [optional]

```markdown
- OS: Windows
- Browser: Chrome
- version: intel i7
```

### Are you working on this issue?

Yes

OshiSharma1222:
Hello,

I have verified this behavior and would like to work on resolving it.

The CircuitVerse logo in the Blog section currently reloads the same page instead of redirecting to the homepage. I believe this is due to the logo link in the Blog layout/template pointing to the Blog route rather than the root path. I will inspect the navigation component and update the reference so that it consistently redirects to / across the entire site.

If this seems appropriate, please assign this issue to me. I can begin working on the fix immediately and will submit a PR once completed.

Thank you

Pulkit7070:
I’ve carefully reviewed the issue and the expected behavior. I understand that clicking the logo from the blog page should redirect to the main CircuitVerse homepage but is currently reloading the blog instead.

I’ve previously worked on React routing and navigation-related bugs, and I’m confident I can identify and fix this issue efficiently. I’m ready to start working on it immediately if assigned.

Kindly consider assigning this issue to me. Looking forward to contributing

for my work you can check my github and here is my portfolio link : https://pulkit-saraf.vercel.app/

lets connect on x: https://x.com/PsudoKit

OshiSharma1222:
sorry but i am already working on this issue and i am about to end

ritiksingh3202:
Hi guys, i myself did identify the issue and already looking into it, you people focus on other issues.

UtkarsHMer05:
Hi guys...Working on this bug

Me-Priyank:
Outdated. Not a valid bug now

PR #6483, opened 2025-12-18
fix: logo in navbar now redirects to home page using root_url helper

UtkarsHMer05:
Fixes #8437

Changed the navbar logo link from hardcoded `href="/"` to `href="<%= root_url %>"` in the header partial to ensure the logo properly redirects to the CircuitVerse home page from all sections.

**What was wrong:**
The logo used `href="/"` which only reloads the current page instead of redirecting to the actual home page.

**What I fixed:**
Now using the Rails `root_url` helper which generates the correct absolute URL to the application root, ensuring proper navigation from any page including blog sections.

N/A - Functional fix, no visual changes to UI

---

- [x] No, I didn't used AI assistance 

**Problem:** The logo link was hardcoded as `href="/"` which doesn't properly redirect to the home page it just reloads the current URL path.

**Solution:** I replaced the hardcoded `/` with Rails' `root_url` helper method. This helper generates the full URL to the application's root route as defined in `config/routes.rb`.

**Why this approach:** Using `root_url` instead of `root_path` ensures we get an absolute URL (e.g., `https://circuitverse.org/`) rather than a relative path, which is important for proper redirection across different subdomains or sections.

**Alternative considered:** Could have used `root_path` but `root_url` is more appropriate here as it provides an absolute URL, making the redirect more reliable.

**How it works:** When a user clicks the logo, the browser navigates to the URL returned by `root_url`, which is the main CircuitVerse homepage, regardless of which section they're currently viewing.

---

- [ ] If it is a core feature, I have added thorough tests (N/A - simple link fix)

## Summary by CodeRabbit

* **Chores**
  * Updated navbar brand link routing to use Rails conventions for improved maintainability.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,6 +1,6 @@
 <nav class="navbar navbar-expand-lg bg-white">
   <div class="container">
-    <a class="navbar-brand" href="/">
+    <a class="navbar-brand" href="<%= root_url %>">
       <%= image_tag("CircuitVerse.svg", class: "navbar-logo", alt: "CircuitVerse Logo") %>
     </a>
     <div class="d-flex d-lg-none align-items-center">


senutpal (commented):
The issue number `#8437` doesn’t appear to exist. I believe this is referring to [#6437](https://github.com/CircuitVerse/CircuitVerse/issues/6437).

Also, `href="/"` does not reload the current page, it navigates to the root of the current domain (for example, `circuitverse.org/`). So if a user is on `/users/123`, clicking a link to `/` correctly takes them to the home page. The blog example isn’t really applicable here since `blog.circuitverse.org` is a separate domain altogether.

Using `root_url` instead of `/` may break the intended behavior, as clicking the logo should navigate to the root of the current domain, not redirect across domains.

Additionally, the issue opener  @ritiksingh3202  mentioned that they were already working on this issue, but a pull request was opened anyway.

**Note:** I’m commenting as a community contributor, not a maintainer.

Me-Priyank:
Outdated. Not a valid bug now

Issue #6620, opened 2026-01-12
UI: Navbar appears left-aligned instead of centered


Nid-debug:
### Is your feature request related to a problem? Please describe

The navbar on the CircuitVerse main website appears to be mostly aligned towards the left side of the page instead of being centered. This affects the visual balance of the layout, especially on larger screen sizes.

<img width="1919" height="948" alt="Image" src="https://github.com/user-attachments/assets/e6a36f90-7c55-45c7-bbcc-4466f7fed7c2" />

### Describe the solution you'd like

This issue might be related to the navbar container width or flex alignment. Adjusting justify-content , margin: auto, or using a centered container could resolve this issue.

### Additional context

I would be happy to work on this issue if it is approved and guidance is provided.

### Are you working on this?

Yes

banavathsrikanth7:
I want to work on this because iam very much interested in frontend development

himanshumoral:
Hi! I noticed there is an open PR (#6622) for this, but it has pending changes requested by the bot and hasn't been updated for a while. Since the issue is still open, I can fix the navbar alignment properly without breaking the layout. Can I take this up?

PR #6622, opened 2026-01-12
fix(issue): Address issue #6620 - UI: Navbar appears left-aligned instead of centered

Harshitshukla0208:
This PR was to address issue #6620.

The navbar is currently left-aligned due to the CSS properties applied to its container. To center the navbar, we need to modify the CSS to use 'justify-content: center' in a flexbox layout. This will ensure that the items within the navbar are evenly distributed and centered within the available space.

## Summary by CodeRabbit

* **Style**
  * Improved visual centering of the Overview section header in the documentation interface.

### DIFF
--- a/apidoc/v1/index.html
+++ b/apidoc/v1/index.html
@@ -389,7 +389,7 @@
     </div>
     <div class="page-wrapper">
       <div class="dark-box"></div>
-      <div class="content">
+      <div class="content" style="display: flex; justify-content: center;">
         <h1 id='overview'>Overview</h1>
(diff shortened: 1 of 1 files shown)


Aaditya-2407 (changes requested):
Hey @Harshitshukla0208, while this centers the header, adding display: flex to the .content div actually breaks the core two-column layout of the API documentation. This container relies on specific float and margin rules to keep the text and code blocks side-by-side.

I'd recommend moving the centering logic to a dedicated CSS rule in the stylesheet or using a separate wrapper so it doesn't interfere with the layout of the rest of the documentation.

naman79820:
stale pr

PR #6623, opened 2026-01-12
fix(issue): Address issue #6596 - ActionView::Template::Error: No route matches {:action=>"show", :controller=>"tags", :tag=>""}, missing required keys: [:tag] (ActionView::Template::Error)

Harshitshukla0208:
This PR was to address issue #6596.

The error indicates that there is an attempt to generate a URL for a tag with an empty string as the tag value. This occurs in the view file where tags are iterated over. To prevent this error, we should ensure that we only generate URLs for tags that have a valid name. We can add a conditional check to skip any tags that do not have a name before attempting to generate the URL.

## Summary by CodeRabbit

* **Refactor**
  * Optimized tag rendering logic to improve code efficiency while maintaining the same user experience.

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -58,7 +58,7 @@
         <div class="row center-row pt-0 pb-4 px-5">
           <div class="d-block mb-4 pl-0 text-left w-75">
             <% if !@project.tags.empty? %>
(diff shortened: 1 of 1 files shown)


naman79820:
stale pr

PR #6807, opened 2026-02-01
fix: replace invalid jQuery ajax 'failure' callback with 'error'

laky241:

Fixed broken error handling in jQuery $.ajax calls across the codebase

jQuery's $.ajax() supports `success`, `error`, `complete` there is no `failure` callback Four ajax calls were using `failure` which meant their error handlers **never fired**

**Changes:**
1. `public/js/data.js` 
2. `simulator/src/ux.js` 
3. `simulator/src/Verilog2CV.js`

---

Was reading through the ajax calls in data.js and noticed `failure:` being used as a callback. jQuery $.ajax only supports `error:` for handling failed requests — `failure` is not a valid option so it just gets ignored silently.
 While fixing them I also spotted that the support email in the error messages was missing a letter (`ciruitverse` instead of `circuitverse`)

For Verilog2CV.js, the ajax call already had a proper `error:` handler, so the `failure:` block below it was just dead code removed it entirely

### DIFF
--- a/public/js/data.js
+++ b/public/js/data.js
@@ -508,7 +508,7 @@ function save() {
                 $('.loadingIcon').fadeOut();
                 localStorage.removeItem("recover");
             },
-            failure: function(err) {
+            error: function(err) {
                 showMessage("There was an error, we couldn't save to our servers")
                 $('.loadingIcon').fadeOut();
             }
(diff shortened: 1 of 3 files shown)


naman79820:
no issue attached...

PR #6909, opened 2026-02-15
fix(groups): Fix params.expect error in promote/demote member modals

### LINKED ISSUE #6913: 🐞 Bug: Promote/Demote member modals fail with params.expect error
## Summary
Promoting a group member to mentor or demoting a mentor to member via the confirmation modals on the group page fails with an `ActionController::ParameterMissing` error.

## Steps to Reproduce

1. Log in as a group admin or owner  
2. Navigate to a group page:  /groups/:id
3. Click **"Make Mentor"** on a member  
OR  
Click **"Make Member"** on a mentor  
4. In the confirmation modal, click the confirm button  

## Expected Behavior

The group member's role should be updated successfully, and the following flash message should appear: 
Group member was successfully updated.

## Actual Behavior

The request fails with:ActionController::ParameterMissing: param is missing or the value is empty: group_member

## Problem

The button_to helper generates a form where parameters are not properly nested under the group_member key in the way expected by the controller:

```
params.expect(group_member: [:mentor])
# OR
params.require(:group_member).permit(:mentor)
```

This causes Rails to raise:

`ActionController::ParameterMissing`

Heyy-Himanshuu:
Fixes #6913 

## Summary

This PR fixes the `ActionController::ParameterMissing` error when promoting a member to mentor or demoting a mentor to member from the confirmation modals.

The issue was caused by `button_to`, which did not structure params correctly for `GroupMembersController#update`, where `params.expect(group_member: [:mentor])` is required.

## Changes made

- Replaced `button_to` with `form_with` in promote/demote confirmation modals
- Added hidden field `group_member[mentor]` to ensure correct param structure
- Added explicit form IDs (`promote-member-form`, `demote-member-form`)
- Updated `groups_controller.js` to dynamically set form action using form IDs instead of fragile DOM traversal
- Updated system tests to work correctly with Bootstrap modal and new form structure

These changes ensure the correct params are submitted and fix the failing system tests.

---

- [x] Yes, I used AI assistance
- [ ] No

- [x] I can explain the purpose and logic of each change
- [x] I have tested the implementation locally
- [x] I modified the AI output to follow project conventions

**Implementation approach**

The issue was caused by improper parameter structure from `button_to`. I replaced it with `form_with` and explicitly added hidden fields to ensure Rails receives params in the expected format. I also updated JavaScript to target forms using stable IDs and updated system tests to align with the new modal behavior.

---

- [x] I have linked the issue
- [x] I have performed a self-review
- [x] I understand all changes made
- [x] I have tested the changes locally
- [x] My code follows project guidelines


Heyy-Himanshuu:
@Nihal4777

magic-peach (commented):
The root cause here is just that `false.to_param` returns `""` in Rails, causing `params.expect` to treat the value as missing. PR #7109 fixes this with a one-line change in each template. The `form_with` rewrite here works, but it's a much larger surface area for the same fix, and the system tests have `sleep 0.5` calls that make them flaky. Worth considering whether this scope is warranted given the simpler alternative.

PR #7109, opened 2026-03-02
fix: resolve ParameterMissing in promote/demote member modals (#6913)

### LINKED ISSUE #6913: 🐞 Bug: Promote/Demote member modals fail with params.expect error

Uday9909:
false.to_param returns an empty string ("") in Ruby/Rails.  

With the Rails 8 upgrade, `params.expect(group_member: [:mentor])` treats empty values as missing and raises `ActionController::ParameterMissing`.  

When demoting a mentor (`mentor: false`), the form submits:

group_member[mentor] = ""

Since `false.to_param → ""`, `params.expect` raises instead of permitting the value.

### Fix

Reverted `group_member_update_params` to:

```ruby
params.require(:group_member).permit(:mentor)
```

require.permit correctly allows empty string values for boolean fields and maintains strong parameter protection for the :mentor attribute.

This change is scoped only to the #update action of GroupMembersController, where toggling mentor between true and false is intended behavior.

Tests

Added a regression test for the demote flow (mentor: false) to ensure both promote and demote actions work correctly.

Closes #6913

---

 Yes, I used AI assistance (continue below)

 I have reviewed every single line of the AI-generated code

 I can explain the purpose and logic of each function/component I added

 I have tested edge cases and understand how the code handles them

 I have modified the AI output to follow this project's coding standards and conventions

---

### Explain your implementation approach:

The issue was caused by a behavioral change introduced during the Rails 8 upgrade when `params.require(...).permit(...)` was replaced with `params.expect(...)`.

Unlike `require.permit`, `expect` raises `ParameterMissing` if a value is blank. Since `false.to_param` serializes to an empty string, demoting a mentor resulted in `group_member[mentor]` being treated as missing.

I considered modifying the form structure to explicitly submit `"false"` instead of an empty string. However, that would require view changes and increase the scope of the fix unnecessarily.

Since the controller already explicitly permits only the `:mentor` attribute and authorization is enforced via Pundit (`authorize @group_member, :primary_mentor?`), reverting to `require.permit` is the smallest, safest, and most consistent solution.

This keeps the fix localized to the update action without affecting other parameter handling logic.
Checklist before requesting a review
- [x] I can explain the purpose of every function, class, and logic block I added

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for demoting a mentor to member role.

* **Bug Fixes**
  * Fixed parameter handling in member role management modals for consistent mentor attribute submission.

### DIFF
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -119,7 +119,13 @@ class GroupMembersController < ApplicationController
     # which would allow changing users of arbitrary groups since
     # the check_access happens before the group is changed and passes
     def group_member_update_params
-      params.require(:group_member).permit(:mentor)
+      # false.to_param serializes to "" which Rails 8's params.expect
+      # treats as missing, raising ActionController::ParameterMissing.
+      # Normalize blank to "false" so the value passes through expect.
+      if params[:group_member]&.key?(:mentor) && params[:group_member][:mentor].blank?
+        params[:group_member][:mentor] = "false"
+      end
+      params.expect(group_member: [:mentor])
     end
 
     def check_access
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -119,12 +119,6 @@ class GroupMembersController < ApplicationController
     # which would allow changing users of arbitrary groups since
     # the check_access happens before the group is changed and passes
     def group_member_update_params
-      # false.to_param serializes to "" which Rails 8's params.expect
-      # treats as missing, raising ActionController::ParameterMissing.
-      # Normalize blank to "false" so the value passes through expect.
-      if params[:group_member]&.key?(:mentor) && params[:group_member][:mentor].blank?
-        params[:group_member][:mentor] = "false"
-      end
       params.expect(group_member: [:mentor])
     end
 
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -119,7 +119,7 @@ class GroupMembersController < ApplicationController
     # which would allow changing users of arbitrary groups since
     # the check_access happens before the group is changed and passes
     def group_member_update_params
-      params.expect(group_member: [:mentor])
+      params.require(:group_member).permit(:mentor)
     end
 
     def check_access
--- a/app/views/groups/_add_member_modal.html.erb
+++ b/app/views/groups/_add_member_modal.html.erb
@@ -21,7 +21,7 @@
             <%= form.hidden_field :group_id, id: :group_member_group_id %>
           </div>
           <div class="field">
-            <%= form.hidden_field :mentor, value: false %>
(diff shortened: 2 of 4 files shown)


Uday9909:
@Asrani-Aman Could you please review this aswell when you get a chance? Thanks!

app/controllers/group_members_controller.rb
```diff
@@ -119,6 +119,12 @@ def group_member_params
     # which would allow changing users of arbitrary groups since
     # the check_access happens before the group is changed and passes
     def group_member_update_params
+      # false.to_param serializes to "" which Rails 8's params.expect
+      # treats as missing, raising ActionController::ParameterMissing.
+      # Normalize blank to "false" so the value passes through expect.
+      if params[:group_member]&.key?(:mentor) && params[:group_member][:mentor].blank?
+        params[:group_member][:mentor] = "false"
+      end
```
tachyons: Could you update frontend to send proper params instead of making band aid fix here ?

Uday9909: @tachyons Updated in #ddad8fb  moved the fix to the frontend views. Both _promote_member_confirmation_modal and _add_member_modal now send "false" as an explicit string instead of the boolean false, and the controller normalization has been removed.

tachyons:
Could you add before and after video ?

Uday9909:
> Could you add before and after video ?

Honestly I traced and fixed this through code reading and the issue description rather than a running local instance ,the Rails 8 changelog and the stack trace in the issue made the root cause clear. I've been working on getting Docker set up but it's been taking longer than expected. The regression test covers the exact demote flow and CodeRabbit has approved the logic. Happy to record the video once my environment is stable but didn't want to hold up the fix.

magic-peach (approved):
Correct minimal fix. `false.to_param` returns `""` in Rails, so `params.expect(group_member: [:mentor])` sees an empty value and raises. Switching to the string `"false"` is the right fix — Rails will still cast it to the boolean `false` when writing to the boolean `mentor` column. The controller test for the demote path is a good addition.

PR #7186, opened 2026-03-20
fix: remove duplicate update! call in Api::V1::AssignmentsController#update

### LINKED ISSUE #7179: Bug: Api::V1::AssignmentsController#update calls update! and update twice with identical params
## Describe the bug

`Api::V1::AssignmentsController#update` applies the same parameter set twice to the database — first via `update!` (which raises on failure) and then immediately again via `update` (which returns a boolean). This means every successful PATCH triggers two sequential SQL UPDATE statements, doubles every `after_update` callback (emails, notifications, deadline jobs), and the error-handling branch (`invalid_resource!`) can never be reached because `update!` would already have raised before it.

## Root cause

`app/controllers/api/v1/assignments_controller.rb`, lines 38–45:

```ruby
def update
  @assignment.update!(assignment_update_params)   # ← first write, raises on error
  if @assignment.update(assignment_update_params) # ← identical second write
    render json: Api::V1::AssignmentSerializer.new(@assignment, @options), status: :accepted
  else
    invalid_resource!(@assignment.errors)          # ← unreachable
  end
end
```

## Steps to reproduce

1. Authenticate as a group mentor via `POST /api/v1/auth/login`.
2. Send `PATCH /api/v1/assignments/:id` with a valid body (e.g. `{"name": "test"}`).
3. Observe two UPDATE queries in the Rails log for a single request.
4. Observe assignment `after_commit` callbacks (e.g. `AssignmentMailer.update_assignment_email`) fire twice per request.

## Expected behavior

One SQL UPDATE per PATCH request; `after_commit` callbacks fire exactly once; invalid params return a proper 422 response via `invalid_resource!`.

## Actual behavior

Two SQL UPDATE queries are executed. `after_commit` callbacks fire twice. The `invalid_resource!` error path is dead code because `update!` raises before the conditional is evaluated.

## Proposed fix

Remove the redundant `update!` call and rely solely on the conditional `update`:

```ruby
def update
  if @assignment.update(assignment_update_params)
    render json: Api::V1::AssignmentSerializer.new(@assignment, @options), status: :accepted
  else
    invalid_resource!(@assignment.errors)
  end
end
```

## Impact

Every teacher who updates an assignment via the API receives duplicate notification emails. In groups with many students, `send_update_mail` fires twice, doubling SMTP load. The `AssignmentDeadlineSubmissionJob` is also enqueued twice per update.

magic-peach:
## Summary

- Removes the redundant `@assignment.update!(assignment_update_params)` call that preceded the `if @assignment.update(...)` conditional in `Api::V1::AssignmentsController#update`
- Every successful PATCH was previously firing all `after_commit` callbacks twice (duplicate emails, duplicate `AssignmentDeadlineSubmissionJob` enqueues)
- The `invalid_resource!` error path was dead code because `update!` would raise before the conditional was evaluated

## What changed and why

**Before:**
```ruby
def update
  @assignment.update!(assignment_update_params)   # ← fires all after_commit callbacks
  if @assignment.update(assignment_update_params) # ← fires them again
    render json: ..., status: :accepted
  else
    invalid_resource!(@assignment.errors)         # ← unreachable
  end
end
```

**After:**
```ruby
def update
  if @assignment.update(assignment_update_params)
    render json: ..., status: :accepted
  else
    invalid_resource!(@assignment.errors)
  end
end
```

## How to test

1. Run the existing spec: `bundle exec rspec spec/requests/api/v1/assignments_controller/update_spec.rb`
2. All four existing contexts (unauthenticated, unauthorized, not found, invalid params, successful update) should pass.
3. Add a test that counts `after_commit` callback invocations and verify it fires exactly once per PATCH.

Closes #7179

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API error handling for assignment updates so validation failures return consistent error responses instead of triggering exceptions.

### DIFF
--- a/app/controllers/api/v1/assignments_controller.rb
+++ b/app/controllers/api/v1/assignments_controller.rb
@@ -36,7 +36,6 @@ class Api::V1::AssignmentsController < Api::V1::BaseController
 
   # PATCH /api/v1/assignments/:id
   def update
-    @assignment.update!(assignment_update_params)
     if @assignment.update(assignment_update_params)
       render json: Api::V1::AssignmentSerializer.new(@assignment, @options), status: :accepted
     else


magic-peach:
@vedant-jain03 could you please review my pr

PR #7187, opened 2026-03-20
fix: scope AssignmentsController#index to current user's groups

### LINKED ISSUE #7180: Security: AssignmentsController#index leaks all assignments to any authenticated user
## Describe the bug

`AssignmentsController#index` returns every assignment in the database to any logged-in user, regardless of whether that user belongs to the relevant group. A student in Group A can access assignment names, deadlines, descriptions, and grading scales for every group in the system by hitting `GET /assignments`.

## Root cause

`app/controllers/assignments_controller.rb`, lines 19–21:

```ruby
def index
  @assignments = Assignment.all  # ← no scope, no authorization
end
```

The controller uses `before_action :authenticate_user!` (requiring login) but there is no `authorize` call and no scope limiting results to the current user's groups. Every other controller action calls `authorize @assignment`, but `index` does not.

## Steps to reproduce

1. Create two groups — Group A (user A as mentor) and Group B (user B as mentor), each with a private assignment.
2. Log in as user A.
3. Send `GET /assignments`.
4. Observe that Group B's assignments are returned in the response.

## Expected behavior

`GET /assignments` should return only assignments belonging to groups the current user is a member of, or should require admin access.

## Actual behavior

All assignments from all groups across the entire application are returned to any authenticated user.

## Proposed fix

Scope the query to the current user's groups and add a Pundit policy check:

```ruby
def index
  @assignments = policy_scope(Assignment)
  # In AssignmentPolicy::Scope:
  #   scope.where(group_id: user.groups.select(:id))
end
```

## Impact

Any registered CircuitVerse user can enumerate assignment names, deadlines, descriptions, and grading configurations for every classroom on the platform, including private educational content from unrelated institutions.

magic-peach:
## Summary

- Adds `AssignmentPolicy::Scope` that returns only assignments belonging to the current user's groups (or all assignments for admins)
- Replaces `Assignment.all` in `AssignmentsController#index` with `policy_scope(Assignment)`

## What changed and why

`GET /assignments` was returning every assignment across the entire platform to any logged-in user. A student in Group A could see assignment names, deadlines, descriptions, and grading scales for Group B's classes. This is a data-exposure bug: `before_action :authenticate_user!` only requires login; there was no authorization check and no scoping.

**`app/policies/assignment_policy.rb`** — new `Scope` class:
```ruby
class Scope < Scope
  def resolve
    if user.admin?
      scope.all
    else
      scope.where(group_id: user.groups.select(:id))
    end
  end
end
```

**`app/controllers/assignments_controller.rb`** — controller change:
```ruby
def index
  @assignments = policy_scope(Assignment)
end
```

## How to test

1. Create two groups with different mentors and assignments.
2. Log in as mentor A — `GET /assignments` should return only Group A's assignments.
3. Log in as admin — `GET /assignments` should return all assignments.
4. Run: `bundle exec rspec spec/requests/` — existing request specs should pass.

Closes #7180

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted assignment visibility so regular users only see assignments tied to their groups, preventing exposure of unrelated records.
  * Administrators retain full visibility of all assignments to preserve management capabilities.
  * This improves data access control and ensures users only access assignments relevant to their group memberships.

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -17,7 +17,7 @@ class AssignmentsController < ApplicationController
   # GET /assignments
   # GET /assignments.json
   def index
-    @assignments = Assignment.all
+    @assignments = policy_scope(Assignment)
   end
 
   # GET /assignments/1
--- a/app/policies/assignment_policy.rb
+++ b/app/policies/assignment_policy.rb
@@ -3,6 +3,16 @@
 class AssignmentPolicy < ApplicationPolicy
   attr_reader :user, :assignment
 
+  class Scope < Scope
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope.where(group_id: user.groups.select(:id))
+      end
+    end
+  end
+
   def initialize(user, assignment)
     super
     @user = user


magic-peach:
@Asrani-Aman could you please reviw my pr

PR #7188, opened 2026-03-20
fix: add includes(:user) to Assignment notification callbacks to avoid N+1 queries

### LINKED ISSUE #7181: Performance: N+1 queries in Assignment notification callbacks (notify_recipient, send_new_assignment_mail, send_update_mail)
## Describe the bug

Three `after_commit` callbacks in `Assignment` iterate over `group.group_members` and call `.user` on each member inside the loop, producing one extra SQL query per member. For a group with 100 students, creating or updating an assignment fires 100 additional `SELECT "users".* WHERE "users"."id" = ?` queries.

## Root cause

`app/models/assignment.rb`, lines 32–49:

```ruby
def notify_recipient                           # line 31
  group.group_members.each do |group_member|  # loads members without user
    NewAssignmentNotification.with(assignment: self)
      .deliver_later(group_member.user)        # N+1: SELECT user for each member
  end
end

def send_new_assignment_mail                   # line 37
  group.group_members.each do |group_member|
    AssignmentMailer.new_assignment_email(group_member.user, self).deliver_later  # N+1
  end
end

def send_update_mail                           # line 43
  return unless status != "closed"
  group.group_members.each do |group_member|
    AssignmentMailer.update_assignment_email(group_member.user, self).deliver_later  # N+1
  end
end
```

All three methods load `GroupMember` records without including the associated `User`, then immediately call `.user` on each. Compare with `notify_members_of_grade` on line 115, which correctly uses `.includes(:user)`.

## Steps to reproduce

1. Create a group with 50+ students.
2. Create a new assignment (triggers `notify_recipient` and `send_new_assignment_mail`).
3. Inspect the Rails log or a query counter: 50+ `SELECT "users".*` queries fire for each callback.

## Expected behavior

Each callback should issue a single JOIN or eager-load query, regardless of group size.

## Actual behavior

`2 * N` extra queries per assignment creation (notify + mail) and `N` extra queries per assignment update.

## Proposed fix

Add `.includes(:user)` to each loop:

```ruby
def notify_recipient
  group.group_members.includes(:user).each do |group_member|
    NewAssignmentNotification.with(assignment: self).deliver_later(group_member.user)
  end
end

def send_new_assignment_mail
  group.group_members.includes(:user).each do |group_member|
    AssignmentMailer.new_assignment_email(group_member.user, self).deliver_later
  end
end

def send_update_mail
  return unless status != "closed"
  group.group_members.includes(:user).each do |group_member|
    AssignmentMailer.update_assignment_email(group_member.user, self).deliver_later
  end
end
```

## Impact

Every assignment creation in a large class causes a query storm that can time out under load (see production timeout errors in #7163 / #7151). Affects all teachers with classes of more than ~30 students.

magic-peach:
## Summary

- Adds `.includes(:user)` to `notify_recipient`, `send_new_assignment_mail`, and `send_update_mail` in `app/models/assignment.rb`
- Reduces per-assignment-create/update query count from `2N+2` to `4` (where N = group member count)

## What changed and why

All three callbacks iterated `group.group_members` without eager-loading the `user` association, then called `.user` on each record — one SQL query per member.

```ruby
# Before (N+1)
group.group_members.each do |group_member|
  AssignmentMailer.new_assignment_email(group_member.user, self).deliver_later
end

# After (1 query)
group.group_members.includes(:user).each do |group_member|
  AssignmentMailer.new_assignment_email(group_member.user, self).deliver_later
end
```

The `notify_members_of_grade` method on line 115 already used `.includes(:user)` correctly. This PR applies the same pattern to the three callbacks that didn't.

## How to test

```bash
bundle exec rspec spec/models/assignment_spec.rb
```

You can also verify with a query counter:
```ruby
group = Group.first
group.group_members.count # => 50
assignment = group.assignments.create!(name: "test", deadline: 1.week.from_now)
# After this fix: 4 queries total (group_members + users JOIN + two mailer saves)
# Before: 50+ queries
```

Closes #7181

## Summary by CodeRabbit

* **Performance**
  * Improved efficiency of assignment notifications and emails to group members, reducing delivery overhead and potential delays while keeping recipients and message content unchanged.


magic-peach:
@senbo1 could you please review my pr

PR #7189, opened 2026-03-20
fix: block grade writes when assignment grades_finalized is true

### LINKED ISSUE #7182: Bug: GradePolicy does not enforce grades_finalized flag — mentors can overwrite grades after finalization
## Describe the bug

Once an assignment's `grades_finalized` column is set to `true`, the grading form is hidden from the UI. However, the backend `GradesController#create` action only calls `authorize @grade, :mentor?`, and `GradePolicy#mentor?` has no check for `grades_finalized`. A mentor can still POST new grades or overwrite existing ones via direct API call after finalization, violating the intention of the finalization workflow.

## Root cause

`app/policies/grade_policy.rb`, lines 11–14:

```ruby
def mentor?
  grade.assignment&.group&.primary_mentor_id == user.id \
  || grade.assignment&.group&.group_members&.exists?(user_id: user.id, mentor: true)
  # ← no check for grade.assignment&.grades_finalized
end
```

`app/controllers/grades_controller.rb`, line 13:

```ruby
authorize @grade, :mentor?  # ← only checks mentor role, not finalization state
```

`AssignmentPolicy#can_be_graded?` (line 46) does check the deadline, but it is only used by the view layer (`app/views/assignments/show.html.erb` line 51/104) to show/hide the form — it is never called during the save path.

## Steps to reproduce

1. Create an assignment, grade all submissions, and set `assignment.grades_finalized = true` (via Rails console or admin).
2. As the mentor, send `POST /grades` with `grade[project_id]=X&grade[assignment_id]=Y&grade[grade]=100`.
3. Observe that the grade record is created or updated without error.

## Expected behavior

`GradesController#create` should return 403 Forbidden when `assignment.grades_finalized` is `true`.

## Actual behavior

Grades can be created or overwritten even after finalization; the `grades_finalized` flag is silently bypassed.

## Proposed fix

Add a finalization guard to `GradePolicy#mentor?`:

```ruby
def mentor?
  return false if grade.assignment&.grades_finalized?
  grade.assignment&.group&.primary_mentor_id == user.id \
  || grade.assignment&.group&.group_members&.exists?(user_id: user.id, mentor: true)
end
```

## Impact

Affects any institution relying on `grades_finalized` for grade integrity. A malicious or mistaken mentor can silently alter official grades after a course closes.

magic-peach:
## Summary

- Adds a `grades_finalized?` guard to `GradePolicy#mentor?` so that `GradesController#create` (and `destroy`) return 403 Forbidden when grades are locked

## What changed and why

`assignments.grades_finalized` is a boolean flag meant to lock grades after a course ends. The UI respects it (hides the grading form), but `GradesController#create` calls only `authorize @grade, :mentor?`, which previously had no finalization check. A mentor could bypass the UI and POST to `/grades` to create or overwrite locked grades.

**Before (`app/policies/grade_policy.rb`)**:
```ruby
def mentor?
  grade.assignment&.group&.primary_mentor_id == user.id \
  || grade.assignment&.group&.group_members&.exists?(user_id: user.id, mentor: true)
end
```

**After**:
```ruby
def mentor?
  return false if grade.assignment&.grades_finalized?

  grade.assignment&.group&.primary_mentor_id == user.id \
  || grade.assignment&.group&.group_members&.exists?(user_id: user.id, mentor: true)
end
```

## How to test

```bash
bundle exec rspec spec/requests/api/v1/grades_controller/
```

Manual test:
1. Create assignment and finalize grades (`assignment.update!(grades_finalized: true)`).
2. POST to `/grades` as the mentor — expect 403 Forbidden.
3. Set `grades_finalized: false` — expect normal grade creation.

Closes #7182

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed permissions to prevent mentors from making changes to assignments after grades are finalized.

### DIFF
--- a/app/policies/grade_policy.rb
+++ b/app/policies/grade_policy.rb
@@ -9,6 +9,8 @@ class GradePolicy < ApplicationPolicy
   end
 
   def mentor?
+    return false if grade.assignment&.grades_finalized?
+
     grade.assignment&.group&.primary_mentor_id == user.id \
     || grade.assignment&.group&.group_members&.exists?(user_id: user.id, mentor: true)
   end


magic-peach:
@Asrani-Aman @tachyons @senbo1 @vedant-jain03 could someone please review my pr

PR #7190, opened 2026-03-20
fix: add partial unique index on assignments.lti_consumer_key for LTI 1.1 lookups

### LINKED ISSUE #7183: Performance: Missing database index on assignments.lti_consumer_key causes full table scan during LTI 1.1 auth
## Describe the bug

The LTI 1.1 launch flow looks up assignments by `lti_consumer_key` to validate the OAuth signature. The `assignments` table has no index on this column, so every LTI launch performs a full sequential scan of the entire `assignments` table.

## Root cause

`db/schema.rb`, lines 125–155 (assignments table definition):

```ruby
create_table "assignments", force: :cascade do |t|
  t.string "name"
  t.datetime "deadline", null: false
  ...
  t.string "lti_consumer_key"     # ← no index
  t.string "lti_shared_secret"
  ...
  t.index ["group_id"],           name: "index_assignments_on_group_id"
  t.index ["lti_deployment_id"],  name: "index_assignments_on_lti_deployment_id"
  # lti_consumer_key has no corresponding index entry
end
```

The LTI controller queries `Assignment.find_by(lti_consumer_key: params[:oauth_consumer_key])` on every LTI launch. With thousands of assignments in production, this forces a sequential scan.

## Steps to reproduce

1. In a Rails console on a large dataset, run:
   ```ruby
   EXPLAIN ANALYZE Assignment.find_by(lti_consumer_key: "some_key")
   ```
2. Observe `Seq Scan on assignments` in the output instead of `Index Scan`.

## Expected behavior

LTI consumer key lookups should use an index scan (O(log n)) regardless of table size.

## Actual behavior

Full sequential scan (O(n)) on every LTI 1.1 authentication request.

## Proposed fix

Generate a migration to add a unique index:

```ruby
class AddIndexToAssignmentsLtiConsumerKey < ActiveRecord::Migration[8.0]
  def change
    add_index :assignments, :lti_consumer_key, unique: true,
              where: "lti_consumer_key IS NOT NULL",
              name: "index_assignments_on_lti_consumer_key"
  end
end
```

`lti_consumer_key` is generated with `SecureRandom.hex(4)` per assignment so uniqueness is the correct constraint here. The partial index (`WHERE NOT NULL`) avoids indexing the majority of assignments that don't use LTI 1.1.

## Impact

Every LTI 1.1 launch by every student in every LMS-integrated classroom performs a full table scan. As the platform grows this will be a progressive performance regression visible in query timeout alerts (similar to #7163).

magic-peach:
## Summary

- Adds a partial unique index on `assignments.lti_consumer_key` (`WHERE lti_consumer_key IS NOT NULL`) via a new migration

## What changed and why

LTI 1.1 authentication calls `Assignment.find_by(lti_consumer_key: params[:oauth_consumer_key])` on every LTI launch. The column had no index, so the database performed a sequential scan of the entire `assignments` table for every student LTI login.

The index is:
- **Partial** (`WHERE NOT NULL`): the vast majority of assignments don't use LTI 1.1, so this avoids indexing them and keeps the index small
- **Unique**: `lti_consumer_key` is generated with `SecureRandom.hex(4)` per-assignment and must be unique for OAuth signature validation to work correctly

**Migration file:** `db/migrate/20260320000001_add_index_to_assignments_lti_consumer_key.rb`

```ruby
add_index :assignments, :lti_consumer_key,
          unique: true,
          where: "lti_consumer_key IS NOT NULL",
          name: "index_assignments_on_lti_consumer_key"
```

## How to test

```bash
bundle exec rails db:migrate
# Then verify in psql:
# \d assignments  => should show the new index
bundle exec rspec spec/requests/lti_spec.rb
```

To verify the index is used:
```sql
EXPLAIN ANALYZE SELECT * FROM assignments WHERE lti_consumer_key = 'abc123';
-- Should show: Index Scan using index_assignments_on_lti_consumer_key
```

Closes #7183

## Summary by CodeRabbit

* **Chores**
  * Database index improvements for assignment records to enhance data handling and system performance.

### DIFF
--- a/db/migrate/20260320000001_add_index_to_assignments_lti_consumer_key.rb
+++ b/db/migrate/20260320000001_add_index_to_assignments_lti_consumer_key.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIndexToAssignmentsLtiConsumerKey < ActiveRecord::Migration[8.0]
+  def change
+    add_index :assignments, :lti_consumer_key,
+              unique: true,
+              where: "lti_consumer_key IS NOT NULL",
+              name: "index_assignments_on_lti_consumer_key"
+  end
+end


magic-peach:
@vedant-jain03 could you please review my pr

PR #7191, opened 2026-03-20
fix: use find_by in AssignmentDeadlineVerificationJob to avoid RecordNotFound on deleted assignments

### LINKED ISSUE #7184: Bug: AssignmentDeadlineVerificationJob raises ActiveRecord::RecordNotFound when assignment is deleted while queued
## Describe the bug

`AssignmentDeadlineVerificationJob#perform` uses `Assignment.find(assignment_id)`, which raises `ActiveRecord::RecordNotFound` if the assignment is deleted before the job runs. The exception propagates unhandled, causing the job to fail, log an error, and retry indefinitely (depending on the queue adapter's retry policy). Its sibling job `AssignmentDeadlineSubmissionJob` correctly uses `find_by` with an early return to handle this case.

## Root cause

`app/jobs/assignment_deadline_verification_job.rb`, line 7:

```ruby
def perform(assignment_id)
  assignment = Assignment.find(assignment_id)   # ← raises if deleted
  assignment.assignment_submissions
            .where(status: :draft)
            .find_each do |submission|
    CircuitVerificationJob.perform_later(submission.id)
  end
end
```

Compare with the safer pattern in `app/jobs/assignment_deadline_submission_job.rb`, line 7:

```ruby
assignment = Assignment.find_by(id: assignment_id)
return if assignment.nil? || (assignment.status == "closed")
```

## Steps to reproduce

1. Create an assignment with auto-grading enabled (which enqueues `AssignmentDeadlineVerificationJob`).
2. Delete the assignment before the job runs (e.g. by deleting the group).
3. Observe `ActiveRecord::RecordNotFound` in the job queue / Sidekiq failed jobs list.

## Expected behavior

If the assignment no longer exists, the job should exit cleanly with no error.

## Actual behavior

`ActiveRecord::RecordNotFound` is raised, the job fails, and Sidekiq schedules retries, generating noise in the error log and consuming worker capacity.

## Proposed fix

Mirror the pattern already used by `AssignmentDeadlineSubmissionJob`:

```ruby
def perform(assignment_id)
  assignment = Assignment.find_by(id: assignment_id)
  return if assignment.nil?

  assignment.assignment_submissions
            .where(status: :draft)
            .find_each do |submission|
    CircuitVerificationJob.perform_later(submission.id)
  end
end
```

## Impact

Affects any class where an instructor deletes an assignment that has auto-grading scheduled. Failed job retries consume Sidekiq worker threads and inflate error monitoring (Sentry) noise.

magic-peach:
## Summary

- Adds `AssignmentDeadlineVerificationJob` using `find_by` with an early `return` guard, preventing `ActiveRecord::RecordNotFound` when an assignment is deleted while the job is queued

## What changed and why

This job triggers circuit auto-verification for draft submissions at the assignment deadline. Using `Assignment.find(id)` raises `ActiveRecord::RecordNotFound` if the assignment is deleted before the job runs (e.g. when a teacher deletes an assignment that still has pending auto-grade jobs). This causes Sidekiq to log an error and retry the job indefinitely.

The sibling job `AssignmentDeadlineSubmissionJob` already uses the safe pattern:
```ruby
assignment = Assignment.find_by(id: assignment_id)
return if assignment.nil? || (assignment.status == "closed")
```

This PR applies the same pattern:
```ruby
def perform(assignment_id)
  assignment = Assignment.find_by(id: assignment_id)
  return if assignment.nil?

  assignment.assignment_submissions
            .where(status: :draft)
            .find_each do |submission|
    CircuitVerificationJob.perform_later(submission.id)
  end
end
```

## How to test

```bash
# Unit test: enqueue the job for a non-existent assignment_id
AssignmentDeadlineVerificationJob.perform_now(0)
# Should complete without raising RecordNotFound
```

Closes #7184

## Summary by CodeRabbit

* **New Features**
  * Added automated background job to identify and process draft submissions associated with assignments.

### DIFF
--- a/app/jobs/assignment_deadline_verification_job.rb
+++ b/app/jobs/assignment_deadline_verification_job.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class AssignmentDeadlineVerificationJob < ApplicationJob
+  queue_as :verification
+
+  def perform(assignment_id)
+    assignment = Assignment.find_by(id: assignment_id)
+    return if assignment.nil?
+
+    assignment.assignment_submissions
+              .where(status: :draft)
+              .find_each do |submission|
+      CircuitVerificationJob.perform_later(submission.id)
+    end
+  end
+end


magic-peach:
@Asrani-Aman @tachyons @senbo1 @vedant-jain03 could someone please review my pr

PR #7215, opened 2026-03-25
Fix ArgumentError in rack_attack due to invalid email encoding

Gouri-Pawar:
- Problem

There was an ArgumentError ("input string invalid") occurring in rack_attack when calling `.downcase` on the email parameter. This happens when the request contains malformed or non-UTF-8 encoded input (often from bots or unexpected requests).

- Solution
Added a small helper method to safely handle email input before calling `.downcase`. It ensures the string is properly encoded and prevents crashes.

Replaced the direct `.to_s.downcase` calls with this safe method in all relevant places.

- Testing
- Tested with normal email → works as expected
- Tested with invalid encoded input → no crash
- Throttling behavior remains unchanged

- Impact
This prevents the middleware from crashing on bad input and makes the application more robust without changing any existing functionality.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email input handling to prevent encoding-related errors and ensure consistent normalization of email addresses, providing better reliability when processing email addresses with unusual or malformed formats.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,11 +1,19 @@
-# frozen_string_literal: true
-
 class Rack::Attack
+  
+  # it safely handle email input to avoid encoding errors
+  def self.safe_email(value)
+    return nil if value.nil?
+
+    value.to_s
+         .encode("UTF-8", invalid: :replace, undef: :replace, replace: "")
+         .downcase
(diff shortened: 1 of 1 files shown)


naman79820:
no issue attached + that issue must have been closed already.. Thanks for contributing though :))

PR #7246, opened 2026-04-04
Revert "build(deps): bump avo from 3.28.0 to 3.30.3"

tachyons:
Reverts CircuitVerse/CircuitVerse#7173


naman79820:
It was causing some issue? (Just curious)

PR #7397, opened 2026-05-24
fix: patch CGI.parse for Ruby 4.0 compatibility in LTI specs

### LINKED ISSUE #7385: 🐞 Bug: Tests are failing of master branch

### Describe the bug

 **bundle exec rspec**  is failing on master branch

### Steps To Reproduce

1. switch to master branch
2. run bundle exec rspec

### Expected Behavior

Test should pass 

### Screenshots

_No response_

```markdown
- OS:
- Browser:
- version:
```

### Additional context

_No response_

No

aryan7071:
Ruby 4.0 removed CGI.parse which is used by the ims-lti gem. Added a compatibility patch to restore CGI.parse in the test environment so LTI specs pass on CI.

Fixes #7385 

 

The ims-lti gem internally calls CGI.parse which was removed in Ruby 4.0. This caused 4 LTI specs in spec/requests/lti_spec.rb to fail on GitHub Actions CI (which was running Ruby 4.0.2) while passing locally. Added a compatibility patch at the top of the LTI spec file that restores CGI.parse if it does not already exist, fixing the failing tests without changing any application code.

No UI changes — this is a test file fix only.

---

The problem was that CGI.parse was removed in Ruby 4.0, but the ims-lti gem still depends on it internally when calling consumer.generate_launch_data. This only failed in CI because GitHub Actions was picking up Ruby 4.0.2 while local development runs Ruby 3.4.1.
I considered two approaches:

Pinning the Ruby version in the CI workflow to 3.x but this would just delay the problem
Adding a CGI.parse compatibility patch directly in the spec file  this is the better fix as it addresses the root cause

I chose Option 2. The patch uses unless CGI.respond_to?(:parse) to safely define CGI.parse only when it doesn't exist, so it won't break on older Ruby versions. The implementation manually parses query strings by splitting on & and =, then unescaping keys and values  which is exactly what the original CGI.parse did.
I verified the fix by running `bundle exec rspec spec/requests/lti_spec.rb` locally which showed 5 examples, 0 failures ✅

---

---

## Summary by CodeRabbit

* **Chores**
  * Added compatibility support for different Ruby runtime environments to ensure query string parsing works correctly across versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CircuitVerse/CircuitVerse/pull/7397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/config/initializers/cgi_parse_patch.rb
+++ b/config/initializers/cgi_parse_patch.rb
@@ -1,0 +1,15 @@
+# Ruby 4.0 removed CGI.parse which is used internally by the ims-lti gem.
+# This patch restores CGI.parse for compatibility across all environments.
+unless CGI.respond_to?(:parse)
+  def CGI.parse(query_string)
+    params = {}
+    query_string.split("&").each do |pair|
+      next if pair.empty?
+      key, value = pair.split("=", 2)
+      key = CGI.unescape(key.to_s)
+      params[key] ||= []
+      params[key] << CGI.unescape(value) if value
+    end
+    params
+  end
+end
--- a/spec/requests/lti_spec.rb
+++ b/spec/requests/lti_spec.rb
@@ -2,23 +2,6 @@
 
 require "rails_helper"
 
-# Ruby 4.0 removed CGI.parse which is used internally by the ims-lti gem.
-# This patch restores CGI.parse for test compatibility.
-unless CGI.respond_to?(:parse)
-  def CGI.parse(query_string)
-    params = {}
-    query_string.split("&").each do |pair|
-      next if pair.empty?
-      key, value = pair.split("=", 2)
-      key = CGI.unescape(key.to_s)
-      value = CGI.unescape(value.to_s)
-      params[key] ||= []
-      params[key] << value
-    end
-    params
-  end
-end
-
 describe LtiController, type: :request do
   before do
     Flipper.enable(:lms_integration)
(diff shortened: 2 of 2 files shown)


tachyons:
@aryan7071 Could you also ensure functionality also works, not just the specs

@magic-peach Assigning you as the reviewer since you need to update this for GSoC

aryan7071:
@tachyons okay sure!! willl work on that

magic-peach (changes requested):
The patch fixes the tests but the actual LTI controller will still crash on Ruby 4.0 for the same reason  @tachyons  flagged this above. The fix needs to live in `config/initializers/` so it applies in all environments, then the block in the spec can be removed entirely.

Also a small bug: keys without a `=` in the query string end up mapped to `[""]` instead of `[]` because of the `nil.to_s` call. Worth checking that edge case matches what the original `CGI.parse` returned.

aryan7071:
Thank you for the review @magic-peach! I will move the patch to config/initializers/ so it applies in all environments and also fix the edge case with keys without = sign. Will update the PR shortly!

aryan7071:
Hi @magic-peach ,   I have moved the CGI.parse compatibility patch from the spec file to config/initializers/cgi_parse_patch.rb so it now applies in all environments including production, not just tests. I have also fixed the edge case where keys without = in the query string were incorrectly mapped to [""] instead of [] by adding a nil check before appending the value. The spec file has been cleaned up and the LTI tests are still passing with 5 examples, 0 failures ✅. Please review the updated changes!

PR #7457, opened 2026-06-03
feat: Add Organizations and OrganizationMembers controllers

### LINKED ISSUE #7455: Feature: Add OrganizationsController and OrganizationMembersController

With the database schema and model layer in place, there are no controllers to handle HTTP requests for organizations and their members. Without controllers, we can't expose any endpoints for creating, viewing, updating, or managing organizations and memberships.

Add two controllers to handle organization-related actions:
1. `OrganizationsController`
2. `OrganizationMembersController`

Also add the corresponding routes in `config/routes.rb` with `organizations`.

No alternatives

### Additional context

 This is part of the GSoC 2026 "Enterprise & Institutional Organization Features" project. 

### Are you working on this?

Yes

naman79820:
Fixes #7455 

Added the basic CRUD controllers for the new Organizations feature to allow users to create and manage organizations and their members. Specifically:

1. OrganizationsController: Handles creating, viewing, updating, and deleting orgs.
2. OrganizationMembersController: Handles adding users to an org, updating their roles, removing them, and allowing users to voluntarily leave an organization via a dedicated leave endpoint.
3. Routing: Added the corresponding nested routes using Rails Shallow Nesting to cleanly allow organizations/:id/groups creation without deep-routing assignments or edit pages.
4. Feature Flags: Gated the controllers behind an :organizations Flipper feature flag scoped to current_user so this can be safely beta-tested in production.
5. Added translation keys in a dedicated locale file config/locales/views/organizations/en.yml
6. Permitted :remove_logo in the params so admin can delete their organization's avatar, and implemented pagination on the index action.
7. Robust Error Handling: Refactored the create transaction to rescue ActiveRecord::RecordInvalid exceptions. If the initial admin assignment fails, the app safely rolls back the transaction and renders the validation error on the form (with a specific t(".failure") flash message) instead of throwing a 500 Server Error.
8. Explore Functionality: Updated the index action to support an ?explore=true query parameter, allowing users to view a paginated list of all public organizations on the platform.
9. Groups Controller Integration & Security:

-  Dynamically authorized users against OrganizationPolicy#create_group? and OrganizationGroupPolicy#manage?.
- Prevented Flipper bypasses by blocking hidden payload injections for organization_id.
- Prevented organization hijacking by locking organization_id out of permitted params during updates.

10. Comprehensive Testing: Added 31 RSpec controller tests to verify CRUD logic, feature flags, unauthenticated redirects, Pundit HTTP 403 Forbidden catches, and specific edge cases (e.g., preventing the sole admin of an organization from demoting or removing themselves).

I decided to follow the exact same pattern used in GroupsController and GroupMembersController to keep things consistent across the codebase. I split the responsibilities into two separate controllers rather than handling member logic inside the main org controller. For member updates, I specifically created organization_member_update_params to only allow role changes, preventing someone from tampering with the user_id on an existing membership.

To ensure safe deployment, I added a before_action Flipper check matching how Contests were originally rolled out, so the feature remains hidden in production until enabled. I scoped this to current_user to allow for selective beta testing.

Finally, I dynamically wired both OrganizationMembersController and GroupsController to enforce our Pundit authorization layer. Because Pundit raises an exception rather than returning false, our tests specifically catch these Pundit::NotAuthorizedError exceptions and verify the ApplicationController correctly rescues them as a HTTP 403 Forbidden status. This securely handles edge cases like ensuring an organization's sole admin cannot accidentally lock themselves out!

---

---

## Summary by CodeRabbit

* **New Features**
  * Organizations management: create, view, edit, and delete organizations via web and API (includes pagination and an "explore" view for public organizations). Access requires authentication and appropriate admin permissions.
  * Organization membership management: add members, change roles, and remove members via web and API with success and validation feedback.

* **Documentation**
  * Added localized messages for organization and membership actions (success and failure).

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -57,7 +57,7 @@ class GroupsController < ApplicationController
   def create
     @group = current_user.groups_owned.new(group_params)
 
-    authorize @group.organization, :create_group? if @group.organization_id.present?
+    authorize @group.organization, :create_group? if @group.organization
 
     respond_to do |format|
       if @group.save
@@ -105,11 +105,9 @@ class GroupsController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      if action_name == "create"
-        params.expect(group: %i[name primary_mentor_id organization_id])
-      else
-        params.expect(group: %i[name primary_mentor_id])
-      end
+      permitted = %i[name primary_mentor_id]
+      permitted << :organization_id if action_name == "create" && Flipper.enabled?(:organizations, current_user)
+      params.expect(group: permitted)
     end
 
     def check_show_access
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -3,6 +3,7 @@
 class GroupsController < ApplicationController
   before_action :set_group, only: %i[show edit update destroy group_invite generate_token]
   before_action :authenticate_user!
+  before_action :check_organizations_feature_flag, only: %i[new create], if: -> { params[:organization_id].present? }
   before_action :check_show_access, only: %i[show edit update destroy]
   before_action :check_edit_access, only: %i[edit update destroy generate_token]
 
@@ -54,6 +55,8 @@ class GroupsController < ApplicationController
   def create
     @group = current_user.groups_owned.new(group_params)
 
+    authorize @group.organization, :create_group? if @group.organization_id.present?
+
     respond_to do |format|
       if @group.save
         format.html { redirect_to @group, notice: "Group was successfully created." }
@@ -100,7 +103,7 @@ class GroupsController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      params.expect(group: %i[name primary_mentor_id])
+      params.expect(group: %i[name primary_mentor_id organization_id])
     end
 
     def check_show_access
@@ -108,6 +111,14 @@ class GroupsController < ApplicationController
     end
 
     def check_edit_access
-      authorize @group, :admin_access?
+      if @group.organization_id.present?
+        authorize @group, :manage?, policy_class: OrganizationGroupPolicy
+      else
+        authorize @group, :admin_access?
+      end
+    end
+
+    def check_organizations_feature_flag
+      redirect_to root_path, alert: "Feature not available" unless Flipper.enabled?(:organizations)
     end
 end
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -3,7 +3,9 @@
 class GroupsController < ApplicationController
   before_action :set_group, only: %i[show edit update destroy group_invite generate_token]
   before_action :authenticate_user!
-  before_action :check_organizations_feature_flag, only: %i[new create], if: -> { params[:organization_id].present? }
+  before_action :check_organizations_feature_flag, only: %i[new create], if: lambda {
+    params[:organization_id].present? || params.dig(:group, :organization_id).present?
+  }
   before_action :check_show_access, only: %i[show edit update destroy]
   before_action :check_edit_access, only: %i[edit update destroy generate_token]
 
@@ -103,7 +105,11 @@ class GroupsController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      params.expect(group: %i[name primary_mentor_id organization_id])
+      if action_name == "create"
+        params.expect(group: %i[name primary_mentor_id organization_id])
+      else
+        params.expect(group: %i[name primary_mentor_id])
+      end
     end
 
     def check_show_access
@@ -119,6 +125,6 @@ class GroupsController < ApplicationController
     end
 
     def check_organizations_feature_flag
-      redirect_to root_path, alert: "Feature not available" unless Flipper.enabled?(:organizations)
+      redirect_to root_path, alert: t("feature_not_available") unless Flipper.enabled?(:organizations, current_user)
     end
 end
--- a/app/controllers/organization_members_controller.rb
+++ b/app/controllers/organization_members_controller.rb
@@ -56,7 +56,13 @@ class OrganizationMembersController < ApplicationController
   # DELETE /organizations/1/leave
   # DELETE /organizations/1/leave.json
   def leave
-    @organization_member = @organization.organization_members.find_by!(user: current_user)
+    @organization_member = @organization.organization_members.find_by(user: current_user)
+
+    if @organization_member.nil?
+      redirect_to organizations_path, alert: t(".not_a_member")
+      return
+    end
+
     authorize @organization, :leave?
 
     @organization_member.destroy
--- a/app/controllers/organization_members_controller.rb
+++ b/app/controllers/organization_members_controller.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+class OrganizationMembersController < ApplicationController
+  before_action :authenticate_user!
+  before_action :check_organizations_feature_flag
+  before_action :set_organization
+  before_action :set_organization_member, only: %i[update destroy]
+  before_action :check_create_access, only: %i[create]
+  before_action :check_access, only: %i[update destroy]
+
+  # POST /organizations/1/organization_members
+  # POST /organizations/1/organization_members.json
+  def create
+    @organization_member = @organization.organization_members.new(organization_member_params)
+
+    respond_to do |format|
+      if @organization_member.save
+        format.html { redirect_to @organization, notice: t(".success") }
+        format.json { render :show, status: :created, location: @organization }
+      else
+        format.html do
+          redirect_to @organization, alert: @organization_member.errors.full_messages.to_sentence
+        end
+        format.json { render json: @organization_member.errors, status: :unprocessable_content }
+      end
+    end
+  end
+
+  # PATCH/PUT /organizations/1/organization_members/1
+  # PATCH/PUT /organizations/1/organization_members/1.json
+  def update
+    respond_to do |format|
+      if @organization_member.update(organization_member_update_params)
+        format.html { redirect_to @organization, notice: t(".success") }
+        format.json { head :no_content }
+      else
+        format.html do
+          redirect_to @organization, alert: @organization_member.errors.full_messages.to_sentence
+        end
+        format.json { render json: @organization_member.errors, status: :unprocessable_content }
+      end
+    end
+  end
+
+  # DELETE /organizations/1/organization_members/1
+  # DELETE /organizations/1/organization_members/1.json
+  def destroy
+    @organization_member.destroy
+    respond_to do |format|
+      format.html { redirect_to @organization, notice: t(".success") }
+      format.json { head :no_content }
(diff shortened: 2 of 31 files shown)


app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      redirect_to root_path, alert: t("feature_not_available")
+    end
+
+    def check_show_access
+      authorize @organization, :show_access?
+    end
+
+    def check_edit_access
+      authorize @organization, :admin_access?
+    end
```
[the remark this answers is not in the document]
naman79820: will create a seprate pr for pundit policy (also mentioned in the description)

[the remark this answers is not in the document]
naman79820: Although i have flipper so it won't break anything in production, still i will consider your recommendation and i will try to get policy pr merged before controllers ok?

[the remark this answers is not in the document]
naman79820: nah it's fine, i'll do it myself thankss :))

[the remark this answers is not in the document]
naman79820: can you please make it resolve as of now then?

naman79820:
Currently it consists only CRUD opertaions, need to discuss something in today's meet :))

app/views/organizations/edit.html.erb
naman79820, line 1: Had to add empty files for now.. else tests were failing locally

config/routes.rb
```diff
@@ -26,6 +26,11 @@
   end
 
   # resources :assignment_submissions
+  resources :organizations do
+    resources :organization_members, only: %i[create update destroy]
+    delete :leave, to: "organization_members#leave"
+    resources :groups, only: %i[index new create]
```
naman79820: instead of nesting everything in (groups assignment). I have ussed shallow routing, now everything is good to go we can just access groups inside org now `/organizations/:organization_id/groups` also assignment is already nested inside groups so no need to worry about it.

naman79820:
sorry for bad commits earlier...

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+      else
+        format.html { render :edit }
+        format.json { render json: @organization.errors, status: :unprocessable_content }
+      end
+    end
+  end
+
+  # DELETE /organizations/1
+  # DELETE /organizations/1.json
+  def destroy
```
PRATHAM2002-DS, line 63: The delete for org is implemented without pre checks. Like what happens to all groups, group members, assignments

app/controllers/organization_members_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+
+      redirect_to root_path, alert: t("feature_not_available")
+    end
+
+    def check_create_access
+      authorize @organization, :admin_access?
+    end
+
+    def check_access
+      authorize @organization_member, :"#{action_name}?"
```
PRATHAM2002-DS: shouldnt we have function with explicit action as whats above? Check whats the pattern used, it can cause Type errors in policy evaluations

app/controllers/organization_members_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    respond_to do |format|
+      format.html { redirect_to organizations_path, notice: t(".success") }
+      format.json { head :no_content }
+    end
+  end
+
+  private
+
+    def set_organization
+      @organization = Organization.find(params.expect(:organization_id))
```
PRATHAM2002-DS: If we want slug management for org_members or link based access, then we should use friendly.find here as well and make it consistent with organisation

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+  before_action :authenticate_user!
+  before_action :check_organizations_feature_flag
+  before_action :set_organization, only: %i[show edit update destroy]
+  before_action :check_show_access, only: %i[show]
+  before_action :check_edit_access, only: %i[edit update destroy]
+
+  # GET /organizations
+  def index
+    @organizations = if params[:explore].present?
+      Organization.where(private: false).paginate(page: params[:page])
```
PRATHAM2002-DS: this has no ordering, I think we should have ordering here to avoid misses , .order()

app/controllers/organizations_controller.rb
naman79820: It becomes standalone group for now. I discussed this in the last meeting, and Vedant bhaiya said we'd discuss later whether we should remove groups entirely or keep them. Currently, I'm thinking of adding an option when deleting an organization that lets users choose whether they also want to delete all its groups or keep them.

app/controllers/organization_members_controller.rb
naman79820: just added friendly in it :))

app/controllers/organizations_controller.rb
naman79820: Added ordering inside it. Now new org will be shown at the top :))

app/controllers/organization_members_controller.rb
naman79820: done

app/controllers/groups_controller.rb
```diff
@@ -54,6 +57,8 @@ def edit; end
   def create
     @group = current_user.groups_owned.new(group_params)
 
+    authorize @group.organization, :create_group? if @group.organization_id.present?
```
vedant-jain03: **Thought** Wondering if we need this right now since on group creation there is no addition of org. Or wait does this refer to creating group inside from organization?

vedant-jain03: We might need to check in params then if it's under organization page

app/controllers/groups_controller.rb
```diff
@@ -100,14 +105,26 @@ def set_group
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      params.expect(group: %i[name primary_mentor_id])
+      if action_name == "create"
+        params.expect(group: %i[name primary_mentor_id organization_id])
+      else
+        params.expect(group: %i[name primary_mentor_id])
```
vedant-jain03: **Suggestion:** I think we can except org id for show as well since we might want to render if group belongs to any org in groups page?

app/controllers/groups_controller.rb
```diff
@@ -100,14 +105,26 @@ def set_group
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      params.expect(group: %i[name primary_mentor_id])
+      if action_name == "create"
+        params.expect(group: %i[name primary_mentor_id organization_id])
```
vedant-jain03: check if expect is hard or soft check

app/controllers/groups_controller.rb
```diff
@@ -100,14 +105,26 @@ def set_group
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def group_params
-      params.expect(group: %i[name primary_mentor_id])
+      if action_name == "create"
```
vedant-jain03:
```suggestion
      if check_organizations_feature_flag
```

app/controllers/groups_controller.rb
naman79820: Just checked and expect is hard check for groups and soft for its attributes

app/controllers/groups_controller.rb
naman79820: yeah this is for creating a group from inside an org. i check @group.organization (not just organization_id.present?) so it only runs the org auth when a real org is attached, if someone passes a junk org id, .organization is nil, we skip authorize, and the save just fails validation instead of blowing up.

app/controllers/groups_controller.rb
naman79820: checked group_params only runs for create/update, show never calls it, so we can render the org by just reading @group.organization in the view. no need to permit it here. (keeping it create-only, for the reason in the group_params thread.) :))

app/views/organizations/show.html.erb
yashikajotwani12, line 1: @naman79820 are these empty files needed

naman79820: Yes i had to, else tests will fail then.... :)) @yashikajotwani12

app/controllers/organization_members_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+    respond_to do |format|
+      format.html { redirect_to @organization, notice: t(".success") }
+      format.json { head :no_content }
+    end
+  end
+
+  # DELETE /organizations/1/leave
+  # DELETE /organizations/1/leave.json
+  def leave
+    @organization_member = @organization.organization_members.find_by!(user: current_user)
```
yashikajotwani12: inline `find_by!` here breaks this controller's before_action pattern, and will raise an unhandled RecordNotFound if the user isn't a member — swap to `find_by` + a flash alert redirect like the other actions use.

app/views/organizations/show.html.erb
yashikajotwani12: which specs are failing?

app/controllers/organization_members_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+  before_action :set_organization
+  before_action :set_organization_member, only: %i[update destroy]
+  before_action :check_create_access, only: %i[create]
+  before_action :check_update_access, only: %i[update]
+  before_action :check_destroy_access, only: %i[destroy]
+
+  # POST /organizations/1/organization_members
+  # POST /organizations/1/organization_members.json
+  def create
+    @organization_member = @organization.organization_members.new(organization_member_params)
```
yashikajotwani12, line 15: Is there a rule in place preventing the same user from being added to an organization more than once — i.e. a uniqueness check on the user+organization pairing?

app/views/organizations/show.html.erb
naman79820: Inside organization controller specs  GET actions will fail as they currently render views templates. 
Also at the end we need those files anyway for the UI. PR -> (https://github.com/CircuitVerse/CircuitVerse/pull/7568)

app/controllers/organization_members_controller.rb
naman79820: Yes yes there's a uniqueness rule on the user + organization pairing. the model has `validates :user_id, uniqueness: { scope: :organization_id }`, and there's a matching `unique DB index on [organization_id, user_id]` too, so the same user can only have one membership per org (and therefore one role).

app/controllers/organization_members_controller.rb
naman79820: okee i'll update it :))

naman79820: Done, leave now uses find_by with a flash redirect for non-members

naman79820:
Once again, thank you everyone for giving your precious time for me. Means a lot 🙏 ❤️ 
@vedant-jain03 @yashikajotwani12 @PRATHAM2002-DS

PRATHAM2002-DS (approved):

PRATHAM2002-DS:
Fix failing CI @naman79820

naman79820:
> Fix failing CI @naman79820

Already resolved in https://github.com/CircuitVerse/CircuitVerse/pull/7606
Just had to rebase :))

vedant-jain03 (approved):
LGTM

PR #7504, opened 2026-06-10
fix(policy): use safe-navigation operator in ProjectPolicy#author_access?

KeerthiKumarR:
## Summary

Fixes an operator-precedence / nil-edge bug in `ProjectPolicy#author_access?`
that gates six destructive or privileged actions across both the web and API
controllers.

## The Bug

```ruby
# BEFORE (broken)
def author_access?
  (user.present? && user.admin?) || project.author_id == (user.present? && user.id)
end
```

In Ruby, `(user.present? && user.id)` short-circuits to `user.id` (an
Integer) when the user is present. This looks correct, but when `user.id`
is `nil` — which can happen with unsaved records, certain test doubles, or
edge-case OAuth flows before a user is fully persisted — the expression
evaluates as:

```
(true && nil)  →  nil
project.author_id == nil  →  true   ← for any project with a NULL author_id
```

This bypasses the ownership check entirely for those records.

Beyond the nil edge, `(user.present? && user.id)` is non-idiomatic Ruby and
makes the intent harder to read than necessary.

## The Fix

```ruby
# AFTER (correct)
def author_access?
  (user.present? && user.admin?) || project.author_id == user&.id
end
```

`user&.id` explicitly returns `nil` when `user` is nil, and `user.id` when
the user is present. It is consistent with how nil-guards are written
throughout this file and across the rest of the codebase.

## Diff

```diff
   def author_access?
-    (user.present? && user.admin?) || project.author_id == (user.present? && user.id)
+    (user.present? && user.admin?) || project.author_id == user&.id
   end
```

## Blast Radius

`author_access?` is the authorization gate for these six call sites:

| File | Line | Action guarded |
|------|------|----------------|
| `projects_controller.rb` | 136 | DELETE project (web) |
| `api/v1/projects_controller.rb` | 140 | DELETE project (API) |
| `collaborations_controller.rb` | 18 | POST collaboration |
| `collaborations_controller.rb` | 47 | PATCH collaboration |
| `collaborations_controller.rb` | 65 | DELETE collaboration |
| `api/v1/collaborators_controller.rb` | 49 | DELETE collaborator (API) |

## Tests

The existing spec in `spec/policies/project_policy_spec.rb` already asserts:

```ruby
context "for a visitor" do
  it { is_expected.not_to permit(:author_access) }
end
```

This test remains green. A tighter context has also been added to explicitly
document the non-author case.

## Checklist

- [x] Single-responsibility change (one method, one expression)
- [x] No new dependencies
- [x] Existing spec coverage confirmed
- [x] Commit follows conventional commits format (`fix(scope): description`)
- [x] PR title follows `fix(scope): description` format
- [x] No migration needed (policy-only change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project author-access checks to ensure only administrators and the project’s author receive author-level access.
  * Prevented visitors and other non-authors from receiving author permissions on public projects.

* **Tests**
  * Expanded authorization coverage for public projects and mismatched author identities.

* **Chores**
  * Improved development and testing workflows.


KeerthiKumarR:
ready for review @tachyons please go through it.

KeerthiKumarR:
@tachyons  please go through it.

KeerthiKumarR:
@vedant-jain03 PTAL when you got time!

PR #7510, opened 2026-06-12
Added the lti-advantage 0.1.0 gem

### LINKED ISSUE #7509: Updating the gemfile with the lti-advantage 0.1.0 gem
The lti-advantage 0.1.0 gem is used in canvas lms and needs to be added in Circuitverse as well in order to integrate the LTI 1.3 version

magic-peach:
Fixes #7509 

 

Added the lti-advantage gem 0.1.0 as a part of upgrading to LTI 1.3 and also ran tests

Testing:
Screenshot 2026-06-12 at 11 54 25 PM

Screenshot 2026-06-12 at 11 55 11 PM

Screenshot 2026-06-12 at 11 55 43 PM

`lti_advantage` gem was not widely documented moeover is not used in Canvas LMS, since our main goal is to integrate LTI with an LMS I decided to add the gem used in Canvas LMS which is `lti-advantage 0.1.0` . After adding it to the gemfile  I ran `bundle install `and then ran tests

---

---

## Summary by CodeRabbit

* **New Features**
  * Added LTI 1.3 (OIDC) support for launches, including dedicated endpoints for OIDC initiation, JWKS key discovery, and tool configuration.
  * Launch requests now route automatically to the appropriate LTI flow (LTI 1.3 when an `id_token` is provided, otherwise LTI 1.1).
* **Security**
  * Enforced JWT and state/nonce checks for LTI 1.3 launches, with clear JSON errors for invalid/unknown deployments.
* **Tests**
  * Updated LTI request coverage to validate the new LTI 1.3 endpoints and launch outcomes.
* **Chores**
  * Updated LTI-related dependencies to use the Canvas LMS–provided `lti-advantage` gem and added `json-jwt`.


Gemfile
```diff
@@ -77,7 +77,12 @@ gem "ahoy_matey"
 gem "i18n-js"
 
 # for lti provider
-gem "ims-lti", "~> 1.2", "< 2.0"
```
Asrani-Aman, line 80: whhy you are removing this gem? 
i guess because of this only tests are failing

Gemfile
```diff
@@ -77,7 +77,12 @@ gem "ahoy_matey"
...
+gem "lti-advantage",
+    git:  "https://github.com/instructure/canvas-lms",
```
Asrani-Aman: @magic-peach are we clonning comlete canvas-lms?

Asrani-Aman (changes requested):

Gemfile
magic-peach: ims-lti only supports 1.1 so I removed it and for the tests failing I thought of addressing them in the next pr since this one was only meant to update the gem

Gemfile
magic-peach: no, we are only targetting the lti-advantage gem in the gemspec. Actually I saw canvas has lti-advantage as their own gem and couldn't find it anywhere on the web

tachyons: This is tricky with git partial clone. Do we have any alternative options, this seems to be their internal gem

Gemfile
tachyons: Iterations shouldn't break CI as it will block next other people working on the same code base

app/controllers/lti_controller.rb
```diff
@@ -1,82 +1,180 @@
...
+
+    def handle_lti_11_launch
+      session[:is_lti] = true
+      if @assignment.blank?
+        flash.now[:notice] = t(".notice_no_assignment")
         render :launch_error, status: :unauthorized
+        return
+      end
+      require "oauth/request_proxy/action_controller_request"
+      @provider = IMS::LTI::ToolProvider.new(
```
anxkhn, line 164: Since `ims-lti` was removed from the bundle, this legacy launch path can no longer load `IMS::LTI::ToolProvider` or the oauth request proxy. A valid LTI 1.1 launch that reaches this branch will raise `LoadError` or `NameError` instead of keeping the existing launch behavior. If LTI 1.1 is still supported, we need to keep `ims-lti`; otherwise this branch should be removed or replaced.

---

**Update (re-checked against `4860a6f`):** Resolved. `gem "ims-lti", "~> 1.2", "< 2.0"` is back in the `Gemfile`/`Gemfile.lock`, so `handle_lti_11_launch` loads `IMS::LTI::ToolProvider` again.

One follow-up: the PR title ("Added the lti-advantage 0.1.0 gem") is now inaccurate. `lti-advantage` is not in `Gemfile.lock`; the only net-new runtime gem is `json-jwt`, and the LTI 1.3 path actually runs on `ruby-jwt` (3.2.0) + `json-jwt`. Worth correcting the title/description so it matches what shipped.

app/controllers/lti_controller.rb
```diff
@@ -1,82 +1,180 @@
...
+      target_link_uri: "#{request.base_url}/lti/launch",
+      oidc_initiation_url: "#{request.base_url}/lti/login",
+      public_jwk_url: "#{request.base_url}/lti/jwks"
+    }
+  end
+
+  private
+
+    def handle_lti_13_launch
+      if session[:lti_state].present? && params[:state] != session[:lti_state]
```
anxkhn: This only rejects a bad state when `session[:lti_state]` exists. If the browser loses the OIDC session, or a launch is posted directly, the state check is skipped and `nonce` is blank so the validator skips nonce validation too. LTI 1.3 launch handling should require both stored values from `/lti/login` and reject the launch when they are missing or mismatched.

---

**Update (re-checked against `4860a6f`):** Resolved. `handle_lti_13_launch` now rejects with 401 whenever the stored state is missing or mismatched (`session[:lti_state].blank? || params[:state] != session[:lti_state]`), and `Lti::JwtValidator.validate!` raises `SecurityError, "Missing nonce"` when the passed nonce is blank. An unbound or replayed launch (no prior `/lti/login` in the session) can no longer skip nonce validation. The new specs cover both the "no session state" and "wrong state" cases.

app/controllers/lti_controller.rb
```diff
@@ -1,82 +1,180 @@
...
-    else # no such user in circuitverse
-      flash[:notice] = t(".notice_no_account_in_cv", email_from_lms: @email_from_lms)
-      render :launch_error, status: :bad_request
     end
-  end
 
-  private
+    def find_lti_13_deployment(token)
+      payload, _header = JWT.decode(token, nil, false)
+      LtiDeployment.find_by!(
```
anxkhn, line 203: This lookup ignores the LTI deployment claim, so a signed token for any deployment under the same issuer and client can be accepted as whichever `LtiDeployment` happens to match first. That matters when a platform has multiple deployments. The lookup and validator should include `https://purl.imsglobal.org/spec/lti/claim/deployment_id` and reject tokens whose deployment id does not match the stored deployment.

---

**Update (re-checked against `4860a6f`):** Resolved. `find_lti_13_deployment` now scopes `LtiDeployment.find_by!` by `issuer`, `client_id` (first `aud`), and `deployment_id` (`DEPLOYMENT_ID_CLAIM`), so a token carrying a deployment id that does not match a registered deployment returns 404 instead of binding to an arbitrary row. The "unregistered deployment_id" spec covers this.

app/services/lti/jwt_validator.rb
```diff
@@ -0,0 +1,52 @@
...
+          true,
+          algorithms: ["RS256"],
+          iss: deployment.issuer,
+          aud: deployment.client_id,
+          verify_iss: true,
+          verify_aud: true
+        )
+
+        raise SecurityError, "Nonce mismatch" if nonce.present? && payload["nonce"] != nonce
+        raise ArgumentError, "Missing required claims" \
```
anxkhn: `ArgumentError` is not rescued by the launch action, so a signed token that is missing one of these claims will return a 500 instead of an unauthorized response. This should raise an error type the controller rescues, or the controller should rescue this validation error and return 401 like the other invalid-token paths.

---

**Update (re-checked against `4860a6f`):** Resolved. The missing-claims branch now raises `JWT::DecodeError, "Missing required claims"`, which the launch action already rescues to 401. Confirmed against `ruby-jwt` 3.2.0 that the other invalid-token paths (`InvalidIssuerError`, `InvalidAudError`, `ExpiredSignature`, `VerificationError`) are all subclasses of `JWT::DecodeError`, so iss/aud/exp/signature failures return 401 consistently as well. The "missing required claim" spec asserts 401 rather than 500.

app/controllers/lti_controller.rb
```diff
@@ -1,82 +1,183 @@
...
+        issuer: payload["iss"],
+        client_id: Array(payload["aud"]).first,
+        deployment_id: payload[DEPLOYMENT_ID_CLAIM]
+      )
+    end
+
+    def find_or_create_user_from_lti13(payload)
+      User.find_or_create_by(email: payload["email"]) do |u|
```
anxkhn: `find_or_create_user_from_lti13` keys both identity and provisioning on `payload["email"]`, but `email` is not in `Lti::JwtValidator::REQUIRED_CLAIMS` (`sub iss aud nonce`) and is optional in LTI 1.3, the platform only sends it when it chooses to share it (or the right scope is granted). Two issues follow:

- **Launch without an `email` claim does not return a clean 401.** `User.find_or_create_by(email: nil)` builds a record that fails `validates :email, presence: true`. The non-bang `find_or_create_by` does not raise, so the invalid, unpersisted `User` is passed straight to `sign_in` and `redirect_to root_path`. With `:trackable` enabled, persisting it (`save(validate: false)`) hits the `NOT NULL` `users.email` column, so a spec-conformant launch that omits `email` breaks (server error) instead of being rejected.
- **`email` is unverified, so matching by it is spoofable.** When the claim is present, `find_or_create_by(email: ...)` can match an existing CircuitVerse account, letting a deployment that asserts someone else's address sign in as that user.

Identity should key on the validated `sub` claim scoped to the deployment, and `email` should either be added to `REQUIRED_CLAIMS` (reject blank with 401) or only used after it is verified.

For reference, CodeRabbit's earlier email comment is marked "Addressed in commits f9679a8 to 6aea8a6", but the current code still keys on `email`, so that one is not actually resolved.

app/controllers/lti_controller.rb
```diff
@@ -1,82 +1,230 @@
...
+
+    # Identify the user by the validated, deployment-scoped LTI subject (sub)
+    # rather than the self-asserted email claim, which is unverified and
+    # spoofable. Email is only used when provisioning a new account.
+    def find_or_create_user_from_lti13(payload, deployment)
+      User.find_or_create_by!(provider: "lti", uid: "#{deployment.id}:#{payload['sub']}") do |u|
+        u.email        = payload["email"]
+        u.name         = payload["name"].presence || payload["email"]
+        u.password     = SecureRandom.hex(16)
+        u.confirmed_at = Time.zone.now
```
tachyons, line 218: This allows people to create any number of confirmed users and can be easily abused

config/environments/development.rb
```diff
@@ -122,4 +122,11 @@
   config.hosts.clear
   config.hosts << /.*\Z/
 
+  # LTI 1.3 launches return via a cross-site POST from the platform (Canvas)
+  # to /lti/launch, so the session cookie carrying the nonce/state must be
+  # SameSite=None; Secure to survive. Requires HTTPS (served via ngrok).
+  config.action_dispatch.cookies_same_site_protection = :none
+  config.session_store :cookie_store, key: "_circuitverse_session",
+                                      same_site: :none, secure: true
+
```
tachyons: Do we need to make same changes for other ends ?

app/controllers/lti_controller.rb
magic-peach: Addressed in fb95f37f. LTI 1.3 — including this auto-provisioning path — is now gated behind a Flipper feature flag (`lti_advantage`) that **defaults to off**, so an operator has to explicitly opt in before any account can be created. On top of that, provisioning is bounded to platforms an admin has explicitly registered as an `LtiDeployment`, and the account is keyed on the signature-validated `sub` from the id_token (not the spoofable email claim). The OIDC endpoints and the 1.3 launch return 404 while the flag is off, with specs covering both states.

PR #7571, opened 2026-06-29
test: add unit tests for EventQueue simulator engine

### LINKED ISSUE #7570: Add unit tests for EventQueue (simulator engine)
### Description

The `EventQueue` class (`simulator/src/eventQueue.js`) — the priority queue that drives the simulation engine's event scheduling — currently has no dedicated unit tests. Its behavior (time-ordered insertion, re-queuing of already-queued elements, `pop` ordering, size limits, and `reset`) is only exercised indirectly through full-circuit simulation specs.

### Proposed solution

Add `simulator/spec/eventQueue.spec.js` covering:

- constructor / `isEmpty`
- `add` (default `propagationDelay` vs explicit delay; time-ordered `pop`)
- re-`add` of an already-queued element (re-time + re-order)
- `addImmediate`
- size-exceeded and empty-queue error paths
- `reset`

This adds direct coverage for a core engine data structure without changing any runtime behavior.

I'd like to work on this.

eeshsaxena:
Fixes #7570

### Description

Adds `simulator/spec/eventQueue.spec.js`, providing direct unit-test coverage for the `EventQueue` class (`simulator/src/eventQueue.js`) — the priority queue that schedules simulation events. Previously it was only exercised indirectly via full-circuit specs.

Coverage includes:

- constructor state and `isEmpty`
- `add` using the element's `propagationDelay` and an explicit delay
- time-ordered `pop` (and the simulation clock advancing to the popped element's time)
- re-`add` of an already-queued element (re-timing and re-ordering)
- `addImmediate`
- the `EventQueue size exceeded` and `Queue Empty` error paths
- `reset` clearing both queue and per-element state

No runtime behavior is changed — this is test-only.

### Type of change

- [x] Adding test cases

### How has this been tested?

Ran the new spec with Jest (the project's existing test runner) against `simulator/src/eventQueue.js`; all 9 tests pass.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for event queue behavior, including adding items, popping in time order, immediate scheduling, re-queuing, capacity limits, empty-queue errors, and reset behavior.

### DIFF
--- a/simulator/spec/eventQueue.spec.js
+++ b/simulator/spec/eventQueue.spec.js
@@ -1,0 +1,114 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import EventQueue from '../src/eventQueue';
+
+/**
+ * Build a minimal stand-in for a CircuitElement, exposing only the
+ * `propagationDelay` and `queueProperties` fields that EventQueue touches.
+ */
+function makeElement(propagationDelay = 0) {
+    return {
+        propagationDelay,
+        queueProperties: { inQueue: false, time: 0, index: undefined },
+    };
+}
+
+describe('EventQueue', () => {
+    test('constructor initializes an empty queue', () => {
+        const queue = new EventQueue(10);
+        expect(queue.size).toBe(10);
+        expect(queue.frontIndex).toBe(0);
+        expect(queue.time).toBe(0);
+        expect(queue.isEmpty()).toBe(true);
+    });
+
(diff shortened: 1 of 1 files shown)


eeshsaxena:
Good catch, thanks! Added a dedicated case asserting that `add(element, 0)` falls back to the element's `propagationDelay` (because `add` uses `delay || obj.propagationDelay`), so the zero-delay edge is now locked in.

PR #7618, opened 2026-07-06
fix: block removing an org member who is a group's primary mentor (#7615)

### LINKED ISSUE #7615: bug: destroy? allows primary mentor to remove own org membership
### Description

OrganizationMemberPolicy#destroy? does not check whether the user leaving is a primary mentor of any org group, unlike OrganizationPolicy#leave? which explicitly blocks this case.

### Steps to reproduce

1. Open `app/policies/organization_member_policy.rb` around line 15
2. A user who is set as primary_mentor_id on one or more organization groups can call the destroy endpoint to remove their own OrganizationMember record, leaving those groups without a primary mentor.

### Expected behavior

destroy? only guards against `leaving_self? && sole_admin?`. OrganizationPolicy#leave? has an additional guard `return false if record.groups.exists?(primary_mentor_id: user.id)` that is entirely absent from destroy?. should be handled correctly.

### Actual behavior

A user who is set as primary_mentor_id on one or more organization groups can call the destroy endpoint to remove their own OrganizationMember record, leaving those groups without a primary mentor.

deepak0x:
Fixes #7615

`OrganizationMemberPolicy#destroy?` was missing the primary-mentor guard that its sibling `OrganizationPolicy#leave?` already has. As a result a user who is the `primary_mentor` of one or more organization groups could remove their own `OrganizationMember` record through the destroy endpoint and orphan those groups. Since `Group` requires a `primary_mentor` (`belongs_to :primary_mentor`), that leaves the group in an invalid state.

This adds a guard to `destroy?` that denies removing any membership whose user is still the primary mentor of a group in the organization:

- In this policy `record` is the membership (not the organization, as in `leave?`), so the groups are reached via `record.organization.groups`.
- The check is keyed on the member being removed (`record.user_id`), so it also covers an org/site admin removing a mentor's membership — because the group is orphaned regardless of who triggers the removal. This is intentionally broader than `leave?`, which only concerns the user leaving themselves.

N/A — this is an authorization-policy change with no UI.

---

- **Problem:** `destroy?` allowed a primary mentor's membership to be deleted, orphaning the group(s) they mentor.
- **Approach:** mirror the existing `leave?` guard, but key it on the membership being removed (`record.user_id`) instead of only the current user, so admin-initiated removals are covered too.
- **Alternative considered:** gating the check on `leaving_self?` only (blocking just self-removal). Rejected because an org/site admin could still orphan a group by removing a mentor.
- **Key method:** `member_primary_mentor_of_any_group?` → `record.organization.groups.exists?(primary_mentor_id: record.user_id)`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened member deletion rules so a person who is the primary mentor for any group can’t have their membership removed.
  * Prevented admins from removing the membership of a primary mentor.
  * Confirmed that members who do not mentor any group can still delete their own membership.

### DIFF
--- a/app/policies/organization_member_policy.rb
+++ b/app/policies/organization_member_policy.rb
@@ -14,6 +14,7 @@ class OrganizationMemberPolicy < ApplicationPolicy
 
   def destroy?
     return false if leaving_self? && sole_admin?
+    return false if member_primary_mentor_of_any_group?
 
     org_admin? || leaving_self? || user.admin?
   end
@@ -32,6 +33,10 @@ class OrganizationMemberPolicy < ApplicationPolicy
 
     def leaving_self?
       record.user == user
+    end
+
+    def member_primary_mentor_of_any_group?
+      record.organization.groups.exists?(primary_mentor_id: record.user_id)
     end
 
     def sole_admin?
--- a/spec/policies/organization_member_policy_spec.rb
+++ b/spec/policies/organization_member_policy_spec.rb
@@ -110,4 +110,45 @@ describe OrganizationMemberPolicy do
     it { is_expected.not_to permit(:update) }
     it { is_expected.not_to permit(:destroy) }
   end
+
+  context "when the primary mentor of a group tries to delete their own membership" do
+    before do
+      @mentor = FactoryBot.create(:user)
+      @mentor_membership = FactoryBot.create(:organization_member, organization: @organization, user: @mentor,
+                                                                   role: :member)
+      FactoryBot.create(:group, name: "Mentored Group", primary_mentor: @mentor, organization: @organization)
+    end
+
+    let(:user) { @mentor }
+    let(:organization_member) { @mentor_membership }
+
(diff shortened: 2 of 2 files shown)


deepak0x:
Hi @naman79820 — opened this per your assignment. It mirrors the `leave?` guard but keys the check on the membership being removed (`record.organization.groups` / `record.user_id`), so it also blocks an admin from removing a mentor and orphaning a group. Added specs for both the self-removal and admin-removal cases. Would appreciate your review

naman79820 (approved):

naman79820:
> Hi @naman79820 — opened this per your assignment. It mirrors the `leave?` guard but keys the check on the membership being removed (`record.organization.groups` / `record.user_id`), so it also blocks an admin from removing a mentor and orphaning a group. Added specs for both the self-removal and admin-removal cases. Would appreciate your review

Thankss @deepak0x :))

PRATHAM2002-DS (approved):

PR #7631, opened 2026-07-08
Convert data to JSON if its a parameter object

### LINKED ISSUE #7444: 🐞 Bug(vuesim): Circuit data API returns invalid payload when created from Vue simulator

### Describe the bug

Projects created from the Vue simulator fail to render after loading, even though the circuit data API responds with HTTP 200.
The frontend successfully fetches circuit data from: `/api/v1/projects/:id/circuit_data` but the simulator canvas remains blank.

### Steps To Reproduce

after the merge of [#1082](https://github.com/CircuitVerse/cv-frontend-vue/pull/1082) 
1. login with admin access and enable the vuesim flipper at `:3000/flipper`
2. Save online any of the project in the simulator 
3. can see that embed view of the circuit doesn't render the circuit properly
4. shows a loading symbol

### Expected Behavior

The saved circuit should render correctly after fetching project data.
The API returns a JSON-encoded string instead of a JSON object, causing the frontend loader to receive invalid data.

### Screenshots

_No response_

```markdown
- OS:
- Browser:
- version:
```

### Additional context

part of gsoc'26 prj-4

Yes

ShinichiShi:
Fixes #7444 

 

This PR fixes the issue where the preview was not showing the circuit image. This is only an issue in the vuesim.

Cause of this issue: the vue simulator sent circuit data as a nested object within the JSON request body. Rails parses this data field into an ActionController::Parameters obejct rather than a JSON string. In earlier legacy simulator, used to have a JSON string. 

---

The Vue simulator sends circuit data as a nested object in the JSON request body. When Rails parses this request, `params[:data]` is received as an `ActionController::Parameters` object rather than a JSON string.

The existing `sanitize_data` method can return early when a project does not belong to an assignment. In that case, the parsed parameters object was returned without being serialized, causing circuit data to be stored in an incorrect format and preventing the project from loading correctly.

The implementation converts `ActionController::Parameters` to JSON before the assignment-based early return:

I considered handling all non-string values generically or checking for a regular `Hash`. However, the Vue request path is parsed by Rails as `ActionController::Parameters`, and a `Hash` check does not match this runtime type. I chose the explicit type check because it addresses the observed request shape with a minimal behavioral change while leaving existing JSON string payloads from the legacy simulator unchanged.

---

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved data handling during sanitization so parameter objects are converted to JSON first, helping prevent unexpected issues and making input cleanup more reliable.


app/helpers/simulator_helper.rb
```diff
@@ -32,6 +32,7 @@ def check_to_delete(data_url)
   end
 
   def sanitize_data(project, data)
+    data = data.to_json if data.is_a?(ActionController::Parameters)
```
tachyons, line 35: This will always be in the format of ActionController::Parameters, right ?

ShinichiShi: yes it's always in this format

Nihal4777 (approved):

app/helpers/simulator_helper.rb
tachyons: Then why this condition?

Nihal4777: When data is submitted through HTML input field in legacy simulator, it's already is flat JSON string but not when request is made through API with `application/json` payload.

tachyons: json serialization and deserialisation are expensive for large circuits, we shouldn't convert them multiple times , we need to make this consistent

Nihal4777: Agreed, 💯.It only acts as the immediate fix.
Should we create the follow up issue for this?

tachyons: yes please

PR #7669, opened 2026-07-19
fix: validate circuit JSON schema by key presence, not order

### LINKED ISSUE #7643: 🐞 Bug : Fragile JSON Validation in `Open.js` breaks file imports if JSON key order changes

### Describe the bug

When importing a saved circuit `.cv` file, the `ValidateData` function in `simulator/src/file/Open.js` relies on a fragile method to validate the JSON schema.

At line 33, it does this:
```javascript
if (JSON.stringify(Object.keys(parsedFileDate)) !== JSON.stringify(JSONSchema)) throw new Error('Invalid JSON data');
```

### Steps To Reproduce

1. Save any circuit locally (it downloads as a `.cv` JSON file).
2. Open the file in a text editor and safely swap the position of two keys at the root level (e.g., put `"timePeriod"` before `"name"`).
3. Save the modified JSON file.
4. Go to the simulator and try to import/open the file.
5. It fails to open and throws an "Invalid JSON data" error, despite the file containing all necessary data.

### Expected Behavior

The schema validation should iterate over the expected keys and check if they exist using `hasOwnProperty` or a proper schema validation library (like `ajv`), rather than relying on exact array stringification and key ordering. It should also safely ignore unrecognized keys to preserve forward compatibility.

### Screenshots

_No response_

```markdown
- OS: Windows
- Browser: Chrome
```

### Additional context

_No response_

Yes

bhuvan-somisetty:
Fixes #7643

## What this fixes

Closes #7643.

Importing a saved circuit (`.cv`) file fails with "Invalid file format" whenever the root-level JSON keys are not in the exact order the app originally wrote them, even though the file is perfectly valid and contains every field the simulator needs.

The cause is in `ValidateData()` in `simulator/src/file/Open.js`:

```js
if (JSON.stringify(Object.keys(parsedFileDate)) !== JSON.stringify(JSONSchema)) throw new Error('Invalid JSON data');
```

This compares the stringified key arrays directly, so it depends on both exact key order and an exact key set. Reordering two keys in a text editor (e.g. moving `timePeriod` before `name`), or having any extra/unrecognized key, causes a valid file to be rejected. The per-scope check a few lines down had the same fragility, just via a manual `keys.includes` loop instead.

## Fix

Added `hasSchemaKeys(data, schema)`, which checks that every expected key exists on the object via `hasOwnProperty`, independent of ordering, and simply ignores any extra keys instead of failing on them (so older/newer save formats stay importable). Both the top-level schema check and the per-scope schema check now use this same helper, removing the duplicated key-checking logic.

No behavior changes for already-valid files; this only relaxes the check to stop rejecting valid data.

## Testing

Added `simulator/spec/open.spec.js` covering:
- keys in original order
- keys in a different order
- extra/unrecognized keys present
- a required key missing
- non-object / null input

Ran `eslint` on the changed files — clean.

---

The bug is that `ValidateData()` compared `JSON.stringify(Object.keys(parsedFileDate))` against `JSON.stringify(JSONSchema)`, which is an order- and set-sensitive string comparison. Any valid `.cv` file whose keys weren't in the exact original order, or that had one extra key, was rejected outright even though nothing was actually wrong with it.

I considered sorting both key arrays before stringifying, but that still fails on extra/unrecognized keys and doesn't clearly express intent. Instead I added a small helper, `hasSchemaKeys(data, schema)`, that checks presence of each required key via `Object.prototype.hasOwnProperty` regardless of order, and doesn't care about any additional keys present. I replaced both the top-level check and the per-scope `keys.includes` loop with this same helper so there's one shared, well-tested piece of validation logic instead of two slightly different ones.

I tested this by adding `simulator/spec/open.spec.js`, covering: keys in original order, keys reordered, extra/unrecognized keys, a missing required key, and invalid (non-object/null) input — verifying `hasSchemaKeys` returns `true`/`false` correctly in each case.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON file validation to accept additional unrecognized fields while still requiring all necessary fields.
  * Improved scope validation by detecting missing required entries.
  * Invalid files continue to display an appropriate format error.

* **Tests**
  * Added coverage for valid, incomplete, reordered, extended, null, undefined, and non-object data.


bhuvan-somisetty:
Hi @tachyons @Nihal4777 @vedant-jain03 @naman79820 — would appreciate a review on this whenever you get a chance. It fixes #7643 by making the circuit import schema check order-independent instead of comparing stringified key arrays. Also looks like the CI workflows are waiting on approval to run since this is from a fork. Thanks!

tachyons:
Missing AI declaration from PR template

Issue #7407, opened 2026-05-25
Feature Grade passback via AGS


magic-peach:
### Is your feature request related to a problem? Please describe

Currently CircuitVerse has no way to send student grades back to the  LMS after an assignment is completed. Instructors have to manually export grades from CircuitVerse and re-enter them into Canvas, Moodle, or Blackboard. For large classes this is a significant overhead and introduces errors. The existing LTI 1.1 Basic Outcomes service is limited to a single decimal score with no support for additional metadata or multiple line items.

### Describe the solution you'd like

Implement grade passback using the LTI 1.3 Assignment and Grade Services (AGS) spec, so CircuitVerse can push scores directly into the LMS gradebook. This includes:

- Parse the AGS endpoint and scope claims from the LTI 1.3 launch params
- Use the lineItems endpoint to create or update gradebook columns per assignment
- Send score results to the LMS when an instructor grades a submission or when autograding completes
- Support partial scores and submission timestamps as part of the score payload
- Handle token refresh for AGS API calls using the OAuth 2.0 client credentials flow

Part of the broader LTI 1.3 upgrade tracked in #7405 

### Are you working on this?

Yes

PR #7764, opened 2026-08-09
feat(lti): add Lti::AccessToken for platform service calls

magic-peach:
Fixes part of #7407

 

none

---

Describe in your own words:
LTI 1.3 uses client-credentials with the idea that we sign the jwt with our private key, POST it to the platform's token endpoint with the scopes we want, and get back a token, which the platform verifies against our published JWKS.
Nothing above can call the API for ags and NRPS without this, so it needs to cache tokens and fetching a fresh one per call would double the round trips on an endpoint Canvas rate-limits.

Cache in a database table:easy to inspect but access tokens expire within the hour and table means a migration, a model, and a sweeper for data that is going to become useless soon
Cache one token per deployment: tokens are scope-bound, so a token issued for the score scope gets rejected by NRPS. That failure shows up as a 403

Tokens are refreshed 30 seconds before they expire, and a token that would already expire inside that window is not cached at all. This avoids the case where a cached token is valid when read and expired by the time the API call lands.
Transport failures, malformed JSON and a response with no access_token all raise `Lti::AccessToken::Error` so only one thing gets addressed
---

---

## Summary by CodeRabbit

* **New Features**
  * Added secure OAuth access-token retrieval for LTI integrations using signed authentication assertions.
  * Tokens are cached and reused while valid, with automatic refresh before expiration to support uninterrupted access.
  * Supports scope-specific token requests with consistent handling across equivalent scope lists.
  * Provides clear error reporting when token services reject requests, return incomplete data, or cannot be reached.
  * Improves reliability when connecting to external LTI services.


app/services/lti/access_token.rb
```diff
@@ -0,0 +1,64 @@
...
+        end
+
+        # The platform resolves kid against the tool's published JWKS.
+        def client_assertion(deployment, signing_key)
+          now = Time.now.to_i
+          JWT.encode(
+            { iss: deployment.client_id, sub: deployment.client_id,
+              aud: deployment.access_token_url, iat: now, exp: now + ASSERTION_TTL,
+              jti: SecureRandom.hex(16) },
+            signing_key, "RS256", { kid: JWT::JWK.new(signing_key).kid }
```
anxkhn: kid here will not match `Lti::KeyManager.public_jwk`.
`KeyManager` uses `JSON::JWK` thumbprint; `JWT::JWK.kid` defaults to a different digest.
platforms resolve kid against `/lti/jwks`, so token requests will fail.
use the same kid `KeyManager` publishes (e.g. `Lti::KeyManager.public_jwk[:kid]`).

app/services/lti/access_token.rb
```diff
@@ -0,0 +1,64 @@
...
+      private
+
+        def request(deployment, scopes, signing_key)
+          response = HTTP.timeout(TIMEOUT)
+                         .post(deployment.access_token_url,
+                               form: grant_params(deployment, scopes, signing_key))
+          raise Error, "token request failed: #{response.status}" unless response.status.success?
+
+          body = JSON.parse(response.body.to_s)
+          [body.fetch("access_token"), body.fetch("expires_in", 3600).to_i - REFRESH_LEEWAY]
```
anxkhn: a blank `access_token` is truthy and gets cached, so later `fetch` calls keep returning `""`.
require a non-blank `access_token` before cache/return; raise `Lti::AccessToken::Error` otherwise.

anxkhn (changes requested):
@magic-peach thanks for the solid work on this. left a couple of review notes below. happy to send a patch or jump on a call if that helps.

app/services/lti/access_token.rb
```diff
@@ -0,0 +1,80 @@
...
+        Rails.cache.write(key, token, expires_in: ttl) if ttl.positive?
+        token
+      end
+
+      private
+
+        def request(deployment, scopes, signing_key)
+          response = HTTP.timeout(TIMEOUT)
+                         .post(deployment.access_token_url,
+                               form: grant_params(deployment, scopes, signing_key))
```
Asrani-Aman, line 28: @magic-peach here How you get to know that these should be the Form params (feilds/data we need to send to the LMS)

is this Defineed in some LTI or LMS docs?

Asrani-Aman (approved):
Good Work @magic-peach r

app/services/lti/access_token.rb
Asrani-Aman: @magic-peach i think still you are using new JWT::JWK.new(signing_key)??

Can you please explain here 
Means I didn't get it how you resolved Anas Comment?

magic-peach: That line is showing the outdated version,  I replaced it in the 0bff731 commit so by that the now code uses a kid_for helper which is the  jwt::jwk.new(....) which is the same computation lti::keymanager#public_jwk so the kid in the assertion now matches what we publish at /lti/jwks

app/services/lti/access_token.rb
magic-peach: i got them mainly from the IMS security framework which is what LTI advantage has for service authorization, it specifies those grant params and also the RFC 7523 which  defines the client_assertion/client_assertion_type params and the required JWT claims

Asrani-Aman (approved):

PR #7767, opened 2026-08-09
feat(lti): add Lti::Score to publish grades to the platform

magic-peach:
Fixes part of #7407

 

none
---

Describe in your own words:
When a teacher grades a project in CircuitVerse; that score has to reach Canvas. AGS publishes scores to a line item's scores endpoint, specifying the user, the score, a timestamp, and how far along the grading is. The timestamp contains the platforms order updates by it and reject stale ones, so a re-grade with an older timestamp than what's already there gets dropped.

Build the scores URL by string concatenation: Canvas returns line item URLs with a query string so concatenation produces .../42?type=assignment/scores, which results in a 404
Return the parsed response body: The score response is either empty or a small object with a resultUrl and callers just need to know if it worked. Returning a boolean and raising on failure gives them to check instead of a body to parse.

The scores URL is built by parsing the line item URL and appending `/scores` to its path
A zero is treated as a real score , and there's a spec pinning that specifically, because conflicting the two would silently turn a legitimate zero into "not yet graded" in someone's gradebook.
The access token and line item URL are parameters rather than things the class fetches, which keeps it independent of the other AGS parts

---

---

## Summary by CodeRabbit

* **New Features**
  * Added support for publishing learner scores to LTI Advantage line items.
  * Score submissions include points earned, maximum points, completion and grading statuses, and a UTC timestamp.
  * Supports partial, zero, and completed scores.
  * Validates score details before submission and reports errors when requests fail or endpoints cannot be reached.


anxkhn (commented):
@magic-peach thanks for this. scores path + query keep looks good. one small edge case noted below. happy to help if you want a hand.

app/services/lti/score.rb
```diff
@@ -0,0 +1,42 @@
...
+        raise Error, e.message
+      end
+
+      private
+
+        # /scores goes on the line item path, leaving any query the platform put
+        # on the line item url intact.
+        def scores_url(lineitem_url)
+          uri = URI.parse(lineitem_url)
+          uri.path = "#{uri.path}/scores"
```
anxkhn: if lineitem_url path ends with /, this becomes `.../42//scores` and some platforms 404.
strip a trailing slash before appending `/scores`.

Asrani-Aman (approved):

Issue #7406, opened 2026-05-25
Feature Roster sync functionality


magic-peach:
### Is your feature request related to a problem? Please describe

When an instructor connects CircuitVerse to their LMS (Canvas, Moodle, 
Blackboard etc.), there is no automatic way to sync the student roster 
from the LMS into a CircuitVerse group. Instructors have to manually 
add students one by one, which is error-prone and time consuming for 
large classes. If a student drops the course or is added late, 
CircuitVerse has no way to know.

### Describe the solution you'd like

Implement roster sync via the LTI Names and Role Provisioning Services 
(NRPS) spec, so CircuitVerse can automatically pull the student list 
from the connected LMS. This includes:

- Fetch enrolled students and their roles from the LMS using the NRPS 
  endpoint provided in the LTI launch params
- Automatically create or map CircuitVerse accounts for enrolled students
- Add synced students to the corresponding CircuitVerse group
- Handle roster updates — remove or deactivate students who have dropped 
  the course
- Expose a manual "Sync Roster" button for instructors in the group 
  settings page

Part of the broader LTI 1.3 upgrade tracked in #<link parent issue number here>

### Additional context

GSoC26

### Are you working on this?

Yes

PR #7779, opened 2026-08-11
feat(lti): add Lti::Membership to read a course roster

magic-peach:
Fixes part of #7406

 

none

---

Describe in your own words:
The roster service needs to know who all are participating in a course this is done using the names and roster service (nrps) which is basically an endpoint that returs the member with their roles and status. Canvas returns rosters a page at a time so if only one page is read the remaining data is lost

Read the first page and stop: this gets rid of any class larger than one page leading to missing students in the roster list
Filter to active learners inside the client: drop handling needs to see the members who are left. If this class discards inactive members, that feature can't be built without changing it

Memberships URL and access token are parameters, not things the class fetches itself and it keeps independent of the token service, lets specs drive it directly, and matches the AGS services from the earlier phase so the whole LTI service layer reads consistently.

---

---

## New Features
- Added support for retrieving complete LTI membership lists, including inactive members.
- Automatically follows paginated membership results and resolves linked pages.
- Preserves membership roles and statuses.
- Supports authenticated requests using standard LTI membership formats.

## Bug Fixes
- Improved handling of unsuccessful responses, malformed data, invalid pagination links, connection failures, and pagination limits.
- Added clearer errors when membership responses are empty or invalid.

### DIFF
--- a/app/services/lti/membership.rb
+++ b/app/services/lti/membership.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Lti
+  class Membership
+    SCOPE = "https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly"
+    MEDIA_TYPE = "application/vnd.ims.lti-nrps.v2.membershipcontainer+json"
+    LEARNER = "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner"
+    INSTRUCTOR = "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor"
+    TIMEOUT = 5
+    MAX_PAGES = 50
+
+    class Error < StandardError; end
+
+    class << self
+      # Returns every member the platform reports, inactive ones included, since
+      # a roster sync has to see who left as well as who is there.
+      def fetch(memberships_url, access_token)
+        members = []
+        url = memberships_url
+
+        MAX_PAGES.times do
+          response = request(url, access_token)
+          members.concat(Array(parse(response)["members"]))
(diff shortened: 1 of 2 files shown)


app/services/lti/membership.rb
```diff
@@ -0,0 +1,50 @@
...
+    class << self
+      # Returns every member the platform reports, inactive ones included, since
+      # a roster sync has to see who left as well as who is there.
+      def fetch(memberships_url, access_token)
+        members = []
+        url = memberships_url
+
+        MAX_PAGES.times do
+          response = request(url, access_token)
+          members.concat(Array(parse(response)["members"]))
```
anxkhn: `Array(nil)` turns a 200 with no members key into `[]`. empty `"members": []` is valid; a missing/non-array members is not.
require `members.is_a?(Array)` and raise `Lti::Membership::Error` otherwise so a full-replace sync cannot treat a bad page as an empty class.

app/services/lti/membership.rb
```diff
@@ -0,0 +1,50 @@
...
+          url = next_page(response)
+          break if url.blank?
+        end
+
+        members
```
anxkhn, line 29: if next is still present after `MAX_PAGES`, this returns a partial roster with no error (and a self-next link duplicates the same page).
raise `Lti::Membership::Error` when the page cap is hit with a remaining next link.

anxkhn (commented):
@magic-peach thanks for the nrps roster work. left two notes on the fetch contract. happy to help with a patch or a quick call if useful.

Issue #7788, opened 2026-08-15
Verilog import: 12 reproducible defects found porting a 7-module CPU design


Spendry:
[janus-circuitverse-evidence.zip](https://github.com/user-attachments/files/31099271/janus-circuitverse-evidence.zip)

# Verilog import: 12 reproducible defects found porting a 7-module CPU design

## Summary

First, thank you for CircuitVerse. I have no formal hardware background, and
I learned to build real sequential logic in your schematic editor over the
past week: subcircuits, tunnels, splitters, a 16-bit datapath with a
registered control path. It worked, it was fun, and I got a machine running
on a canvas I could show people. That part of the tool taught me something.

Then I tried to bring a larger design in through **Circuit → New Verilog
Module**, and that is what this issue covers. The design is a seven-module
CPU architecture, about 2,600 gates. The same Verilog compiles and simulates
correctly under Icarus Verilog on EDA Playground and produces the waveform I
expect, so I do not think the design is at fault. I hit twelve separate
problems in the importer, and I am writing them up in the hope they are
useful.

I understand the schematic editor is the heart of the project and the Verilog
path is newer. Nothing below is a complaint about the tool as a whole. I have
saved-offline `.cv` exports from each stage and will attach or send any of
them.

---

## 1. Re-importing a module adds a scope instead of replacing it, and the global clock keeps simulating orphans

If you only look at one item here, I would suggest this one, because it
makes the others hard to diagnose.

After several import attempts my project held 41 scopes: 9 copies of one
module, 5 copies each of four more, 4 of another. Only 3 of the 41 were
referenced by anything. Deleting a subcircuit from the canvas leaves its
scope in place, and each re-import adds another.

`clockEnabled` is a project-level setting, so CircuitVerse simulates every
orphan, including the broken superseded ones. Their errors raise the same
toasts as live circuits, and the toast does not say which scope raised it.
I fixed a module, re-imported it, saw the same error, and concluded the fix
had failed. It had not. An old copy was still running.

I only found this by exporting the JSON and counting scopes.

**What would help**

- Name the scope in the error toast.
- Offer replace-on-reimport for a module of the same name.
- Drop or list scopes that nothing references.
- Skip simulation for scopes unreachable from the open circuit.

---

## 2. A port named `clk` disappears

The importer binds a port named `clk` to the global clock and generates no
pin. Renaming it to `clkin` produces a normal input pin. Nothing in the UI
[...]

PR #7793, opened 2026-08-17
fix(verilog): preserve clk/clock ports as input pins during module import (#7788)

UwaisBaig:
### Summary of Changes

Fixes part of CircuitVerse/cv-frontend-vue#1225 (Defect 2: A port named `clk` disappears):
- In `simulator/src/VerilogClasses.js`, updated `verilogInput` to always instantiate standard `Input` pins for input ports (including `clk` and `clock`) rather than silently converting them into standalone `Clock` oscillator generators.
- Ensures imported modules and subcircuits preserve their clock input pins on the circuit interface so parent circuits can connect and drive clock signals to them.

### Related Issue

- Fixes part of CircuitVerse/cv-frontend-vue#1225

## Summary by CodeRabbit

* **Bug Fixes**
  * Inputs named `clk` or `clock` are now handled consistently as standard input signals.
  * Configured bit widths are preserved for these inputs.

### DIFF
--- a/simulator/src/VerilogClasses.js
+++ b/simulator/src/VerilogClasses.js
@@ -101,12 +101,7 @@ class verilogUnaryGate{
 class verilogInput extends verilogUnaryGate {
     constructor(deviceJSON) {
         super(deviceJSON);
-        if (deviceJSON["net"] == "clk" || deviceJSON["net"] == "clock") {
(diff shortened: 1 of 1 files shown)


naman79820:
Wrong repo.

PR #7797, opened 2026-08-19
fix(verilog): resolve Verilog import defects and multi-module synthesis limits

### LINKED ISSUE #1225: Verilog import: 12 reproducible defects found porting a 7-module CPU design
[janus-circuitverse-evidence.zip](https://github.com/user-attachments/files/31099271/janus-circuitverse-evidence.zip)

# Verilog import: 12 reproducible defects found porting a 7-module CPU design

## Summary

First, thank you for CircuitVerse. I have no formal hardware background, and
I learned to build real sequential logic in your schematic editor over the
past week: subcircuits, tunnels, splitters, a 16-bit datapath with a
registered control path. It worked, it was fun, and I got a machine running
on a canvas I could show people. That part of the tool taught me something.

Then I tried to bring a larger design in through **Circuit → New Verilog
Module**, and that is what this issue covers. The design is a seven-module
CPU architecture, about 2,600 gates. The same Verilog compiles and simulates
correctly under Icarus Verilog on EDA Playground and produces the waveform I
expect, so I do not think the design is at fault. I hit twelve separate
problems in the importer, and I am writing them up in the hope they are
useful.

I understand the schematic editor is the heart of the project and the Verilog
path is newer. Nothing below is a complaint about the tool as a whole. I have
saved-offline `.cv` exports from each stage and will attach or send any of
them.

---

## 1. Re-importing a module adds a scope instead of replacing it, and the global clock keeps simulating orphans

If you only look at one item here, I would suggest this one, because it
makes the others hard to diagnose.

After several import attempts my project held 41 scopes: 9 copies of one
module, 5 copies each of four more, 4 of another. Only 3 of the 41 were
referenced by anything. Deleting a subcircuit from the canvas leaves its
scope in place, and each re-import adds another.

`clockEnabled` is a project-level setting, so CircuitVerse simulates every
orphan, including the broken superseded ones. Their errors raise the same
toasts as live circuits, and the toast does not say which scope raised it.
I fixed a module, re-imported it, saw the same error, and concluded the fix
had failed. It had not. An old copy was still running.

I only found this by exporting the JSON and counting scopes.

**What would help**

- Name the scope in the error toast.
- Offer replace-on-reimport for a module of the same name.
- Drop or list scopes that nothing references.
- Skip simulation for scopes unreachable from the open circuit.

---

## 2. A port named `clk` disappears

The importer binds a port named `clk` to the global clock and generates no
pin. Renaming it to `clkin` produces a normal input pin. Nothing in the UI
says this happened, so the subcircuit arrives with a pin missing and no
explanation.

---

## 3. A module cannot reference a module from an earlier import

`janus_touch` instantiates `janus_cell`. Importing the cell first, then the
touch, fails:

```
Module `\janus_cell' referenced in module `\janus_touch' in cell `\fit'
is not part of the design.
```

Pasting both modules in one import works. Import order does not matter,
because a previously imported circuit stays invisible to a later paste.

---

## 4. Paste size caps at 10,000 bytes

My eight-module design needed comment stripping and whitespace minification
to fit. I found the limit by hitting it. Documenting it would save time.

---

## 5. Memory arrays are unsupported

```verilog
reg [15:0] next_free [0:3];
```

This produces width errors rather than a message about arrays. I unrolled
four arrays into sixteen named registers with `if`-chain selection, and the
module imported.

---

## 6. Combinational `always @*` blocks appear unsupported

Five modules imported cleanly. All five used continuous `assign` plus
sequential `always @(posedge ...)` blocks and nothing else. The two that
failed were the two using `always @*`. Rewriting their combinational logic
as continuous assigns changed the failure.

---

## 7. Constructs that produce width errors on import

Each of these broke an import, and rewriting each one changed the error:

- ternaries inside wire declarations
- replication constants such as `{{(N-1){1'b0}}, 1'b1}`
- shift operators
- constants written directly in an instantiation port list
- part-selects written directly in an instantiation port list, e.g. `.req_row(rd_addr[7:0])`
- dangling output ports, e.g. `.demote()`

---

## 8. Vectors narrower than the design's main width fail

A 3-bit counter beside 32-bit data threw a width error. Widening the counter
to 32 bits fixed it. The same held for a 5-bit counter, an 8-bit output, an
8-bit register, and a 2-bit selector.

This one has a cost outside CircuitVerse. I now maintain a CircuitVerse
variant of the design that synthesizes about 8% larger under Yosys, 2,583
cells against 2,789 on one module, because those widened counters become
real flip-flops in silicon.

---

## 9. Long chains of magnitude comparisons break the importer

One module compared a register against nine constants in an if-ladder. Its
imported scope contained intermediate nodes at widths 5, 6, 26, 27, 28, 29,
30 and 31, and it would not place without errors. Cutting the ladder to two
comparisons changed the behavior.

Modules with two or three comparators import fine and also contain odd-width
nodes, so I suspect the count rather than the construct.

---

## 10. An unconnected subcircuit input poisons the whole circuit

Place an imported subcircuit and leave any input unwired, and CircuitVerse
raises a cascade of errors like `BitWidth Error 1 and 32`. An undefined
1-bit value reaches wide internal logic and every downstream net objects.

All six of my modules did this when placed alone, unwired, with no switches
on. The messages name internal widths, which reads as a defect inside the
module. I rewrote four modules chasing what turned out to be a missing wire, which is
the single biggest time sink in this list.

A message naming the pin, such as `input in_x of janus_touch is
unconnected`, would have saved me most of a day.

---

## 11. Changing an element's BitWidth after wiring leaves stale wire nodes

I wired everything while my tunnels sat at the default width 1, then set the
correct widths of 32, 16 and 2. The endpoints updated. The wire nodes between
them did not.

In the exported JSON this appears as a 32-bit subcircuit output connected to
a width-1 wire node. I counted 24 such mismatches across 12 wires. Deleting
and redrawing each wire fixed it. Nothing in the UI flags the stale node, and
you cannot repair it in place.

---

## 12. Tunnel identifiers cap at 5 characters, and mismatched pairs fail silently

The 5-character limit does not appear in the docs I found. Two tunnels
sharing a name but differing in width do not join and raise nothing.

---

## Minor: three number bases in one workflow

- The memory-block editor takes decimal and silently misreads hex-looking
  input. Typing `63` from a hex table stores sixty-three, not ninety-nine.
- Constant and Input bit arrays take binary. Typing `10` for sixteen gives
  two.
- Tunnel value readouts display hex.

I hit all three in one afternoon.

---

## Evidence

I can supply any of the following:

- `.cv` exports from each stage, including the 41-scope project and a clean
  8-scope one
- Verilog for all seven modules, plus the CircuitVerse variants
- an Icarus run of the same design on EDA Playground with the expected
  waveform: a control unit locks onto a data stream at cycle 3, detects a
  regime change at cycle 40 with one miss, and re-locks at cycle 43, while a
  companion counter walks 1, 2, 4, 8, 16, halves to 8 on the miss, then
  climbs by ones
- CircuitVerse's own Verilog export of the imported project, which shows
  correct port declarations, so the module interfaces survive the round trip
  and the trouble sits in the body

All of the above is attached as `janus-circuitverse-evidence.zip`, since
GitHub does not accept `.cv`, `.sv` or `.vcd` directly. The archive holds a
README describing each file.

Happy to split this into separate issues if that suits your workflow better.
I would also gladly package the design as a regression test. It is small,
self-contained, and exercises every construct above.

For what it is worth, I plan to keep building in the schematic editor either
way. Hand-wiring the cell taught me more about clocked logic than any
document could have, and I would not have gotten there without CircuitVerse.
If the importer reaches the point where it can take a design this size, I
will bring the rest of it over.

Ar1es-XD:
Fixes CircuitVerse/cv-frontend-vue#1225

This PR fixes multiple critical defects identified in Verilog import and synthesis for multi-module digital designs (#7788) and addresses code review feedback:

1. **Array Signal Vector Width Bug in `getBitWidth` (Defects 7, 8, 9, 10)**:
   - Updated `getBitWidth` to correctly inspect array lengths (`bitsJSON.length`), numbers, objects, and strings instead of using `Math.max` across net ID values.
   - Refactored using `Object.values` and `Number.isNaN` for strict ESLint compliance.

2. **Clock Port Exposure and Subcircuit Pin Loss (Defect 2)**:
   - Preserved `Input` creation for all input ports with their respective net names rather than converting to `Clock` elements.

3. **Inverted Comparison Gate Operations (Defect 3/Logic Bugs)**:
   - Fixed ALU input wiring order in `verilogGtGate` and `verilogLeGate` to align with ALU SLT (control code 7) semantics.

4. **Addition Gate Output Bitwidth Handling**:
   - Updated `verilogAdditionGate` to safely handle `outBitWidth <= bitWidth` and unexpected widths without producing disconnected/undefined outputs.

5. **Orphaned Scope Proliferation and Safe Circuit Deletion (Defect 1)**:
   - `YosysJSON2CV` tracks generated subcircuit IDs in `rootScope.verilogMetadata.subCircuitScopeIds`.
   - Updated `deleteCurrentCircuit` in `circuit.js` to account for subcircuit scopes, preventing the deletion of the last visible circuit and cleaning up all associated child scopes and DOM tabs when valid.
   - Added null safety checks in `Scope.checkDependency`.

6. **Intermediate Wire Node & Tunnel Bitwidth Propagation (Defect 11)**:
   - Added `propagateBitWidth` to propagate width updates across connected intermediate wire nodes (`type === 2`).
   - Fixed parameter mutation in `propagateBitWidth`.
   - Added bitwidth propagation in `Tunnel.prototype.setIdentifier` when synchronizing with paired tunnels.

7. **Unified Code Paste Size Limit Expansion (Defect 4)**:
   - Centralized `MAX_CODE_SIZE = 250_000` (250KB) in `SimulatorHelper`, shared by `SimulatorController` and `Api::V1::SimulatorController`.

8. **Tunnel Identifier Length Restriction (Defect 12)**:
   - Increased `mutableProperties.identifier.maxlength` on `Tunnel` from `5` to `50`.

9. **Unit, Integration, and Controller Tests**:
   - Added comprehensive Jest unit test suite in `simulator/spec/verilog.spec.js` (including deletion regression tests and tunnel bitwidth propagation).
   - Added RSpec controller tests in `spec/controllers/simulator_controller_spec.rb` and `spec/requests/api/v1/simulator_controller/verilog_cv_spec.rb` verifying exact 250KB limits and boundary rejection.

---

### PR Artifacts:
- **Diff**: https://github.com/CircuitVerse/CircuitVerse/pull/7797.diff
- **Patch**: https://github.com/CircuitVerse/CircuitVerse/pull/7797.patch

---

The root causes were isolated by tracing synthesis output from Yosys/DigitalJS through CircuitVerse's `YosysJSON2CV` and `VerilogClasses`. The signal width mismatch was caused by treating net ID arrays as numerical values rather than vector length. Comparison bugs were fixed by aligning with CircuitVerse ALU SLT semantics (control code 7). Orphan scope accumulation and deletion issues were solved by tracking child circuit IDs and calculating remaining scope counts before confirmation. All constants and controllers are unified in `SimulatorHelper`.

---

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum accepted Verilog code size to 250 KB.
  * Support tunnel identifiers up to 50 characters.
* **Bug Fixes**
  * Improved bit-width propagation across connected elements and tunnels.
  * Corrected comparison, addition, wiring, and subcircuit handling.
  * Improved Verilog input and clock labeling.
  * Improved cleanup when deleting circuits with generated subcircuits.
* **Tests**
  * Added coverage for Verilog imports, synthesis, code-size limits, bit widths, comparisons, additions, wiring, and subcircuits.

### DIFF
--- a/app/controllers/api/v1/simulator_controller.rb
+++ b/app/controllers/api/v1/simulator_controller.rb
@@ -27,8 +27,9 @@ class Api::V1::SimulatorController < Api::V1::BaseController
     render json: { success: true, message: "Issue submitted successfully" }, status: :ok
   end
   # rubocop:enable Metrics/MethodLength
+  include SimulatorHelper
 
-  MAX_CODE_SIZE = 250_000 # 250KB limit
+  MAX_CODE_SIZE = SimulatorHelper::MAX_CODE_SIZE
 
   # POST /api/v1/simulator/verilogcv
   def verilog_cv
--- a/app/controllers/api/v1/simulator_controller.rb
+++ b/app/controllers/api/v1/simulator_controller.rb
@@ -28,7 +28,7 @@ class Api::V1::SimulatorController < Api::V1::BaseController
   end
   # rubocop:enable Metrics/MethodLength
 
-  MAX_CODE_SIZE = 10_000 # 10KB limit
+  MAX_CODE_SIZE = 250_000 # 250KB limit
 
   # POST /api/v1/simulator/verilogcv
   def verilog_cv
--- a/app/controllers/simulator_controller.rb
+++ b/app/controllers/simulator_controller.rb
@@ -130,7 +130,7 @@ class SimulatorController < ApplicationController
     head :ok, content_type: "text/html"
   end
 
-  MAX_CODE_SIZE = 10_000 # 10KB limit
+  MAX_CODE_SIZE = 250_000 # 250KB limit
 
   def verilog_cv
     if params[:code].to_s.bytesize > MAX_CODE_SIZE
--- a/app/controllers/simulator_controller.rb
+++ b/app/controllers/simulator_controller.rb
@@ -130,7 +130,7 @@ class SimulatorController < ApplicationController
     head :ok, content_type: "text/html"
   end
 
-  MAX_CODE_SIZE = 250_000 # 250KB limit
+  MAX_CODE_SIZE = SimulatorHelper::MAX_CODE_SIZE
 
   def verilog_cv
     if params[:code].to_s.bytesize > MAX_CODE_SIZE
--- a/app/helpers/simulator_helper.rb
+++ b/app/helpers/simulator_helper.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 module SimulatorHelper
+  MAX_CODE_SIZE = 250_000 # 250KB limit
+
   def return_image_file(data_url)
     str = data_url[("data:image/jpeg;base64,".length)..]
     if str.to_s.empty?
--- a/simulator/spec/verilog.spec.js
+++ b/simulator/spec/verilog.spec.js
@@ -84,6 +84,16 @@ describe('Verilog Import and Synthesis Fixes', () => {
         // outBitWidth < bitWidth
         const addNarrower = new yosysTypeMap.Addition({ bits: { in1: 8, in2: 8, out: 4 } });
         expect(addNarrower.outputSplitter).toBeDefined();
+
+        // outBitWidth > bitWidth + 1 (e.g. 4-bit addition with 8-bit output)
+        const addMuchWider = new yosysTypeMap.Addition({ bits: { in1: 4, in2: 4, out: 8 } });
+        expect(addMuchWider.outputSplitter).toBeDefined();
+        expect(addMuchWider.zeroConstant).toBeDefined();
+        expect(addMuchWider.zeroConstant.state).toBe('000');
+        expect(addMuchWider.outputSplitter.bitWidthSplit).toEqual([4, 1, 3]);
+        expect(addMuchWider.outputSplitter.outputs[2].connections)
+            .toContain(addMuchWider.zeroConstant.output1);
+        expect(addMuchWider.output.bitWidth).toBe(8);
     });
 
     test('YosysJSON2CV tracks subcircuit scopes in rootScope.verilogMetadata.subCircuitScopeIds', () => {
--- a/simulator/spec/verilog.spec.js
+++ b/simulator/spec/verilog.spec.js
@@ -9,12 +9,12 @@ import yosysTypeMap from '../src/VerilogClasses';
 import Input from '../src/modules/Input';
 import Tunnel from '../src/modules/Tunnel';
 import Node, { propagateBitWidth } from '../src/node';
-import { scopeList } from '../src/circuit';
+import { scopeList, newCircuit, deleteCurrentCircuit } from '../src/circuit';
 
 jest.mock('codemirror');
 
 describe('Verilog Import and Synthesis Fixes', () => {
-    CodeMirror.fromTextArea.mockReturnValueOnce({ setValue: (text) => {} });
+    CodeMirror.fromTextArea.mockReturnValueOnce({ setValue: () => {} });
     setup();
 
     test('verilogInput keeps clk and clock as Input element and retains port label', () => {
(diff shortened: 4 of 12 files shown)


naman79820:
Wrong repository.

PR #7799, opened 2026-08-19
Feature: add email-based member invitations (backend)

### LINKED ISSUE #7798: Feature: Add email-based member invitations for organizations (backend)

Organizations currently have no way to invite members by email. Group invitations already support this (existing users are added directly, new emails receive a pending invitation), but organizations lack the equivalent backend. Without it, there's no data model or controller logic to power an email-based invite flow, and invited people who don't yet have an account can't be added on signup.

Add backend support for email-based organization member invitations, mirroring the existing Groups pattern:

- Extend `PendingInvitation` to belong to an organization (with a role), so it can represent a group OR an organization invitation.
- Update the organization members `create` action to accept multiple emails and a role  adding existing users as members immediately, and creating pending invitations for emails without an account.
- On signup, consume any pending organization invitations so the new user joins with the invited role.

This is the backend for the invite UI, which is handled in a separate PR.

_No response_

### Additional context

GSoC 26 - Project 6 Enterprise & Institutional Organization

### Are you working on this?

Yes

naman79820:
Fixes #7798

 

Adds backend support for email-based organization member invitations, mirroring the existing Groups pattern.

- Extended `PendingInvitation` to optionally belong to an organization (with a role), so it can represent either a group or an organization invitation. Added a validation requiring at least one of group/organization to be present.
- Rewrote `OrganizationMembersController#create` to accept multiple emails and a role: existing users are added as members immediately with the chosen role, and emails without an account get a `PendingInvitation` carrying that role.
- Extended the signup consumption (`User#create_members_from_invitations`) so that when an invited email registers, their pending organization invitation becomes a membership with the invited role.
- Added two migrations (add `organization_id` + `role` + index to `pending_invitations`; add the organization foreign key) and updated the schema.

---

The problem: organizations had no way to invite members by email. Groups already solved this, so I followed the same pattern for consistency rather than inventing a new one.

The core decision was how to represent an organization invitation for someone who doesn't have an account yet. I extended the existing `PendingInvitation` model to optionally belong to an organization (adding `organization_id` and `role` columns, and making the group association optional) instead of creating a separate model. This reuses the existing mailer and the signup-consumption logic, and keeps the behavior identical to Groups. I added a validation so an invitation must belong to either a group or an organization, never neither.

The `create` action extracts valid emails using `Devise.email_regexp`, filters out people who are already members and the current user, then for each remaining email either adds them directly (if they have an account) or creates a pending invitation (if they don't)  in both cases stamping the chosen role. The role is whitelisted against `%w[admin mentor member]` and defaults to member. I extracted the per-email branch into a private `invite_by_email` helper to keep the action within the method-length limit. Empty submissions are handled by wrapping the emails param in `Array()` so nothing is called on nil.

On signup, `create_members_from_invitations` checks whether each pending invitation is for a group or an organization and creates the appropriate membership, passing the stored role for organization invitations. I verified this end to end: a pending mentor invitation plus a new user with that email results in the user joining as a mentor.

---

---

## Summary by CodeRabbit

* **New Features**
  * Invite multiple people to an organization using email addresses.
  * Assign roles when sending organization invitations.
  * Automatically add existing members or invite new people.
  * Support processing both organization and group invitations.

* **Bug Fixes**
  * Filter invalid, duplicate, and self-addressed email invitations.
  * Preserve assigned roles when processing invitations.
  * Display invitation email status after submission.
  * Prevent duplicate pending invitations for the same organization and email address.


db/migrate/20260818200955_add_organization_to_pending_invitations.rb
```diff
@@ -0,0 +1,9 @@
+class AddOrganizationToPendingInvitations < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :pending_invitations, :organization, null: true, index: { algorithm: :concurrently }
+    add_column :pending_invitations, :role, :integer
```
vedant-jain03: I think let's add null to be true here so it's should not break for existing groups records!

app/models/user.rb
```diff
@@ -62,13 +62,23 @@ class User < ApplicationRecord
...
+    if invitation.group_id.present?
       group = invitation.group
-      ActiveRecord::Base.transaction do
-        GroupMember.where(group_id: group.id, user_id: id).first_or_create!
-        group.add_member_to_organization(self)
-        invitation.destroy!
-      end
+      GroupMember.where(group_id: group.id, user_id: id).first_or_create!
+      group.add_member_to_organization(self)
+    elsif invitation.organization_id.present?
```
PRATHAM2002-DS, line 74: This important piece of the PR is untested there are no tests for " invited email signs up, becomes a member with the right role"

app/controllers/organization_members_controller.rb
```diff
@@ -12,19 +12,16 @@ class OrganizationMembersController < ApplicationController
...
-        format.json { render json: @organization_member.errors, status: :unprocessable_content }
-      end
-    end
+    role = organization_member_params[:role].presence_in(%w[admin mentor member]) || "member"
+    submitted = Array(organization_member_params[:emails]).map { |e| e.to_s.strip.downcase }
+    emails = submitted.grep(Devise.email_regexp)
+    present_members = User.where(id: @organization.organization_members.pluck(:user_id)).pluck(:email)
+    newly_added = emails - present_members - [current_user&.email&.downcase]
+    newly_added.each { |email| invite_by_email(email, role) }
+    notice = Utils.mail_notice(submitted, emails, newly_added)
```
PRATHAM2002-DS: issue: This makes the notice English-only — t(".success") was already translated in 7 locales, and Utils.mail_notice returns hardcoded English. Worth moving the message into the locale files with interpolation so Crowdin can pick it up.

app/models/pending_invitation.rb
```diff
@@ -1,10 +1,22 @@
 # frozen_string_literal: true
 
 class PendingInvitation < ApplicationRecord
-  belongs_to :group
+  belongs_to :group, optional: true
+  belongs_to :organization, optional: true
+
+  validate :group_or_organization_present
+
```
PRATHAM2002-DS, line 8: blocker: Org invitations will crash the mailer,  @group is nil and the template calls @group.name, so the invite row saves but no email goes out. Needs an org branch in the mailer and its own template.

basically this callback fires for org invitations too, but the mailer only knows about groups — PendingInvitationMailer#new_pending_email does @group = pending_invitation.group, which is nil here, and the template calls @group.name. So org invites raise in the job: the row saves, the page says "will be invited", but no email is ever sent. Subject line is wrong too ("Added to a group in CircuitVerse").

app/controllers/organization_members_controller.rb
```diff
@@ -96,6 +93,19 @@ def check_organizations_feature_flag
       redirect_to root_path, alert: t("feature_not_available")
     end
 
+    def invite_by_email(email, role)
+      user = User.find_by(email: email)
+      if user.nil?
+        PendingInvitation.create_or_find_by!(organization_id: @organization.id, email: email) do |invite|
```
PRATHAM2002-DS: issue: If the invitation already exists, the block never runs, so the role stays whatever it was.

@naman79820 Test this scenario and tell me what you observe: " Invite someone as admin, then re-invite as member"

app/controllers/organization_members_controller.rb
```diff
@@ -12,19 +12,16 @@ class OrganizationMembersController < ApplicationController
...
-    end
+    role = organization_member_params[:role].presence_in(%w[admin mentor member]) || "member"
+    submitted = Array(organization_member_params[:emails]).map { |e| e.to_s.strip.downcase }
+    emails = submitted.grep(Devise.email_regexp)
+    present_members = User.where(id: @organization.organization_members.pluck(:user_id)).pluck(:email)
+    newly_added = emails - present_members - [current_user&.email&.downcase]
+    newly_added.each { |email| invite_by_email(email, role) }
+    notice = Utils.mail_notice(submitted, emails, newly_added)
+    redirect_to members_organization_path(@organization), notice: notice
+  rescue ActiveRecord::RecordInvalid => e
```
PRATHAM2002-DS, line 26: Check once if we are accounting for right exception here

app/controllers/organization_members_controller.rb
```diff
@@ -12,19 +12,16 @@ class OrganizationMembersController < ApplicationController
...
-        end
-        format.json { render json: @organization_member.errors, status: :unprocessable_content }
-      end
-    end
+    role = organization_member_params[:role].presence_in(%w[admin mentor member]) || "member"
+    submitted = Array(organization_member_params[:emails]).map { |e| e.to_s.strip.downcase }
+    emails = submitted.grep(Devise.email_regexp)
+    present_members = User.where(id: @organization.organization_members.pluck(:user_id)).pluck(:email)
+    newly_added = emails - present_members - [current_user&.email&.downcase]
+    newly_added.each { |email| invite_by_email(email, role) }
```
PRATHAM2002-DS: suggestion: here we are looping over each email without batching and no cap too, say admin submitted 10K emails, that will result in 10K queries

PRATHAM2002-DS (changes requested):

PRATHAM2002-DS:
@vedant-jain03 can you review the migration files in this PR

app/controllers/organization_members_controller.rb
naman79820: Yes you are right

app/models/pending_invitation.rb
naman79820: I've already made those changes in #7771. I wanted to keep this PR backend-focused.

app/models/user.rb
naman79820: I already have it on #7771 but ig i should add those tests inside this pr.

app/controllers/organization_members_controller.rb
naman79820: changed to 50 now

PRATHAM2002-DS (approved):

PRATHAM2002-DS (approved):

vedant-jain03 (approved):

PR #7807, opened 2026-08-22
fix: remove stray Zone.Identifier blocking Windows checkout

YashKumar-404:
This deletes one file: `app/assets/images/organization_banner.png:Zone.Identifier`.

It is an empty (0-byte) Windows "mark of the web" artifact - the marker NTFS attaches to files
downloaded through a browser. It was committed accidentally alongside the banner image in #7785
(`f96e139`).

**Why it needs to go:** a colon is reserved on NTFS as the alternate-data-stream separator, so git
cannot write this path into the index on Windows. The symptom is that `git clone` reports

```
warning: clone succeeded, but checkout failed
```

and leaves an **empty working tree** - while `git log` looks perfectly healthy, which makes it easy
to misdiagnose as a network or disk problem. In practice the repository cannot be checked out on
Windows at all unless the contributor knows to set `core.protectNTFS=false` locally.

Two things worth noting for reviewers:

- **Sparse-checkout does not work around this.** The path is rejected while git writes the *index*,
  which happens before any working-tree filter applies.
- **This is the only affected path.** `git ls-files` finds exactly one tracked path containing a
  colon, so this single deletion fully resolves it.

The file is empty and referenced nowhere. `organization_banner.png` itself is untouched and still
used by `app/component/organization_card_component.html.erb`.

As a side note, both review bots on this PR reported "no reviewable changes" for the file, which is
consistent with it being empty.

Fixes #7785


naman79820:
Thanks @YashKumar-404, it was a mistake on my side while working on my organization index PR. I was planning to remove that file in PR #7771, but never mind, you handled my work. Thanks man, totally appreciated!

naman79820:
Since I'm using WSL, whenever I paste a PNG file it creates a Zone.Identifier clone, and while working on my PR I accidentally added that file... one more reason for me to switch to Linux o.o

PR #7834, opened 2026-09-02
fix: prevent ArgumentError from invalid search page param (#7830)

### LINKED ISSUE #7830: ArgumentError: invalid value for Integer(): "1' AND 1=1 UNION SELECT NULL-- -" (ArgumentError)
Sentry Issue: [CIRCUITVERSE-CORE-155](https://circuitverse.sentry.io/issues/144080612/)

```
ArgumentError: invalid value for Integer(): "1' AND 1=1 UNION SELECT NULL-- -" (ArgumentError)

      value = Integer(value)
                      ^^^^^
  app/lib/adapters/pg_adapter.rb:48:in `perform_search'
    results.paginate(page: query_params[:page], per_page: MAX_RESULTS_PER_PAGE)
  app/lib/adapters/pg_adapter.rb:24:in `search_project'
    perform_search(
  app/lib/search.rb:15:in `call'
    @adapter.send("search_#{@relation.name.downcase}", @relation, @query_params)
  app/queries/generic_query.rb:11:in `results'
    @search.call
  app/helpers/search_helper.rb:14:in `query'
    return ProjectsQuery.new(query_params, Project.public_and_not_forked).results,
...
(113 additional frame(s) were not displayed)
```
This issue was automatically created by Sentry via [New issue](https://sentry.io/organizations/circuitverse/monitors/alerts/12755/)

YashKumar-404:
## Description

Fixes #7830.

Project and user search crashed with `ArgumentError: invalid value for Integer(): "..."` whenever the `page` query param was not a valid integer. The value from the Sentry report (`CIRCUITVERSE-CORE-155`) was the SQL-injection probe `"1' AND 1=1 UNION SELECT NULL-- -"`.

### Root cause

`query_params[:page]` was passed directly to `will_paginate`'s `paginate(page: ...)`, which calls `Integer()` on the value. Any non-numeric string raises `ArgumentError`.

### Fix

- Add `BaseAdapter#sanitize_page`, which coerces the page param to a safe positive integer using `Integer(page.to_s, 10, exception: false)` and defaults `nil`, blank, non-numeric, and non-positive values to `1` — never raising.
- Apply it in `PgAdapter#perform_search` and both `SolrAdapter` search methods before pagination.

Handling `nil` is covered: `nil.to_s` is `""`, which `Integer(..., exception: false)` returns `nil` for, falling back to page `1` (this was the gap flagged on the earlier attempt in #7832).

### Tests

Added specs to `spec/lib/adapters/pg_adapter_spec.rb` covering the SQL-probe string, `nil`, non-positive (`"0"`), and valid numeric-string (`"3"`) page values.

## Bug Fixes
- Improved search pagination handling for invalid, missing, or non-positive page values.
- Numeric page values provided as text are now processed correctly.
- Prevented malformed page input from causing search errors, including injection-style values.
- Search results now consistently fall back to the first page when pagination input cannot be interpreted safely.

### DIFF
--- a/.github/workflows/greet.yml
+++ b/.github/workflows/greet.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Welcome new contributors
         uses: actions/first-interaction@v3
         with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
           issue_message: |
             ## 🚀 Welcome to CircuitVerse! 
             
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -118,6 +118,7 @@ RSpec/SpecFilePathSuffix:
 RSpec/VerifiedDoubles:
   Exclude:
     - 'spec/lib/adapters/pg_adapter_spec.rb'
+    - 'spec/lib/adapters/solr_adapter_spec.rb'
 
 # Offense count: 6
 Rails/HasManyOrHasOneDependent:
--- a/app/lib/adapters/base_adapter.rb
+++ b/app/lib/adapters/base_adapter.rb
@@ -2,5 +2,17 @@
 
 module Adapters
   class BaseAdapter
+    protected
+
+      # Coerces an arbitrary page param into a safe, positive integer.
+      #
+      # will_paginate calls Integer() on the page value, so a non-numeric
+      # string (e.g. a SQL-injection probe) raises ArgumentError. This
+      # normalises nil, blank, non-numeric, array and non-positive values
+      # to page 1 without ever raising.
+      def sanitize_page(page)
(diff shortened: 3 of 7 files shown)


JituRewar (commented):
@YashKumar-404 Could we add tests for the other pagination paths as well? The fix is applied to `PgAdapter#search_user`, `SolrAdapter#search_project`, and `SolrAdapter#search_user`, but the current specs only cover `PgAdapter#search_project`. A malformed `page` value test for each path would ensure the fix is consistently applied.

YashKumar-404:
Thanks @JituRewar — added malformed-page coverage for the remaining paths. PgAdapter#search_user now has its own sanitization block, and I added spec/lib/adapters/solr_adapter_spec.rb covering SolrAdapter#search_project and #search_user (asserting the sanitized page inside the Sunspot search block). Each path now has a malformed-page example plus numeric-string coercion.

PR #7835, opened 2026-09-02
feat: add CommentPolicy and CommentThreadPolicy (part 2 of Commontator removal)

Shruti2110-coder:
Depends on #7831.

Second step of replacing the Commontator gem. Ports the comment and thread
permission rules into Pundit policies.

## What this adds

- `CommentThreadPolicy` — read, comment, subscribe, close/reopen
- `CommentPolicy` — show, create, update, destroy, restore, vote
- Policy specs (32 examples)

## Rules ported

| Rule | Behaviour |
|---|---|
| Read | Project is public, or the user has project view access |
| Moderator | Any site admin |
| Create | Signed in, thread open, and can read |
| Edit | Author only, thread open, not deleted, not edited by someone else |
| Delete | Author when thread is open, **or** any moderator even when closed |
| Undelete | Any moderator, or the author if they were the one who deleted it |
| Vote | Signed in, not the author, thread open, comment not deleted |

## Note on how these were derived

Rules were read from `commontator-7.0.1`'s `Comment` model rather than
inferred from the initializer. That surfaced three behaviours the config alone
does not describe:

1. In `can_be_deleted_by?`, the moderator branch returns **before** the
   closed-thread check, so moderators may delete inside a closed thread.
2. `can_be_edited_by?` also requires `editor.nil? || user == editor`.
3. `can_be_voted_on?` blocks voting on closed threads and deleted comments.

## What this does not do

Nothing calls these policies yet. The controllers still use Commontator, so
there is no behaviour change.

## Approach note

Neither policy calls `super`. `ApplicationPolicy#initialize` raises for nil
users, but comments on public projects must stay readable anonymously — the
existing API spec asserts this. `ProjectPolicy` skips `super` for the same
reason.

## Summary by CodeRabbit

* **New Features**
  * Added project comment threads with replies and chronological discussions.
  * Added comment editing, soft deletion, restoration, and voting permissions.
  * Added thread subscription permissions.
  * Added moderator controls to close and reopen discussions.
  * Added visibility and access controls for public and private projects.
  * Added validation to prevent invalid reply relationships and comment cycles.
* **Tests**
  * Added coverage for comment behavior, thread lifecycle, validation, restoration, and access permissions.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+# Mirrors Commontator::Comment. Belongs to a thread rather than directly to the
+# commentable, so that thread level state (open/closed, subscriptions) has a
+# home and the existing API shape is preserved.
+class Comment < ApplicationRecord
+  BODY_MAX_LENGTH = 10_000
+
+  belongs_to :comment_thread, inverse_of: :comments
+  belongs_to :user
+  belongs_to :editor, class_name: "User", optional: true
+  belongs_to :parent, class_name: "Comment", optional: true
+
+  has_many :replies, class_name: "Comment",
+                     foreign_key: :parent_id,
+                     inverse_of: :parent,
+                     dependent: :destroy
+
+  validates :body, presence: true, length: { maximum: BODY_MAX_LENGTH }
+  validate :parent_shares_thread
+
+  scope :kept, -> { where(deleted_at: nil) }
+  scope :roots, -> { where(parent_id: nil) }
+  scope :chronological, -> { order(created_at: :asc) }
+
+  def deleted?
+    deleted_at.present?
+  end
+
+  def edited?
+    editor_id.present?
+  end
+
+  def soft_delete!
+    update!(deleted_at: Time.current)
+  end
+
+  def restore!
+    update!(deleted_at: nil)
+  end
+
+  private
+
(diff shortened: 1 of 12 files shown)


JituRewar (commented):
@Shruti2110-coder The cycle guard is currently application-level only, so writes through `update_column`, bulk SQL, or concurrent updates can still introduce a parent cycle. Since `replies` uses `dependent: :destroy`, such a cycle could cause recursive destruction issues.
Could we ensure the upcoming Commontator data migration and any bulk-write paths explicitly validate/preserve this invariant? If DB-level enforcement isn't practical, a migration-time validation would at least prevent importing invalid cycles.

Shruti2110-coder:
@JituRewar Yes, agreed — that's the plan. The data migration will do a bulk import from commontator_comments, which bypasses model validations entirely, so it's the one path where a cycle could realistically be introduced.

I'll include in that PR:

A cycle check on the source data before insert, so anything invalid in the Commontator tables is caught rather than carried over.
A verification query after the import confirming no parent_id chain loops.
The migration failing loudly rather than skipping bad rows, so a problem is visible instead of silent.

Beyond that, any bulk write goes through the model rather than raw SQL. I'll link back to this thread from the migration PR so the requirement is traceable.

JituRewar:
@Shruti2110-coder Great

PR #7876, opened 2026-09-17
Fix JSON response when leaving non-member organization

### LINKED ISSUE #7729: bug: leave action sends HTML redirect for JSON clients when user is not a member

### Describe the bug

The `leave` action in `OrganizationMembersController` sends an HTML 302 redirect to JSON clients when the requesting user is not a member of the organization.

At [line 62](https://github.com/CircuitVerse/CircuitVerse/blob/50c6b5fff210e8a151d10b34b5ecfaccc0874326/app/controllers/organization_members_controller.rb#L62), the early-return guard uses a bare `redirect_to` that is not wrapped in a `respond_to` block:

```ruby
# DELETE /organizations/1/leave
# DELETE /organizations/1/leave.json
def leave
  @organization_member = @organization.organization_members.find_by(user: current_user)

  if @organization_member.nil?
    redirect_to organizations_path, alert: t(".not_a_member")   # line 62 — bare redirect_to
    return
  end

  authorize @organization, :leave?

  @organization_member.destroy
  respond_to do |format|
    format.html { redirect_to organizations_path, notice: t(".success") }
    format.json { head :no_content }
  end
end
```

In Rails, `redirect_to` always issues an HTTP 302 redirect regardless of the request `Accept` header or format. Every other success and error path in this controller correctly uses `respond_to` with both `format.html` and `format.json` handlers (at lines 17, 33, 50, and 69). This one path was missed.

Because the leave route has no format restriction (`delete :leave, to: 'organization_members#leave'`), JSON requests can reach this code path. A JSON client that sends `DELETE /organizations/:id/leave.json` while not a member receives a 302 HTML redirect instead of a JSON error response.

### Steps To Reproduce

1. Authenticate as a user who is **not** a member of a given organization.
2. Send `DELETE /organizations/<org_id>/leave.json` with `Accept: application/json` (e.g. via `curl -X DELETE https://<host>/organizations/<org_id>/leave.json -H "Accept: application/json"` with a valid session cookie).
3. Observe the response.

This can also be verified by static inspection:

```
grep -n 'redirect_to\|respond_to' app/controllers/organization_members_controller.rb
```

This shows four `respond_to` blocks (lines 17, 33, 50, 69) and one bare `redirect_to` at line 62 that is outside any `respond_to` block.

### Expected Behavior

A `DELETE /organizations/:id/leave.json` request from a user who is not a member should receive a JSON error response (e.g. `{ "error": "..." }`) with an appropriate HTTP status code, consistent with how all other error and success paths in this controller handle JSON requests.

### Screenshots

No response

- OS:
- Browser:
- version:

### Additional context

Confirmed at commit `50c6b5ff`. The existing specs do not test the JSON format or the not-a-member case for the `leave` action, so this path is not covered by the test suite.

No

VishalSirimalla:
## Summary

Fixes #7729.

When a user who is not a member of an organization sends a JSON request to leave the organization, the controller now returns a JSON error response instead of an HTML redirect.

### Changes
- Preserve the existing HTML redirect behavior.
- Return a JSON error with HTTP 422 for JSON requests.
- Added controller specs covering both HTML and JSON behavior.

## Testing
- `git diff --check` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling when someone attempts to leave an organization they do not belong to.
  - HTML requests redirect to the organizations page with a localized error message.
  - JSON requests return a localized error payload with an appropriate unprocessable-content status.

- **Tests**
  - Added coverage confirming consistent behavior for both HTML and JSON leave requests from non-members.


JituRewar (commented):
@VishalSirimalla The translation key in this expectation seems to be incorrect.
`OrganizationMembersController#leave` uses `t(".not_a_member")`, which resolves to `organization_members.leave.not_a_member`, but the test currently checks `organizations.leave.not_a_member`.
Could you update the expectation to use `I18n.t("organization_members.leave.not_a_member")` so the test validates the actual error message returned by the controller?

VishalSirimalla:
> @VishalSirimalla The translation key in this expectation seems to be incorrect. `OrganizationMembersController#leave` uses `t(".not_a_member")`, which resolves to `organization_members.leave.not_a_member`, but the test currently checks `organizations.leave.not_a_member`. Could you update the expectation to use `I18n.t("organization_members.leave.not_a_member")` so the test validates the actual error message returned by the controller?

Thanks for pointing this out, and sorry for the delayed response. You're right — the expectation is using the wrong translation scope. I'll update it to I18n.t("organization_members.leave.not_a_member") and push the fix.

Issue #7878, opened 2026-09-18
🐞 Bug: Pin labels are all rendered at the top-left corner in circuit layout


michael-lehn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When editing the layout of a circuit, the labels of the input/output pins are no longer positioned next to their corresponding pins. Instead, all pin labels are rendered at approximately the same position in the top-left corner of the circuit.
The pin nodes themselves are positioned correctly; only their labels are misplaced.
This seems to be a recent regression. I have used the layout functionality before without this problem.

### Steps To Reproduce

1. Open a circuit containing several input/output pins.
2. Open the circuit layout editor.
3. Observe the labels of the pins.

### Expected Behavior

Each pin label should be displayed next to its corresponding pin.

### Screenshots

<img width="1792" height="800" alt="Image" src="https://github.com/user-attachments/assets/71a62144-b4ba-4cd9-a03d-c3801e27aec9" />

### Device Information [optional]

```markdown
- OS: macOS
- Browser: Safari, Chrome, Firefox
- version: 14.7.6
```

### Additional context

This affects existing circuits as well, so it does not appear to be related to how a particular circuit was created.

### Are you working on this issue?

No

maltheshhh:
Thanks for the detailed report! I dug into this and found the root cause.

This isn't actually in this repo — the simulator now lives in
[cv-frontend-vue](https://github.com/CircuitVerse/cv-frontend-vue), vendored
here as a submodule. The bug is in `src/simulator/src/layoutMode.ts`,
in `renderLayout()`:

```ts
tempBuffer.Input[i].x + typeof info[1] === "number" ? info[1] : parseInt(info[1] as string)
```

Due to operator precedence (`+`/`typeof` bind tighter than `===`, which binds
tighter than `?:`), this parses as:

```ts
(tempBuffer.Input[i].x + typeof info[1]) === "number" ? info[1] : parseInt(...)
```

`typeof info[1]` is always `"number"`, so `node.x + "number"` is string
concatenation — which never strictly equals `"number"`. The ternary always
falls to the `else` branch, silently dropping `node.x`/`node.y` and returning
just the small offset from `determineLabel()`. That's why every label lands
near the canvas origin while the pin markers (drawn separately) stay correct.

Confirmed by comparing against the older `v1/src/simulator/src/layoutMode.ts`
in the same repo, which does the straightforward (correct)
[...]

PR #7880, opened 2026-09-18
fix: correct operator precedence in layout pin-label positioning (#7878)

maltheshhh:
## Description
Fixes a regression where, in the circuit layout editor, all input/
output pin labels rendered clustered near the top-left corner of the
canvas instead of next to their pins. The pin markers themselves were
positioned correctly — only the labels were affected.

## Root cause
In `renderLayout()` (src/simulator/src/layoutMode.ts), the label
position was computed with a type-guard ternary that was missing
parentheses:

```ts
tempBuffer.Input[i].x + typeof info[1] === "number" ? info[1] : parseInt(info[1] as string)
```

Due to JS operator precedence (`+`/`typeof` > `===` > `?:`), this
evaluates as `(node.x + typeof info[1]) === "number" ? ... : ...`.
`typeof info[1]` is always the string `"number"`, so `node.x + "number"`
is string concatenation, which never strictly equals `"number"` — the
ternary always falls to the `else` branch (`parseInt(info[1])`),
silently dropping `node.x`/`node.y` from the final coordinate.

Confirmed against the older, pre-TS-conversion copy of this file
(`v1/src/simulator/src/layoutMode.ts`), which does the straightforward
(correct) `tempBuffer.Input[i].x + info[1]`.

## Fix
Wrap the ternary in parentheses so it's evaluated before being added
to the node's x/y, for both the Input and Output label loops.

## Testing
- Opened a circuit with several input/output pins → Edit Layout.
- Before: all labels bunched near (0,0)-ish, regardless of pin position.
- After: each label renders next to its correct pin.

Fixes #<issue-number-in-this-repo-if-any> / CircuitVerse/CircuitVerse#7878

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled frontend component to a newer revision.
  * No user-facing feature or behavior changes were identified.


naman79820:
https://github.com/CircuitVerse/CircuitVerse/issues/7878#issuecomment-5812562639

PR #7894, opened 2026-09-20
fix: keep search dropdown highlight on the selected option

### LINKED ISSUE #7893: 🐞 Bug: search resource dropdown keeps the arrow-key highlight after you back out

### Describe the bug

Open the resource dropdown in the search bar, move through it with the arrow keys, then back out with Escape or by clicking away. The dropdown keeps the green highlight on whatever you arrowed to, even though nothing was selected. Reopen it and it still shows the wrong row as chosen.

The button next to it still reads the real value, so the two disagree and the dropdown is the one lying. It is easy to believe you are about to search Users when the form will still search Projects.

The cause is that one class does two jobs. `setSelectedOption` marks the committed choice with `.active` (`app/javascript/controllers/search_bar_controller.js` line 91, via `updateActiveOption`), and `navigateOptions` moves that same class around for the arrow keys (line 150). Nothing puts it back when the dropdown closes without a choice, and `closeDropdown` only removes the `show` class.

Escape goes through `handleEscapeKey` to `closeDropdown`, clicking away goes through `handleOutsideClick` to the same place. Both leave the stale highlight behind.

### Steps To Reproduce

1. Go to https://circuitverse.org/search?q=adder&resource=Projects
2. Click the "Projects" button to the left of the search box.
3. Press the down arrow once. "Users" lights up green.
4. Press Escape.
5. Open the dropdown again. "Users" is still green, while the button still says "Projects".

### Expected Behavior

Backing out of the dropdown should leave it showing the option that is actually selected.

### Screenshots

Step 5 on circuitverse.org. The button says Projects, the dropdown says Users.

dropdown disagrees with the button

- OS: Windows 11
- Browser: Chrome
- version: 152.0.7977.84

### Additional context

Reading the state off the live page while doing the steps above: `hiddenSelect` stays `Projects` and the button label stays `Projects` throughout, while the highlighted option goes `Projects` on load, `Users` after the arrow key, and stays `Users` after Escape and after reopening.

Yes

iamtanishqjain:
Fixes #7893

`closeDropdown` now puts the `.active` class back on the option that is actually selected, using the value already sitting in the hidden field. Ten lines, one file. The issue has the why.

Both ways of backing out go through `closeDropdown`, Escape and clicking away, so one place covers them.

Selecting something still works the way it did. `setSelectedOption` writes the hidden field before it closes, so the restore lands on the option just picked, and the `updateActiveOption` call after it agrees.

I checked the three paths in a browser, with the controller, the markup and the styles taken from the repo:

| | selected | button | highlighted |
|---|---|---|---|
| arrow to Users, click away | Projects | Projects | Projects |
| arrow to Users, press Enter | Users | Users | Users |
| click Projects in the list | Projects | Projects | Projects |

Before the change the first row highlighted Users while the button read Projects.

No test. There are no tests for anything under `app/javascript` at the moment, and `jest.config.js` only picks up `simulator/spec`, so adding one would mean widening the test setup in a bug fix. Happy to do it in a follow-up, or here if you would rather.

Both after backing out of the dropdown with Escape and reopening it.

Before:

before

After:

after

---

One class was doing two jobs, marking the committed choice and marking wherever the arrow keys had got to, with nothing to tell them apart once the dropdown closed.

The other way to fix it is to give the keyboard cursor its own class and leave `.active` alone. That is arguably tidier, but it means new CSS and touching the component template, and the highlight would then have to be cleared somewhere anyway. Restoring on close is the smaller change and keeps the existing styling.

`highlightSelectedOption` reuses `getSelectOptions`, which was already there for the arrow keys, rather than introducing another query. It does nothing if no option matches, so a resource value that is not in the list leaves the dropdown alone instead of throwing.

---

---

## Summary by CodeRabbit

- **Bug Fixes**
  - The search dropdown now keeps the currently selected option highlighted after it closes.
  - Reopening the dropdown makes it easier to identify the active selection, improving clarity when reviewing or changing search criteria.


iamtanishqjain:
@tachyons I've updated the branch to master, which kicked off the workflow runs, but they are waiting on approval so the required checks can't start. Could you approve them when you have a minute?

It's a ten line change to one Stimulus controller. Steps to reproduce on circuitverse.org are in #7893.

iamtanishqjain:
@naman79820 @vedant-jain03 this one has had no human review yet and no reviewer was auto-requested. Would either of you have time? It's 10 lines in one Stimulus controller, and #7893 has steps to reproduce on circuitverse.org.

cc @CircuitVerse/circuitverse-core-team since CODEOWNERS covers `app/javascript` under the global rule.

